PR #4, opened 2019-02-06
Added lesson03 assignments

williamchang6:


students/will_chang/lesson03/mailroom.py
```diff
@@ -0,0 +1,56 @@
...
+            donor[3] = donor[1]/donor[2]
+            print("\nDear {},\n\nThank you for your generous donation of ${:,}.\n\nSincerely,\nThe Best Charity Ever".format(donor[0],ask_donor))
+    if(not name_in_list): # Check if the entered name is not in the list 
+        ask_donor = int(input("Please enter a donation amount: "))
+        initial_donor_list.append([ask_name,ask_donor,1,ask_donor])
+        print("\nDear {},\n\nThank you for your generous donation of ${:,}.\n\nSincerely,\nThe Best Charity Ever".format(ask_name,ask_donor))
+
+if __name__ == '__main__':
+    quit_criteria = False # while loop flag
+    while(not quit_criteria): # loop until ended by user input
```
natasha-aleksandrova: you dont need the parens here -> `while not quit_criteria`

also, I would suggest to use `while True:` for simpler code.

students/will_chang/lesson03/mailroom.py
```diff
@@ -0,0 +1,56 @@
...
+        user_prompt = input("\nPlease choose from the following menu of actions:\n[1] Send a Thank You\n[2] Create a Report\n[3] Quit\n\nInput: ")
+        if(user_prompt == '1' or user_prompt.lower() == "send a thank you"):
+            thank_you()
+        elif(user_prompt == '2' or user_prompt.lower() == "create a report"):
+            list_data(initial_donor_list)
```
natasha-aleksandrova: you don't need to pass this data, as it is available globally

students/will_chang/lesson03/mailroom.py
```diff
@@ -0,0 +1,56 @@
+initial_donor_list = [
+    ["Warren Buffett", 650.00, 2, 325.00],
+    ["Jack Bogle", 15000.50, 1, 15000.50],
+    ["William Boeing", 450.45, 3, 150.15],
+    ["George Clooney", 10000.00, 2, 5000.00],
+    ["Orville Wright", 95000000.00, 1, 950000.00]
+]
+
+def list_data(input_tuple):
```
natasha-aleksandrova: is it really passing in a tuple? you're passing in a list here; however, no need to pass anything in per my other comment.

students/will_chang/lesson03/mailroom.py
```diff
@@ -0,0 +1,56 @@
...
+
+def thank_you():
+    """Prompts user to input new donor information."""
+    ask_name = input("\nPlease enter a full name or type 'list' for a list of the donor names: ")
+    name_in_list = False
+    while(ask_name.lower() == "list"):
+        print([donor[0] for donor in initial_donor_list])
+        ask_name = input("\nPlease enter a full name or type 'list' for a list of the donor names: ")
+    for donor in initial_donor_list:
+        if(ask_name == donor[0]): # Check if the entered name is already in the list
```
natasha-aleksandrova: no need for parens here as well.

students/will_chang/lesson03/mailroom.py
```diff
@@ -0,0 +1,56 @@
...
+            name_in_list = True
+            ask_donor = int(input("Please enter a donation amount: "))
+            donor[1] += ask_donor
+            donor[2] += 1
+            donor[3] = donor[1]/donor[2]
+            print("\nDear {},\n\nThank you for your generous donation of ${:,}.\n\nSincerely,\nThe Best Charity Ever".format(donor[0],ask_donor))
```
natasha-aleksandrova: you should `break` here to avoid continue looping once you would the donor.

students/will_chang/lesson03/mailroom.py
```diff
@@ -0,0 +1,56 @@
...
+    if(not name_in_list): # Check if the entered name is not in the list
```
natasha-aleksandrova: instead use an `else` with for loop, if no entry found then it will go into else block. let me know if you need me to expand on this.

students/will_chang/lesson03/slicing_lab.py
```diff
@@ -0,0 +1,65 @@
+def exchange_first_last(seq):
+    """Returns a copy of the given sequence with the first and last values swapped."""
+    if(len(seq) == 1): #Prevents duplicates if sequence only has one value.
+        seq_copy = seq[:]
```
natasha-aleksandrova: no need for special case here, your line 6 will work for all cases.

students/will_chang/lesson03/slicing_lab.py
```diff
@@ -0,0 +1,65 @@
...
+    return seq_copy
+    
+    
+def exchange_every_other(seq):
+    """Returns a copy of the given sequence with every other item removed."""
+    seq_copy = seq[:0]
+    for i in range(len(seq)):
+        if i % 2 == 0: #Copy every other item of the original sequence into the new sequence.
+            seq_copy += seq[i:i+1]
+    return seq_copy
```
natasha-aleksandrova: the answer is simply : `seq[::2]`

PR #10, opened 2019-02-19
adding lesson06 assignments

williamchang6:


students/will_chang/lesson06/test_mailroom_part_4.py
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/env python
+
+from mailroom_part_4 import find_max_len_str, list_data, active_names, name_prompt, update_donor_list, thank_you, write_letter
+import datetime
+import os.path
+import pathlib
+
+def test_1():
```
natasha-aleksandrova: it is best to name tests after what you're testing, for example: `test_find_max_len_str`

PR #26, opened 2019-03-12
Completed HTML render assignment

KLehning:
Completed HTML render project for Lehning. Only HTML render assignment should be in this pull, if anything else is added it is on accident :)

### DIFF
--- a/students/kyle_lehning/Lesson07/html_render.py
+++ b/students/kyle_lehning/Lesson07/html_render.py
@@ -12,6 +12,7 @@ from abc import ABC
 class Element(object):
 
     tag = "html"
+    indent = "    "
 
     def __init__(self, content=None, **kwargs):
         if content is None:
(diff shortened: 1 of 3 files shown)


students/kyle_lehning/Lesson07/html_render.py
```diff
@@ -23,24 +27,30 @@ def __init__(self, content=None, **kwargs):
     def append(self, new_content):
         self.contents.append(new_content)
 
-    def render(self, out_file):
```
natasha-aleksandrova: this comment is actually for line 23

```
>>> def test(**kwargs): print(kwargs)
...
>>> test()
{}
```
you dont need the `if kwargs.__len__() is not 0:` check at all .. and normally you would check via `if kwargs:`

you also shouldn't be using `__len__` magic method but instead use `len()` method.

as you can see in my above example kwargs is `{}` when nothing is passed in so you can safely use for loop over empty dictionary.

PR #27, opened 2019-03-14
Lesson 3 

ericae14:
I'm checking in all assignments for lesson 3


students/erica_edwards/Lesson03/Lesson3_Slicing_Lab.py
```diff
@@ -0,0 +1,47 @@
+def exchange_first_last(sequence):
+    a = sequence[-1:]  
+    b = sequence[1:-1]  
+    c = sequence[0:1]
+    result = a + b + c
+    return result
```
natasha-aleksandrova: can simplify to `return seq[-1:] + seq[1:-1] + seq[:1]`

students/erica_edwards/Lesson03/Lesson3_Slicing_Lab.py
```diff
@@ -0,0 +1,47 @@
...
+
+def every_other_item(sequence):
+    result = sequence[0:-1:2]
```
natasha-aleksandrova: can simplify to `seq[::2]`

students/erica_edwards/Lesson03/Lesson3_Slicing_Lab.py
```diff
@@ -0,0 +1,47 @@
...
+    return result
+
+def first_last_middle(sequence):
+    x = int(len(sequence)/3)
+    result = sequence[x:-x:2]
+    return result
```
natasha-aleksandrova: there should be one with "With the first and last 4 items removed, and every other item in between" ?

students/erica_edwards/Lesson03/mailroom.py
```diff
@@ -0,0 +1,97 @@
+
+from operator import itemgetter
+
+Donors = [('Sandy Pie', [75]),
```
natasha-aleksandrova: lower case (snake_case) for vars -> `donors`

students/erica_edwards/Lesson03/mailroom.py
```diff
@@ -0,0 +1,97 @@
...
+          ('Judy Smith', [75, 100, 1000]),
+          ('Mike Jones', [75, 1000]),
+          ('Joe Smith', [75, 100, 2000]),
+          ('Kelly Blue', [75, 150, 275])]
+
+def menu():
+    quit = False
+    while not quit:
```
natasha-aleksandrova: use `while True` with `break` statement

bool flags are useful for other use cases.

students/erica_edwards/Lesson03/mailroom.py
```diff
@@ -0,0 +1,97 @@
...
+        print('''Menu:   
+                    1. Send a Thank You.
+                
+                    2. Create a Report.
+            
+                    3. Quit.''')
+        print()
+        response = input("Please enter a number to make your selection. ")
+        if response == '1':
+            donor_name()
```
natasha-aleksandrova: function name is a bit confusing and doesn't tell the reader what the function does?

`send_thank_you` would be more suitable

students/erica_edwards/Lesson03/mailroom.py
```diff
@@ -0,0 +1,97 @@
...
+            
+        if response == '2':
+            totals()    
+
+        if response == '3':
+            quit = True
+
+def list_donors():
+    print(Donors)
+    donor_name()
```
natasha-aleksandrova: * list donors functionality should list donor names (not just print the whole data structure)
* calling donor_name() should be done outside of this function

students/erica_edwards/Lesson03/mailroom.py
```diff
@@ -0,0 +1,97 @@
...
+
+def donor_name():
+    done = False
+    
+    # While collecting user input.
+    while not done:
+        donorName = input("Enter the full name of the donor. (or q to quit) ")
```
natasha-aleksandrova: snake_case style for var names

students/erica_edwards/Lesson03/mailroom.py
```diff
@@ -0,0 +1,97 @@
...
+        
+        if donorName == 'q':
+            return
+        # If user types list print Donors       
+        elif donorName.upper() == "LIST":
+            list_donors()
+            continue
```
natasha-aleksandrova: per my comment above, `list_donors` should not call your `donor_name` function as you're already in a while loop and it will stay in this function still

students/erica_edwards/Lesson03/mailroom.py
```diff
@@ -0,0 +1,97 @@
...
+        else:  # user didn't type list    
+            if donorName not in [item[0] for item in Donors]:
```
natasha-aleksandrova: it would be more efficient to use `for` loop and exit out of the loop as soon as donor is found.

students/erica_edwards/Lesson03/mailroom.py
```diff
@@ -0,0 +1,97 @@
...
+                Donors.append((donorName,list()))
+        donation(donorName)
+        done = True
+    
+def list_donors():
+    print(Donors)
+    donor_name()
+    return
```
natasha-aleksandrova: function defined twice? no need for return here

students/erica_edwards/Lesson03/mailroom.py
```diff
@@ -0,0 +1,97 @@
+
+from operator import itemgetter
+
+Donors = [('Sandy Pie', [75]),
+          ('Judy Smith', [75, 100, 1000]),
+          ('Mike Jones', [75, 1000]),
+          ('Joe Smith', [75, 100, 2000]),
+          ('Kelly Blue', [75, 150, 275])]
```
natasha-aleksandrova: 👍 correct data structure

students/erica_edwards/Lesson03/mailroom.py
```diff
@@ -0,0 +1,97 @@
...
+    print(Donors)
+    donor_name()
+    return
+
+def donation(donorName):    
+    donationAmount = float(input("Enter the donation amount (or q to quit) "))
+    if donationAmount == 'q':
+        return
+    else:
+        donorIndex = [item[0] for item in Donors].index(donorName)
```
natasha-aleksandrova: same here, you're looping over full list and then list has to search index over the list - could be optimized with a single for loop or by passing the donor entry that you have already found prior

students/erica_edwards/Lesson03/mailroom.py
```diff
@@ -0,0 +1,97 @@
...
+        name, donations = Donors[donorIndex]
+        donations.append(donationAmount)
+        send_thank_you(name, donationAmount)
+        return
+
+def send_thank_you(name, donationAmount):
+    print(f'Thank you {name} for your donation of ${donationAmount} dollars.')
+    print('We appeciate your generous suport of our club.')
+    print()
+    return
```
natasha-aleksandrova: no need for `return` statement

students/erica_edwards/Lesson03/mailroom.py
ericae14: Is this what you mean? 
```python
            for item in donors:
                if item[0] == donor_name:
                    break
                else: 
                    donors.append((donor_name, list()))
```

natasha-aleksandrova: close 🙂 

```
for donor in donors:
    if donor[0] == donor_name:
        donor[1].append(donation)  # update existing donor
else:  # notice indentation here
    # new donor
    donors.append((name, [donation]))
```

PR #29, opened 2019-03-15
Completed Lesson04 Files for Lombardi

lombardic:
The following files have been updated for lesson04 and are ready for review:
mailroom_part2.py
trigrams.py

### DIFF
--- a/students/chris_lombardi/Lesson04/mailroom_part2.py
+++ b/students/chris_lombardi/Lesson04/mailroom_part2.py
@@ -69,10 +69,10 @@ def send_letters():
 
 def thankyou_note(entry):
     note = (f'Dear {entry},\n\n\tThank you for your generous donation of '
-          f'${sum(donor_log.get(entry)):,.2f}!\n\tWe appreciate the '
-          f'{len(donor_log.get(entry)):d} total donation(s) that you have made.'
-          '\n\tYour donation will be put to good use.'
-          '\n\n\tSincerely,\n\t-The Mailroom Team')
+            f'${sum(donor_log.get(entry)):,.2f}!\n\tWe appreciate the '
+            f'{len(donor_log.get(entry)):d} total donation(s) that you have made.'
+            '\n\tYour donation will be put to good use.'
+            '\n\n\tSincerely,\n\t-The Mailroom Team')
     return note
 
 def initialize_donors():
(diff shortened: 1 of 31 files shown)


students/chris_lombardi/Lesson04/mailroom_part2.py
```diff
@@ -69,10 +69,10 @@ def send_letters():
...
 def thankyou_note(entry):
     note = (f'Dear {entry},\n\n\tThank you for your generous donation of '
-          f'${sum(donor_log.get(entry)):,.2f}!\n\tWe appreciate the '
-          f'{len(donor_log.get(entry)):d} total donation(s) that you have made.'
-          '\n\tYour donation will be put to good use.'
-          '\n\n\tSincerely,\n\t-The Mailroom Team')
+            f'${sum(donor_log.get(entry)):,.2f}!\n\tWe appreciate the '
+            f'{len(donor_log.get(entry)):d} total donation(s) that you have made.'
+            '\n\tYour donation will be put to good use.'
+            '\n\n\tSincerely,\n\t-The Mailroom Team')
```
natasha-aleksandrova: since your code has already been merged for mailroom I can leave all comments in one block here:
* line 68: `f.close()` no need to close since `with` handles it for you
* line 41: use `sum()` function instead of manually adding donations
* line 28: create_report function is missing sorting feature
* line 92: no need to use `set` or extract into variable -> `if option not in dict_menu_opts` is the same

students/chris_lombardi/Lesson04/trigrams.py
```diff
@@ -0,0 +1,70 @@
...
+    new_text = []
+    first_pair = random.choice(list(trigrams.keys()))
+    new_text.extend(first_pair)
+    avail_words = trigrams.get(first_pair)
+    new_text.append(random.choice(trigrams.get(first_pair)))
+
+    for i in range(1,length_string-1):
+        next_pair = tuple(new_text[i:i+2])
+        #Test for a valid key in the trigram dictionary.
+        if trigrams.get(next_pair) == None:
```
natasha-aleksandrova: you would use `is` to compare objects (None is an object); for example `is None` `is True`

in this case however, you can just do `if trigrams.get(next_pair):`

students/chris_lombardi/Lesson04/trigrams.py
```diff
@@ -0,0 +1,70 @@
...
+
+    return " ".join(new_text)
+
+def read_data_file(filename):
+    """Convert a *.txt file into a string"""
+    data_file = ""
+    with open(filename, 'r') as f:
+        for line in f:
+            data_file += line
+    f.close()
```
natasha-aleksandrova: no need for f.close()

PR #31, opened 2019-03-15
Adding lesson09 assignment files

williamchang6:


students/will_chang/lesson09/mailroom_oo.py
```diff
@@ -0,0 +1,187 @@
...
+        Parameters
+        ----------
+        donor_list : list
+            List of donors with their associated total donation, number of donations, and average donation
+        """
+        self.donor_list = []
+        
+    def add_donor(self, donor):
+        """Add donor information to the donor list"""
+        self.donor_list.append([donor.name, donor.total_donation(), donor.num_donation(), donor.avg_donation()])
```
natasha-aleksandrova: your donor list should contain donor objects only, no need to extract all of the information out of them.

for example:
```
self.donor_list = [Donor("bob"), Donor("john")]
for donor in self.donor_list:
    print(donor.name)
```
makes a lot more sense that storing all the primitive data out of those objects.

students/will_chang/lesson09/mailroom_oo.py
```diff
@@ -0,0 +1,187 @@
...
+        
+    def donor_names(self):
+        """Return a list of the active donor names"""
+        self.names = []
+        return [donor[0] for donor in self.donor_list]
```
natasha-aleksandrova: then this would be `return [donor.name for donor in self.donor_list]` which also gives you readability.

donor[0] is less readable than donor.name

students/will_chang/lesson09/mailroom_oo.py
```diff
@@ -0,0 +1,187 @@
...
+    while True:  # loop until user inputs something other than the keys in option_select
+        user_prompt = input("\nPlease choose from the following menu of actions:\n[1] Send a Thank You\n[2] Create a Report\n[3] Send letters to everyone\n[Press any other key to quit.]\n\nInput: ")
+        if user_prompt not in option_select:
+            quit_program()
+        elif user_prompt == '1':
+            thank_you()
+        elif user_prompt == '2':
+            donor_collection.list_data()
+        elif user_prompt == '3':
+            donor_collection.write_letter()
```
natasha-aleksandrova: you could have still used the switch dict here

PR #40, opened 2019-03-20
adds string formatting exercise

jkrecek3:


students/jim_krecek/lesson03/strformat_lab.py
```diff
@@ -0,0 +1,50 @@
...
+# Task 2
+########
+print("\nTask 2\n")
+
+print("file_{:0>3d} :   {:.2f}, {:.2E}, {:.3g}".format(*start))
+
+# Task 3
+########
+def formatter(tup):
+    L = len(tup)
```
natasha-aleksandrova: variables should be lower case and preferably more than one letter

natasha-aleksandrova:
don't forget to submit on the course website as well

PR #55, opened 2019-03-27
adds $ prompt for new donor

jkrecek3:


students/jim_krecek/lesson03/mailroom.py
```diff
@@ -8,37 +8,44 @@ def send_ty():
...
+    data.sort(reverse=True, key=lambda list: sum(list[1:]))
     for d in data:
-        total = sum(d[1:])
-        average = total/(len(d)-1)
-        print("{:<25} ${:>11.2f}  {:>10}  ${:>12.2f}".format(d[0],total,len(d)-1,average))
+            total = sum(d[1:])
+            average = total/(len(d)-1)
+            num = len(d)-1
+        
+            print("{:<25s} ${:>11.2f}  {:>10d}  ${:>12.2f}".format(d[0],total,num,average))
```
natasha-aleksandrova: this is over-indented

students/jim_krecek/lesson03/mailroom.py
```diff
@@ -8,37 +8,44 @@ def send_ty():
...
+                d.append(int(askmoney))
+                emailtext = "\nDear {},\n\nThank you for your generous donation of ${}.\n\nSincerely, Jim".format(d[0],askmoney)
+                print(emailtext)
+                break
     
-    for d in data:
-        if askname.lower() == d[0].lower():
-            usename = askname
+        if askname not in d[0] and askname.lower() != "list":
+            data.append([askname])
```
natasha-aleksandrova: I was hoping you would catch the bug here from my previous comments 🙂 

This code still needs to be corrected, I will provide 2 additional comments 1. bug repro 2. suggestion on the code.

natasha-aleksandrova: **BUG** line 20 in your code
```
What do you want to do?
 -1 Send a Thank You
 -2 Create a Report
 -3 Quit
>>> 1

Enter the full name of the donor or type 'list' to see the list of donors:
>>> joel

How much did this person donate?
>>> 2

Dear joel,

Thank you for your generous donation of $2.

Sincerely, Jim

What do you want to do?
 -1 Send a Thank You
 -2 Create a Report
 -3 Quit
>>> 1

Enter the full name of the donor or type 'list' to see the list of donors:
>>> joe

What do you want to do?
 -1 Send a Thank You
 -2 Create a Report
 -3 Quit
>>> 2
Donor Name               | Total Given | Num Gifts | Average Gift
-----------------------------------------------------------------
Pat Carrier               $     106.00           2  $       53.00
Jim Krecek                $      93.00           3  $       31.00
Nate Secinaro             $      65.00           3  $       21.67
Zac Fisher                $      21.00           3  $        7.00
Jess Reid                 $       3.00           1  $        3.00
joel                      $       2.00           1  $        2.00

What do you want to do?
 -1 Send a Thank You
 -2 Create a Report
 -3 Quit
>>>
```

natasha-aleksandrova: suggested code:
```
def send_thank_you():
    response = input("<...ask>").lower()
    if response == "list":
        for d in data
            print(d[0])
    else:
        donation = float(input("Amount?"))
        for donor in data:
            if response == donor[0]:
                donor.append(donation)
                break
        else:
            data.append([response, donation])
        print("thanks ...")

```

jkrecek3: Clever bug you produced there with the 'joe' vs 'joel' case... I have modified my new code to use '!=' instead of 'not in,' and it appears to work. Thanks for the suggestion. Let me know what you think of the new code when posted.

natasha-aleksandrova: that solves the bug 🙂 

this is good to merge now, however the code I suggested is still better and more pythonic. Couple additional things on this line:
`if askname != d[0] and askname.lower() != "list":`

if you started with an empty `data` structure your program would crash because `d` is not defined 
`askname.lower() != "list"` is redundant because you've already checked for that in line 7

PR #63, opened 2019-03-29
Resumbit of mailroom.py for Assignment 1, with corrections made

RandiPet:


students/randi_peterson/session03/mailroom.py
```diff
@@ -1,44 +1,42 @@
...
+        donations = len(donor_lst[j]) - 1
         # Add up all donations
-        k = 1
-        while k < donations + 1:
-            total_given += donorlst[j][k]
-            k += 1
+        total_given = sum(donor_lst[j][1:])
         avg_gift = total_given / donations
         donor_report.append([name, total_given, donations, avg_gift])
         j += 1
```
natasha-aleksandrova: same comment as before about the for loop -> `for donor in donor_lst` is all you need.

you're also incrementing `j` here while using it in a loop with enumerate.

PR #65, opened 2019-03-30
adds dict_lab exercise

jkrecek3:


students/jim_krecek/lesson04/dict_lab.py
```diff
@@ -0,0 +1,73 @@
...
+d1.popitem()
+print(d1)
+
+#add 'fruit':'Mango' to dictionary
+d1['fruit'] = 'Mango'
+print(d1)
+
+#display dictionary keys
+print("Below are the keys for this dictionary:")
+for x in d1.keys():
```
natasha-aleksandrova: note that `keys()` is not required `for x in d1:` will work too

PR #68, opened 2019-04-01
added fibonacci file

deitesla:
I'm having trouble getting my assertions code to work in the tests piece of the fibonacci assignment. When I run the tests manually in iPython via the functions (e.g. fibonacci(3)), they work fine, but when I run the entire series.py, I get an assertion error, no matter which test I run. Any thoughts?

Also, my apologies if I went about asking my question through the wrong channel. Let me know if I should do something different in the future. Thanks!

### DIFF
--- a/students/lee_deitesfeld/series.py
+++ b/students/lee_deitesfeld/series.py
@@ -1,0 +1,80 @@
+#--------------------------
+#Title: Number sequences
+#Description: Determine value of nth value in sequence
+#Original Dev: Lee Deitesfeld
+#Change Log:
+#20190331LAD Created functions
+#--------------------------
+
+def fibonacci(n):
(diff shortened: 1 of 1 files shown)


students/lee_deitesfeld/series.py
```diff
@@ -0,0 +1,80 @@
...
+    while b < 10000000000000000000000:
+        a, b = b, a+b
+        fib_list.append(a)
+    #print(fib_list)
+
+    #presets value with index zero, else calculates value
+    if n == 0:
+        print('0')
+    else:
+        print(fib_list[n-1])
```
natasha-aleksandrova: your function is supposed to return nth value - right now your functions don't return any value.

also, you don't want to continue looping once you found the nth value.

PR #69, opened 2019-04-01
Adding the lesson 03 exercise files

rfelts:


rfelts:
Adding the mailroom assignment to the PR

students/rfelts/lesson03/list_lab.py
```diff
@@ -0,0 +1,135 @@
...
+    fruit_list = ["Pineapples"] + fruit_list[:]
+    print("Adding Pineapples to the beginning of the list using +:\n", fruit_list, "\n")
+
+    fruit_list.insert(0, "Cherries")
+    print("Adding Cherries to the beginning of the list using insert:\n", fruit_list, "\n")
+
+    # Display all the fruits that begin with “P”
+    print("Printing all fruits in the list that start with the letter P:\n")
+    for x in fruit_list:
+        if x[:1] == "P":
```
charlesrobison: Nice use of slicing here.  Recommend using lower() as well for case sensitivity

students/rfelts/lesson03/mailroom.py
```diff
@@ -0,0 +1,144 @@
...
+
+def prompt_user():
+    """
+    Prompt the user to choose from a menu of 3 actions: “Send a Thank You”, “Create a Report” or “quit”.
+    """
+    input_string = "\nPlease enter one of the following options: " + ', '.join(menu_options) + ": "
+    response = input(input_string)
+
+    while not "quit" == response.lower():
+        if response.lower() == "send thank you":
```
charlesrobison: You can use slicing here as well - so users don't have to input entire phrase

PR #72, opened 2019-04-02
corrects dict lookup code, adds switch dict feature

jkrecek3:


students/jim_krecek/lesson04/mailroom_p2.py
```diff
@@ -10,15 +10,13 @@ def send_ty():
...
-    
-        if askname != d:
+        if data.get(askname):
+            askmoney = input("\nHow much did this person donate?\n>>> ")
+            data[askname].append(int(askmoney))
+            emailtext = "\nDear {},\n\nThank you for your generous donation of ${}.\n\nSincerely, Jim".format(askname,askmoney)
+            print(emailtext)
+
+        else:
             askmoney = input("\nHow much did this person donate?\n>>> ")
```
natasha-aleksandrova: Much better! 

Something I would recommend improving maybe in the next mailroom assignment is some of the duplication.
It's one of the best practices to remove code that is obviously duplicated. Right lines 13 through 24 share a lot of duplicate code and can be fairly easily addressed 🙂 

https://refactoring.guru/smells/duplicate-code

students/jim_krecek/lesson04/mailroom_p2.py
```diff
@@ -44,16 +42,17 @@ def send_all():
     print("\nLetters have been printed and are saved in the current directory")
     
 def main():
+    main_menu = {'1': send_ty,'2': create_report,'3':send_all,'4':False}
```
natasha-aleksandrova: map `4` to quit program - the advantage of using switch_dict is how simple your main loop becomes

```
def quit_program():
    print("bye")
    sys.exit()

    main_menu = {'1': send_ty,'2': create_report,'3':send_all,'4': quit_program}
```

students/jim_krecek/lesson04/mailroom_p2.py
```diff
@@ -44,16 +42,17 @@ def send_all():
...
-        if response == "1" or response.lower() == "send a thank you":
-            send_ty()
-        elif response == '2' or response.lower() == "create a report":
-            create_report()
-        elif response == '3' or response.lower() == "send letters to all donors":
-            send_all()
-        elif response == '4' or response.lower() == "quit":
-            break
+        menu_function = main_menu.get(response)
+        if response == '1' or response == '2' or response == '3' or response == '4':
```
natasha-aleksandrova: `if response in main_menu`

students/jim_krecek/lesson04/mailroom_p2.py
```diff
@@ -44,16 +42,17 @@ def send_all():
...
+            if response == '1' or response == '2' or response == '3':
+                menu_function()
+            else:
+                break
```
natasha-aleksandrova: see comment above on how to simplify this part

students/jim_krecek/lesson04/mailroom_p2.py
jkrecek3: touche! I definitely see some lines that could be moved in that block of code. thanks for pointing that out.

PR #86, opened 2019-04-10
Committing all of session 03

humbertojrg:


students/humberto_gonzalez/session03/mailroom.py
```diff
@@ -0,0 +1,108 @@
...
+    print('-'*64)
+    formatter = "{:20}   ${:>10}   {:>10}   ${:>10}"
+    for donor_info in donor_db:
+        donor = donor_info[0]
+        donations = donor_info[1]
+        total = round(sum(donations),2)
+        num = len(donations)
+        average = round(total/num,2)
+        print(formatter.format(donor,total,num,average))
+    main()
```
natasha-aleksandrova: missing sorting feature here

PR #88, opened 2019-04-11
Lehning final assignment

KLehning:
Pull request for Lehning final assignment


students/kyle_lehning/Lesson09/donor_models.py
```diff
@@ -0,0 +1,133 @@
...
+        """returns a string saying what class is being used and the objects name"""
+        return (self.__class__.__name__ + " with {} donors").format(len(self.donor_list))
+
+    def add_new_donor(self, donor_name):
+        """Create a donor object based off the donor_name and add that object to the donor_list"""
+        self.donor_list.append(Donor(donor_name))
+
+    def add_existing_donor(self, don_name, don_total, don_number):
+        """Create a donor based off Donor's from_existing() function and add that donor to the donor_list"""
+        self.donor_list.append(Donor.from_existing(don_name, don_total, don_number))
```
natasha-aleksandrova: I don't really understand the purpose of this function here - in both cases you're adding a new donor object.

students/kyle_lehning/Lesson09/cli_main.py
```diff
@@ -0,0 +1,140 @@
...
+                donation_amount = round(float(donation_amount), 2)
+            except ValueError:
+                print("Please provide a valid amount without $ symbol")
+            else:
+                if donation_amount > 0:
+                    break
+                else:
+                    print("Please provide a valid amount without $ symbol")
+        current_donor.new_donation(donation_amount)
+        print(current_donor.generate_recent_thanks())
```
natasha-aleksandrova: I was a bit confused reading the code in `thank_you` because it only called `write_letter` and couldn't find the code that adds / updates donors - this was because I expected `write_letter` to write a letter and nothing else.

This function ideally should be renamed or broken down into two functions that do appropriate things.

PR #99, opened 2019-04-19
lesson 3 labs

franjaku:
Adding slicing,list and string labs.

### DIFF
--- a/students/franjaku/Lesson_3/list_lab.py
+++ b/students/franjaku/Lesson_3/list_lab.py
@@ -1,0 +1,96 @@
+#!/usr/bin/env python3
+
+# Series 1
+print('===========================')
+print('Series 1')
+Fruits = ["Apples","Pears","Oranges","Peaches"]
+
+print('Fruits = {}'.format(Fruits))
+
+New_Fruit = input("Add another fruit: ")
+
+Fruits.append(New_Fruit.title())
+
+print('Fruits = {}'.format(Fruits))
+
+NumInvalid = 1
+while NumInvalid:
+    number = input("Type a number between 1 and {:d}: ".format(len(Fruits)))
+    number = int(number)
+    if number > 0 and number < (len(Fruits)+1):
+        NumInvalid=0
+
+print("{:d} {}".format(number,Fruits[number-1]))
+
+New_Fruit2 = input("Add another fruit: ")
+
+Fruits = Fruits + [New_Fruit2.title()]
+
+print('Fruits = {}'.format(Fruits))
+
+New_Fruit3 = input("Add another fruit: ")
+
+Fruits.insert(0,New_Fruit3.title())
+
+print('Fruits = {}'.format(Fruits))
+
+for fruit in Fruits:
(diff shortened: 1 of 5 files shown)


students/franjaku/Lesson_3/list_lab.py
```diff
@@ -0,0 +1,96 @@
+#!/usr/bin/env python3
+
+# Series 1
+print('===========================')
+print('Series 1')
+Fruits = ["Apples","Pears","Oranges","Peaches"]
```
natasha-aleksandrova: var names should be snake_case -> `fruits` `new_fruit`

students/franjaku/Lesson_3/list_lab.py
```diff
@@ -0,0 +1,96 @@
...
+print('Fruits = {}'.format(Fruits))
+
+New_Fruit = input("Add another fruit: ")
+
+Fruits.append(New_Fruit.title())
+
+print('Fruits = {}'.format(Fruits))
+
+NumInvalid = 1
+while NumInvalid:
```
natasha-aleksandrova: better to use `while True` here and use `break` when you want to stop the loop

students/franjaku/Lesson_3/mailroom.py
```diff
@@ -0,0 +1,166 @@
...
+            donor_found=1
+
+            #update donor data
+            total_donation_new= donor[1] + donation_amount
+            number_donations_new = donor[2] + 1
+            average_donation_new = total_donation_new/number_donations_new
+            donor_updated = (name,total_donation_new,number_donations_new,
+                            average_donation_new)
+
+            database[idx] = donor_updated
```
natasha-aleksandrova: you can use an `else` statement with for loop instead of `donor_found` check to simplify your code.

students/franjaku/Lesson_3/mailroom.py
```diff
@@ -0,0 +1,166 @@
...
+
+    database = sorted(database, key=sort_key, reverse=True)
+    print('-----Donation Report-----')
+    print('\n{:<15} | {:>14} | {:>11} | {:>16}'.format('Donor Name', 'Total Donation', '# donations','Average Donation'))
+    print('-'*(15+14+11+16+10))
+    for donor in database:
+        print(f"{donor[0]:<15} | ${donor[1]:>13.2f} | {donor[2]:^11} | ${donor[3]:>15.2f}")
+
+    print('\n')
+    return None
```
natasha-aleksandrova: you don't need to have a return statement for this function

students/franjaku/Lesson_3/mailroom.py
```diff
@@ -0,0 +1,166 @@
...
+            print('User option 2')
+            create_report(database)
+        elif UserAction == '3':
+            #3 quit
+            print('Goodbye!')
+            return False
+        else:
+            print('Not a valid option...\n')
+
+    return None
```
natasha-aleksandrova: no need for return statement

PR #101, opened 2019-04-20
Added lesson04

Chieu1:
Added dict_lab.py and mailroom_part2.py on lesson04

### DIFF
--- a/students/chieu_quach/lesson03/mailroom.py
+++ b/students/chieu_quach/lesson03/mailroom.py
@@ -3,6 +3,7 @@
 # Assignment -  Lesson 3
 # Exercise   -  Mailroom Part 1
 
+# mailroom part 1
 import sys
 global donation_list, len_donation_list
 donation_list = {}
--- a/students/chieu_quach/lesson04/dict_lab.py
+++ b/students/chieu_quach/lesson04/dict_lab.py
@@ -1,0 +1,102 @@
+
+# Author      :  Chieu Quach
+# Assignment  :  Lesson 4
+# Exercise    :  Dictionary and Set Lab
+
+# first variable = key : second variable = value
+# Example = "Name" is key : "Chris" is value
+
+def dictionaries():
+
+
+    dict_name = {'Name': 'Chris', 'City': 'Seattle', 'Cake': 'Chocolate'}
+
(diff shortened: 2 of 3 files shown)


students/chieu_quach/lesson03/mailroom.py
```diff
@@ -3,6 +3,7 @@
 # Assignment -  Lesson 3
 # Exercise   -  Mailroom Part 1
 
+# mailroom part 1
```
charlesrobison: @Chieu1 - I'm unable to do inline comments on this file due to the issues with submitting this for review.  But, overall it works well.  There are a few PEP8 issues you'll want to address, like extra blank lines.  Take a look at installing a linter on you text editor to catch those.  Also, your import statements should all be at the top of the file.

students/chieu_quach/lesson04/dict_lab.py
```diff
@@ -0,0 +1,102 @@
+
+# Author      :  Chieu Quach
+# Assignment  :  Lesson 4
+# Exercise    :  Dictionary and Set Lab
+
+# first variable = key : second variable = value
+# Example = "Name" is key : "Chris" is value
+
+def dictionaries():
+
```
charlesrobison: Remove these blank lines

students/chieu_quach/lesson04/mailroom_part2.py
```diff
@@ -0,0 +1,133 @@
...
+       response = input(prompt)
+       response = int(response)
+       switch[response]()
+       
+     
+
+# Main function
+if __name__ == "__main__": 
+  
+   main()
```
charlesrobison: @Chieu1 - this version isn't working as expected.  When sending a thank you to a single donor, if I enter a name, it breaks out of the program.  Also, the program should continue to loop until the user exits.  Also, there are a lot of PEP8 issues - mainly spacing.  All indentations should be 4 spaces in Python.

Chieu1:
Charles,

I will try to install Linter for text editor.
As you mentioned on the mailroom for single donor, it will work if you enter full name. I will fix it so you can just emter the first name and it will work. I will create a loop  to the menu.

Thanks,

Chieu

PR #110, opened 2019-04-23
Added a README file

trvvss:
Following the example from https://help.github.com/en/articles/creating-a-pull-request-from-a-fork

### DIFF
--- a/students/travis_nelson/lesson02/fizzBuzz.py
+++ b/students/travis_nelson/lesson02/fizzBuzz.py
@@ -1,0 +1,38 @@
+
+def fizzBuzz1(fizzUntil=100):
+    """This function will fizzBuzz from 1 through user-input integer"""
+    for i in range(1,fizzUntil+1):
+        if i % 3 == 0:
+            if i % 5 == 0:
+                print("FizzBuzz")
+            else: 
+                print("Fizz")
+        elif i % 5 == 0:
+            print("Buzz")
+        else:
+            print(i)
+
+def fizzBuzz2(fizzUntil=100):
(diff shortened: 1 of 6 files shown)


students/travis_nelson/lesson02/fizzBuzz.py
```diff
@@ -0,0 +1,38 @@
+
+def fizzBuzz1(fizzUntil=100):
```
natasha-aleksandrova: function names should follow snake_case convention -> `fizz_buzz1`, same with `fizzUntil` 

you can read pep8 style guide or use an online linter to catch violations - however, I usually recommend Pycharm as everything is configured out of the box and it will flag any violations as you write code.

students/travis_nelson/lesson02/gridPrinter.py
```diff
@@ -0,0 +1,107 @@
...
+        print('+', end=' ')
+        print('- '*lengthOfCell, end= '')
+        print('+')
+    def grid_row(lengthOfCell):
+        """Prints a row of vertical grid borders"""
+        print('|', end = ' ')
+        print(' '*(lengthOfCell*2), end= '')
+        print('|', end = ' ')
+        print(' '*(lengthOfCell*2), end= '')
+        print('|')
```
natasha-aleksandrova: although inner functions have its uses be careful about over-using them. functions here are best when defined as regular functions.

students/travis_nelson/lesson02/fizzBuzz.py
trvvss: Understood! I'm looking at linters now (in VS Code), and will consider switching over to Pycharm. I'll definitely be more attentive to this in the future.

PR #111, opened 2019-04-23
Pkoleyni adding Lesson2 Grid and fizz_buzz

pkoleyni:


students/pkoleyni/Lesson02/Grid-01.py
```diff
@@ -0,0 +1,41 @@
...
+    """This function prints the final result
+    it will take n as number of rows and will
+    print rows of 2 block grid"""
+    if n == 0:
+        return print_error()
+    for i in range(0, n):
+        print_two_blocks()
+    print(line_style_1())
+
+print_result(2)
```
charlesrobison: works well!

students/pkoleyni/Lesson02/Grid-02.py
```diff
@@ -0,0 +1,51 @@
...
+    for i in range(0,n):
+        if i ==n//2:
+            line_style_1(n)
+        else:
+            line_style_2(n)
+    line_style_1(n)
+
+
+if __name__ == "__main__":
+    grid(9)
```
charlesrobison: Nice!

students/pkoleyni/Lesson02/Grid_03.py
```diff
@@ -0,0 +1,41 @@
...
+    line_style_1(n,m)
+    for i in range(0,n):
+        for i in range(0, m):
+            line_style_2(n, m)
+        line_style_1(n, m)
+
+
+
+if __name__ == "__main__":
+    result(3, 3)
```
charlesrobison: Good job on these!

students/pkoleyni/Lesson02/fizz_Buzz.py
```diff
@@ -0,0 +1,20 @@
...
+            print('FizzBuzz')
+        elif multiples_of_num(num,3):
+            print('Fizz')
+        elif multiples_of_num(num,5):
+            print('Buzz')
+        else:
+            print(num)
+
+
+fizz_buzz()
```
charlesrobison: Interesting solution!  By the way, you can use def fizz_buzz(n): and then ask for num in range(1, n) so the user can pass in any number.

students/pkoleyni/Lesson02/fizz_Buzz.py
```diff
@@ -0,0 +1,20 @@
+
+def multiples_of_num(num1,num2):
+    if num1%num2 ==0:
+        return True
+    else:
+        return False
+
+def fizz_buzz():
+    for num in range(1,101):
```
charlesrobison: Per PEP8 you'll want to add a space between these variables - for num in range(1, 101):

Also, for the ones below.

students/pkoleyni/Lesson02/Grid-01.py
pkoleyni: Thank you

students/pkoleyni/Lesson02/Grid-02.py
pkoleyni: Thank you

students/pkoleyni/Lesson02/Grid_03.py
pkoleyni: Thank you

pkoleyni:
I added space between variables based on PEP8, also modified fizz_buzz() so the user can pass in any number

PR #112, opened 2019-04-24
Adding Lesson09 files

lombardic:
The following files for Lesson09 are ready for review:
-cli_main.py
-donor_models.py
-test_donor_models.py

### DIFF
--- a/students/chris_lombardi/Lesson09/donor_models.py
+++ b/students/chris_lombardi/Lesson09/donor_models.py
@@ -1,0 +1,103 @@
+class Donor:
+    """
+    A class for a donor object
+    Stores the name of the donor and donation history data.
+    """
+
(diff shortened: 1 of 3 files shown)


students/chris_lombardi/Lesson09/donor_models.py
```diff
@@ -0,0 +1,103 @@
...
+    A class containing information of a list of Donor objects
+    """
+
+    def __init__(self):
+        """Create a collection of donor objects."""
+        self.list_donors = []
+
+    def list_donors(self):
+        """Return a list of all donors in the collection."""
+        return list_donors
```
natasha-aleksandrova: looks like this could be best as a `property`? otherwise both your variable and method names are the same name which is confusing.

PR #113, opened 2019-04-25
Made suggested updates to the lesson05 mailroom assignment file

rfelts:


natasha-aleksandrova:
@rfelts looks much better!

PR #115, opened 2019-04-26
Finished lesson03 exercises

gregdevore:


students/gregdevore/lesson03/list_lab.py
```diff
@@ -0,0 +1,130 @@
...
+printseries(4)
+
+# Recover original list
+fruits = original_fruits[:]
+
+# Display current list of fruits
+print('Current list of fruits:',fruits)
+
+# Create list of fruit but with each fruit reversed
+reverse_fruits = [fruit[::-1] for fruit in fruits[:]]
```
natasha-aleksandrova: nice use of comprehension! note that you shouldn't need to create a copy via `fruits[:]`  as comprehension will create a new list for you

PR #120, opened 2019-04-28
Update to lesson 5

franjaku:
Added another exception. 

Check mail_room() and get_donation_amount() for try/except blocks.

Struggle to find places in my current code to use comprehension without getting side effects. However current code isn't pretty and needs to be reworked. May be able to find some good places to use comprehension after that. Very possible I'm missing something. If so please let me know.


natasha-aleksandrova:
@franjaku your additions for try/except fine and there are no obvious places for comprehensions which is ok.

There are some areas in your code that could use improvements to be more pythonic - you can address those in your next revision of the mailroom (lesson 6).

* line 174: remove `return None` it's a "void" function
* replace any instances like `Email` with lower case words separated by underscores (`email`) - this is pep8 style violation
* you are returning `database` in `add_new_donation` and `add_new_donor` - don't return those, you are modifying your database in place
* remove `outfile.close` - `with` statement closes the file for you already.
* on line 86 you are calling `return get_donation_amount()` on error - use a while True loop there, and continue asking for valid amount.

let me know if you have any questions on my above comments.

franjaku:
Awesome I appreciate the feedback thank you!

PR #125, opened 2019-05-03
adding Slicing_Lab.py

pkoleyni:
I need feedback on how to improve these functions, make my code better and smaller 
Thank you

### DIFF
--- a/students/pkoleyni/Lesson03/Slicing_Lab.py
+++ b/students/pkoleyni/Lesson03/Slicing_Lab.py
@@ -1,0 +1,66 @@
+def exchange_first_last(seq):
+    """
+    :param seq: any python seq, like String,List,Tuple
+    :return:
+    """
+    first = seq[0]
+    last = seq[-1]
+    middle = seq[1:-1]
+    if type(seq) == str:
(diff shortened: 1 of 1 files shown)


students/pkoleyni/Lesson03/Slicing_Lab.py
```diff
@@ -0,0 +1,66 @@
...
+
+def every_other_item(seq):
+    new_list=[]
+    new_string = ''
+    for i in range(0,len(seq),2):
+        new_list.append(seq[i])
+    if type(seq) == list:
+        return new_list
+    elif type(seq) == str:
+        return new_string.join(new_list)
```
natasha-aleksandrova: the above two functions are the only ones that have issues, below functions look fine!

the issue with above function is you're not using slicing exclusively - you are mixing indexing and loops as well.

Try to use slicing only (hint: this should be only one line of code) - if you still get stuck comment here and I will help out!

natasha-aleksandrova:
@pkoleyni see my above comment for feedback.

natasha-aleksandrova:
closing for now - re-open when you have fixed the first two functions.

PR #126, opened 2019-05-05
Fixed the first 2 functions of Slicing_lab.py

pkoleyni:


students/pkoleyni/Lesson03/Slicing_Lab.py
```diff
@@ -0,0 +1,49 @@
+def exchange_first_last(seq):
+    """
+    :param seq: any python seq, like String,List,Tuple
+    :return:
+    """
+    return (seq[-1:]+seq[1:-1]+seq[0])
```
natasha-aleksandrova: this would give you an error with a list or a tuple:
```
>>> seq = [1, 2, 3, 4, 5, 6]
>>> seq[-1:]+seq[1:-1]+seq[0]
  File "<stdin>", line 1, in <module>
TypeError: can only concatenate list (not "int") to list
```

remember that you should **_only_** use slicing; `seq[0]` is called indexing

students/pkoleyni/Lesson03/Slicing_Lab.py
```diff
@@ -0,0 +1,49 @@
...
+
+def every_other_item(seq):
+    return (seq[0::2])
```
natasha-aleksandrova: you don't need parens with return statement, it's not a function call -> `return seq[0::2]`

also, it can just be `return seq[::2]` as 0 is a default value

students/pkoleyni/Lesson03/list_lab.py
```diff
@@ -0,0 +1,78 @@
...
+    seq.append(input('Enter name of a fruit >'))
+    return seq
+
+def find_by_index(seq):
+    user_input = input('Enter a number in range of 1 to {} >'.format(len(seq) + 1))
+    return ('You have entered {} which is {} from the fruit_list'.format(user_input,seq[int(user_input)-1]) )
+
+def find_by_letter(seq, letter ='P'):
+    for item in fruit_list:
+        if letter in item:
```
natasha-aleksandrova: your logic should ensure that it checks that the string starts with P, not that P is somewhere in the word.

PR #127, opened 2019-05-06
Added two revised programs Mailroom 2 from lesson 4 and exceptions Exercise lesson 5

Chieu1:


students/chieu_quach/lesson05/except_exercise.py
```diff
@@ -0,0 +1,108 @@
...
+# parameters. (pun intended)
+
+langs = ['java', 'c', 'python']
+
+try:
+    more_joke = more_fun(langs[0])
+except IndexError:
+    pass
+else:
+    pass
```
natasha-aleksandrova: within above try/except apply the requirements

> If there are no exceptions, call the more_fun function with the last
> language in the list
> 
> Finally, while still in the try/except block and regardless of whether
> there were any exceptions, call the function last_fun with no
> parameters.

remove all of the code below this comment

natasha-aleksandrova: in this try/except block please apply the requirements listed:

> If there are no exceptions, call the more_fun function with the last
> language in the list
> 
> Finally, while still in the try/except block and regardless of whether
> there were any exceptions, call the function last_fun with no
> parameters. (pun intended)

remove all of the code below

natasha-aleksandrova:
your mailroom is still missing the sorting feature and "list" feature - we can discuss your mailroom in office hours today

PR #130, opened 2019-05-07
adds lesson 6 files

jkrecek3:
trying to figure out why when I call 'pytest -vv,' letter_test() won't show up as a test that has passed or failed. I get the output '3 passed in 0.05 seconds' even though I have 4 tests.

### DIFF
--- a/students/jim_krecek/lesson06/test_mailroom_p4.py
+++ b/students/jim_krecek/lesson06/test_mailroom_p4.py
@@ -1,0 +1,57 @@
+import mailroom_p4
+from mailroom_p4 import *
+import os
+import pytest
+
+def test_create_report():
+    report = """Donor Name               | Total Given | Num Gifts | Average Gift
+-----------------------------------------------------------------
+Pat Carrier               $     106.00           2  $       53.00
+Jim Krecek                $      93.00           3  $       31.00
+Nate Secinaro             $      65.00           3  $       21.67
+Zac Fisher                $      21.00           3  $        7.00
+Jess Reid                 $       3.00           1  $        3.00"""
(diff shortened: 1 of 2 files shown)


natasha-aleksandrova:
@jkrecek3 all of your test functions need to start with `test_` in order for pytest to pick them up

students/jim_krecek/lesson06/test_mailroom_p4.py
```diff
@@ -0,0 +1,57 @@
...
+
+Thank you for your generous donation of $39.
+
+Sincerely, Jim"""
+
+def letter_test():
+    # test the amount of letters generated
+    data = {"Nate Secinaro":[12,44,9],"Jess Reid":[3],"Pat Carrier":[65,41]}
+    for k,v in data.items:
+        mailroom_p4.write_letters(k,v)
```
natasha-aleksandrova: since you're only testing `write_letters()` function, you shouldn't actually call it in a loop.

you can execute it once with a donor and assert that a single file was created.

students/jim_krecek/lesson06/test_mailroom_p4.py
```diff
@@ -0,0 +1,57 @@
...
+    
+    # tests update of current donor
+    assert mailroom_p4.thanks(data, 'Nate Secinaro', 39) == """
+Dear Nate Secinaro,
+
+Thank you for your generous donation of $39.
+
+Sincerely, Jim"""
+
+def letter_test():
```
natasha-aleksandrova: per my other comment this should be `test_letter`

natasha-aleksandrova:
don't forget to submit on the course website.

PR #134, opened 2019-05-07
uploaded lesson04 assignments

deitesla:


students/lee_deitesfeld/lesson04/mailroom.py
```diff
@@ -0,0 +1,139 @@
...
+                {"name": "Joe Grimes",
+                 "donations": [102.00, 45.00, 70.00]
+                },
+                {"name": "Andrea Funk",
+                 "donations": [21.00, 21.00]
+                },
+                {"name": "Mickey Mouse",
+                 "donations": [500.00, 250.00, 100.00]
+                }
+                ]}
```
natasha-aleksandrova: even though this is a dictionary, it isn't exactly the dictionary ideal for this data structure.

The benefit of using dict for donors data structure is its efficiency in look ups by key - in our case donor's name makes a perfect key since we are usually searching donors by name to see if donations should be updated or added.

In the above structure you are still dealing with a list of donors thus not really satisfying the requirement.

students/lee_deitesfeld/lesson04/mailroom.py
```diff
@@ -0,0 +1,139 @@
...
+        response = input(prompt)
+        if main_dispatch[response]() == "exit menu":
+            break
+
+def sub_menu():
+    menu_selection(sub_prompt, sub_dispatch)
+
+def quit():
+    print("Quitting this menu")
+    return "exit menu"
```
natasha-aleksandrova: you can you use here `sys.exit()` and then you won't need to check for "exit menu" string.

students/lee_deitesfeld/lesson04/mailroom.py
```diff
@@ -0,0 +1,139 @@
...
+        x['sum'] = int_ttl_amt
+    dic_donors['donor_db'].sort(key=itemgetter('sum'), reverse = True) #sort on key='sum'
+    for x in dic_donors['donor_db']:
+        ttl_amt_2dec = '%.2f' % x['sum'] #limits total to two decimal points
+        x['sum'] = ttl_amt_2dec
+
+    #removes history of donations, since it's not in final report
+    for x in dic_donors['donor_db']:
+        del x['donations']
+
```
natasha-aleksandrova: you have 6 loops above when it seems you can have just one loop since you're just looping over the same data structure?

also, I would avoid modifying the main data structure in place - for you report you should create a new data structure to hold your report data.

students/lee_deitesfeld/lesson04/mailroom.py
```diff
@@ -0,0 +1,139 @@
...
+            continue
+    if donor_question not in item["name"]:
+        dic_new_donor= {}
+        dic_new_donor["name"] = donor_question
+        dic_donors['donor_db'].append(dic_new_donor)
+        donation_amount = input("Please enter donation amount for donor: ")
+        dic_new_donor['donations'] = donation_amount
+        email = 'Dear {name}, Thank you for your generous donation of {donations}. We hope you will consider donating again!'.format(**dic_new_donor)
+        print(email)
+        dic_new_donor['donations'] = [float(donation_amount)]
```
natasha-aleksandrova: switch your main data structure to a dictionary and the code in this function can be reduced in half.

students/lee_deitesfeld/lesson04/trigrams.py
```diff
@@ -0,0 +1,79 @@
+#!/usr/bin/env python3
+import sys
+import random
+
+def read_in_data(filename):
+    '''Reads the lines of the file'''
```
natasha-aleksandrova: docstrings use triple quoted strings:
```
def read_in_data(filename):
    """Reads the lines of the file"""
```

students/lee_deitesfeld/lesson04/mailroom.py
natasha-aleksandrova: so a better data structure here would be:
```
donors = {"Jeff Bezos": [993, 999, 232], "Tom Jerry": [33, 244]}
```

PR #137, opened 2019-05-08
added revised mailroom 2 from lesson04 and except_excercise from lesson05

Chieu1:


students/chieu_quach/lesson04/mailroom_part2.py
```diff
@@ -43,15 +41,16 @@
       "\n\n"
       )
 
+from operator import *
```
charlesrobison: Make sure import statements are at the top of the file.  Also, you very rarely want to use the star wildcard when importing.

students/chieu_quach/lesson04/mailroom_part2.py
```diff
@@ -43,15 +41,16 @@
...
 def send_thankyou():
       
   name_found = ""
```
charlesrobison: Make sure you have 4 spaces for indentions throughout your file.

Chieu1:
Charles,

Thank you. I will indent 4 lines next time when I submit my program.

Thanks,

Chieu

PR #138, opened 2019-05-09
Slicing n Dicing 

trvvss:
Completed the exercise from here: https://uwpce-pythoncert.github.io/PythonCertDevel/exercises/slicing.html


trvvss:
I also noticed that when I pushed the commits, they went through with my other github account? Maybe it relates to this: https://stackoverflow.com/questions/21615431/git-pushes-with-wrong-user-from-terminal

Do you have any suggestions? I don't think it's a big deal, but I'm curious as to how to fix it.

charlesrobison:
> I also noticed that when I pushed the commits, they went through with my other github account? Maybe it relates to this: https://stackoverflow.com/questions/21615431/git-pushes-with-wrong-user-from-terminal
> 
> Do you have any suggestions? I don't think it's a big deal, but I'm curious as to how to fix it.

Hi @trvvss - if you have more than one account, then you'll want to set your config for each repo to right one.  Take a look at the config section in the course site and let me know if you need assistance.

students/travis_nelson/lesson03/list_lab.py
```diff
@@ -0,0 +1,97 @@
...
+    Add another fruit to the beginning of the list using "+" and display list.
+    Add another fruit to the beginning of the list using insert().
+    Display all fruits that begin with "P" using a for loop.
+    """
+    fruit = ["Apples", "Pears", "Oranges", "Peaches"]
+    print(fruit)
+    received_fruit = input("Another fruit, please. ---> ").title()
+    fruit.append(received_fruit)
+    print(fruit)
+    received_number = input("What number fruit would you like to know? ---> ")
```
charlesrobison: You can use len() here to give the user the length of the list

students/travis_nelson/lesson03/list_lab.py
```diff
@@ -0,0 +1,97 @@
...
+    print("Fruit number {} is ".format(received_number) +
+          fruit[int(received_number)-1])
+    fruit = ["Durian"] + fruit
+    print(fruit)
+    fruit.insert(1, "Cantaloupe")
+    print(fruit)
+    for i in fruit:
+        if i[0].upper() == 'P':
```
charlesrobison: Nice use of upper()!

students/travis_nelson/lesson03/list_lab.py
```diff
@@ -0,0 +1,97 @@
...
+
+def series_3():
+    """Using the list created in series_1():
+    Ask the user if they like each fruit in the list.
+    Delete all fruit the user doesn't like.
+    For any answer that isn't 'yes' or 'no', prompt user for a new answer.
+    Display the list.
+    """
+    fruit = series_1()
+    for i in fruit[:]:
```
charlesrobison: Nice!

students/travis_nelson/lesson03/list_lab.py
```diff
@@ -0,0 +1,97 @@
...
+    """
+    fruit = series_1()
+    reversed_fruit = []
+    for i in fruit[:]:
+        reversed_fruit.append(reverse_items(i))
+    del fruit[-1:]
+    print(fruit, reversed_fruit)
+
+
+def reverse_items(sequence):
```
charlesrobison: Recommend moving this up since series_4 calls it

students/travis_nelson/lesson03/slicing_lab.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+def exchange_first_last(sequence):
+    """Returns a passed sequence with its first and last items exchanged"""
+    sequence_copy = sequence[:]
+    first_item = sequence_copy[:1]
+    last_item = sequence_copy[-1:]
+    sequence_copy_sans_first_and_last = sequence_copy[1:len(sequence_copy)-1]
+    a_new_sequence = last_item + sequence_copy_sans_first_and_last + first_item
+    return a_new_sequence
```
natasha-aleksandrova: you can simplify this to `return seq[-1:] + seq[1:-1] + seq[:1]`

students/travis_nelson/lesson03/slicing_lab.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+def remove_every_other(sequence):
+    """Returns a passed sequence with every other item removed"""
+    sequence_copy = sequence[:]
+    a_new_sequence = sequence_copy[::2]
+    return a_new_sequence
```
natasha-aleksandrova: same here `seq[::2]`

students/travis_nelson/lesson03/slicing_lab.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+def remove_first_last_four(sequence):
+    """Returns a passed sequence with the first and last 4 items removed,"""
+    """then every other element remaining"""
+    sequence_copy = sequence[:]
+    a_new_sequence = sequence_copy[4:-4]
+    return a_new_sequence
```
natasha-aleksandrova: `seq[4:-4:2]` this should also remove every other element

students/travis_nelson/lesson03/slicing_lab.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+def reverse_sequence(sequence):
+    """Returns a passed sequence with the elements reversed with slicing"""
+    sequence_copy = sequence[:]
+    a_new_sequence = sequence_copy[::-1]
+    return a_new_sequence
```
natasha-aleksandrova: no need to copy it - the slice will create a new list

PR #143, opened 2019-05-11
Finished mailroom part 2 and trigrams assignments

gregdevore:


students/gregdevore/lesson04/mailroom.py
```diff
@@ -0,0 +1,129 @@
...
+                return
+            else: # Otherwise, convert donation amount to float
+                amount = float(amount)
+            # Add donation to database
+            donations.append(amount)
+            donors[name] = donations
+            # Generate & print email, return to main program
+            email = generate_email(name, amount, sum(donations))
+            print(email)
+            return
```
natasha-aleksandrova: no need for `return` here

students/gregdevore/lesson04/mailroom.py
```diff
@@ -0,0 +1,129 @@
...
+    while True:
+        response = input(prompt)
+        if response in response_dict:
+            response_dict[response]()
+        else:
+            print('Not a valid option! Please try again.')
+
+if __name__ == "__main__":
+    # Driver for main function
+    main()
```
natasha-aleksandrova: excellent work on this version of mailroom!

PR #148, opened 2019-05-14
Adding mail room assignment

humbertojrg:


students/humberto_gonzalez/session05/mailroom.py
```diff
@@ -0,0 +1,143 @@
...
+
+def getKey(item):
+    return item[1]
+
+def ordered_db(db):
+    keys = list(db.keys())
+    temp = []
+    for name in keys:
+        donation_total = sum(db[name])
+        temp.append((name,donation_total))
```
natasha-aleksandrova: simplify this:
```
def ordered_db():
    for donor, donations in donor_db.items():
        temp.append((donor, sum(donations)))
```


however, given that you should use comprehensions in this lesson:
```
name_and_donations_sum = [name, sum[donations] for name, donations in donor_db.items()]
```

students/humberto_gonzalez/session05/mailroom.py
```diff
@@ -0,0 +1,143 @@
...
+main_prompt = "\n".join(("Welcome to the Mailroom!",
+              "Please choose from below options:",
+              "1 - Send a Thank You",
+              "2 - Create a report",
+              "3 - Send letters to all donors",
+              "4 - Exit Program",
+              ">>> "))
+
+    
+
```
natasha-aleksandrova: spacing should be two lines, here and between functions.

students/humberto_gonzalez/session05/mailroom.py
```diff
@@ -0,0 +1,143 @@
...
+    formatter = '''Dear {},\n    Thank you for your generous donation of ${}. \n    Your donation will be put to great use. \n         Sincerely, \n          -The Organization'''
+    path = tempfile.gettempdir()
+    path = path + "/" + "Letters to Donors"
+    os.mkdir(path)
+    for donor in donor_db:
+        temp = path + "/" + donor.replace(' ','_') + '.txt'
+        with open(temp,'w') as file:
+            txt = formatter.format(donor,donor_db[donor][0])
+            file.write(txt)
+            file.close()
```
natasha-aleksandrova: no need to call `close()`, `with` takes care of that for you.

also, `file` is a reserved keyword.

students/humberto_gonzalez/session05/mailroom.py
```diff
@@ -0,0 +1,143 @@
...
+        # now redirect to feature functions based on the user selection
+        menu_options = {"1":send_thank_you,
+                        "2":create_report,
+                        "3":send_letters,
+                        "4":exit_program}
+        if response not in menu_options:
+            print()
+            print("Please select one of the available options")
+            print("You will be returned to the main menu")
+            main()
```
natasha-aleksandrova: no need to call main again, which can create sub loops, you can use `continue` instead.

students/humberto_gonzalez/session05/mailroom.py
```diff
@@ -0,0 +1,143 @@
...
+
+def send_thank_you():
+    '''Prompts user for a donor and donation amount, and then prints out a thank you email'''
+    donor_name = input('What is the full name of the donor you would like to thank? ')
+    if donor_name.lower() == "quit":
+        main()
+    if donor_name not in donors:
+        donor_db[donor_name] = []
+    try:
+        donation = input('What was the donation amount? ')
```
natasha-aleksandrova: this should not be inside of try/except

students/humberto_gonzalez/session05/mailroom.py
```diff
@@ -0,0 +1,143 @@
...
+
+def main():
+    while True:
+        response = input(main_prompt)  # continuously collect user selection
+        # now redirect to feature functions based on the user selection
+        menu_options = {"1":send_thank_you,
+                        "2":create_report,
+                        "3":send_letters,
+                        "4":exit_program}
+        if response not in menu_options:
```
natasha-aleksandrova: can you use exception handling here instead?

students/humberto_gonzalez/session05/mailroom.py
```diff
@@ -0,0 +1,143 @@
...
+        main()
+    donor_db[donor_name].append(float(donation))
+    print()
+    print()
+    print(f'Dear {donor_name},\n    Thank you for your generous donation of ${donation}')
+
+
+
+def create_report():
+    '''Creates a report of the current donor database'''
```
natasha-aleksandrova: docstrings use triple quoted strings `"""Here is a comment about function"""`

PR #150, opened 2019-05-14
Updated mailroom.py per recommendations

bryancoates:
Modified loop to leverage "else" per recommendation
Modified function names to be more descriptive per recommendation
Added sorting to report for missed requirement

### DIFF
--- a/students/bcoates/lesson03/mailroom.py
+++ b/students/bcoates/lesson03/mailroom.py
@@ -35,7 +35,7 @@ Please enter a number to select the action:
         else:
             print("Please select a valid number...")
 
-def prompt_donor_name():
+def prompt_and_add_donor():
     """ Prompt for a donor name, list if needed, add if not found """
 
     while True:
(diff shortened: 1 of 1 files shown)


students/bcoates/lesson03/mailroom.py
```diff
@@ -79,10 +76,10 @@ def prompt_donation_amount(full_name):
             float_amount = float(donation_amount)
 
             # Add the donation to the donor's list of donations
-            for donors in donor_history:
-                if donors[0] == full_name:
+            for donor in donor_history:
+                if donor[0] == full_name:
                     print("Adding new donation of {} from {}".format(donation_amount, full_name))
-                    donors[1].append(float_amount)
+                    donor[1].append(float_amount)
```
natasha-aleksandrova: I still think you can improve your program by working with donor item from the list instead of the name since you're looping through your structure all over again when have already found the donor previously (see last PR comments).

PR #156, opened 2019-05-16
finished part 2 of mailroom

brgalloway:
mailroom part 2 ready for grading


students/brgalloway/Lesson_4/mailroom_part_2.py
```diff
@@ -36,16 +36,15 @@ def menu_selection(prompt, dispatch_dict):
         response = input(prompt)
         if dispatch_dict[response]() == "quit":
             sys.exit()
```
natasha-aleksandrova: put `sys.exit()` into your `quit_app` function.

PR #160, opened 2019-05-18
Added sorting to mailroom and fixed the code based in the comments

pkoleyni:


students/pkoleyni/Lesson03/mailroom.py
```diff
@@ -121,17 +126,46 @@ def average_gifts(name):
...
+    report_list = [[] for i in range(donor_list_length)]
     print()
+
+
+    for i in range(0,len(list_donors())):
+        donor_name = list_donors()[i]
+        report_list[i].append(donor_name)
+        report_list[i].append(total_donation(donor_name))
+        report_list[i].append(num_gifts(donor_name))
+        report_list[i].append(average_gifts(donor_name))
```
natasha-aleksandrova: I'm a bit confused on the above code, here is what i see:
* first loop on line 143 via call to list_donors()
* second loop on 144 to generate empty list of lists
* another loop on line 148 to call list_donors() (now inside of a for loop, so called each list item)
* then 3 more additional loops via total_donation(), num_gifts(), average_gifts()


above is all very expensive, you are looping over your list way too many times.

This should all be done in a single loop:
```
for donor in donor_list:  # loop over items in a list, don't use range()
    name = donor[0]
    donations = donor[1:]
    num_gifts = len(donations)
    total_donations = sum(donations)
    avg_gifts = total_donation / num_gifts
    report_list.append([name, total_donations, num_gifts, avg_gifts)
```

please review above code and let me know if you have any questions
    report_list.append([


```

File: students/pkoleyni/lesson04/mailroom_2.py [hunk 2]
```diff
@@ -0,0 +1,180 @@
...
+    1. Send a thank you
+    2. Create a report
+    3. Send_letter_to_all
+    4. Quit
+
+    """
+    main_menu_dic = {'1': thank_you,
+                     '2': report,
+                     '3': write_letter_for_everyone,
+                     '4': quit
```
natasha-aleksandrova: map this to `sys.exit()` so you don't have to check return value of your function below

students/pkoleyni/lesson04/mailroom_2.py
```diff
@@ -0,0 +1,180 @@
...
+                  }
+
+
+def list_donors():
+    """
+    Return list of name of donors by getting keys for the donor_list_dic dictionary
+    :return:
+    """
+    donor_list = donor_list_dic.keys()
+    return donor_list
```
natasha-aleksandrova: I would recommend to not continue using this function.

For pythonic style, use `.items()` and loop over keys and values

students/pkoleyni/lesson04/mailroom_2.py
```diff
@@ -0,0 +1,180 @@
...
+    line3 = '\tIt will be put to very good use.\n'
+    line4 = '\tSincerely,\n\t     -The Team'
+
+    file = file_name(name)
+    with open(file, 'w') as f:
+        f.write(line1.format(name))
+        f.write(line2.format(name))
+        f.write(line3.format(name))
+        f.write(line4.format(name))
+        f.close()
```
natasha-aleksandrova: no need to call `close()`, `with` statement takes care of that for you

students/pkoleyni/Lesson03/mailroom.py
pkoleyni: Thank you for the comment, I see that i was doing many unnecessary steps.
I have update the code 
https://github.com/pkoleyni/SP_Online_PY210/commit/87b4491cbc03dbe053214a3a6ec71faf31d9b01c

natasha-aleksandrova: sounds great!

PR #168, opened 2019-05-21
Lesson08 Circle

ericae14:


students/erica_edwards/Lesson08/circle.py
```diff
@@ -0,0 +1,73 @@
...
+    def __init__(self, radius):
+        super().__init__(radius)
+
+    @property
+    def volume(self):
+        return 4/3 * math.pi * self.radius**3
+
+    @staticmethod
+    def from_diameter(diameter):
+        return Sphere(math.pi * diameter**2)
```
natasha-aleksandrova: From instructions:
```Check that the Sphere.from_diameter() alternate constructor actually creates a Sphere! (you DO NOT have to write a new classmethod for that!) – pretty cool, eh?```

File: students/erica_edwards/Lesson08/circle.py [hunk 2]
```diff
@@ -0,0 +1,73 @@
...
+    def diameter(self, diameter):
+        self.radius = diameter / 2
+
+    @property
+    def area(self):
+        return math.pi * self.radius**2
+
+    @staticmethod
+    def from_diameter(diameter):
+        return Circle(diameter/2)
```
natasha-aleksandrova: so this should be a `classmethod` not a staticmethod.

PR #170, opened 2019-05-21
Solutions to lesson 4

scottebetts:


students/scottebetts/lesson04/mailroom.py
```diff
@@ -0,0 +1,118 @@
+#!/usr/bin/env python
+from time import strftime
+
+donors = [['Tim Berners-Lee', 120000.00, 3, 40000.00],
+          ['Ada Lovelace', 600.00, 2, 300.00],
+          ['Marie Curie', 1000.00, 2, 500.00],
+          ['Nicola Tesla', 20000.00, 4, 5000.00],
+          ['Hans Holbein', 300.00, 3, 100.00]]
```
natasha-aleksandrova: you should be using a dictionary here

students/scottebetts/lesson04/mailroom.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+def menu_selection(prompt, dispatch_dict):
+    while True:
+        response = input(prompt)
+        if dispatch_dict[response]() == "exit menu":
+            break
```
natasha-aleksandrova: use `sys.exit()` to avoid having to check return string

students/scottebetts/lesson04/mailroom.py
```diff
@@ -0,0 +1,118 @@
...
+        file_date = strftime("%Y%m%d")
+        filename_format = "{}_{}.txt"
+        donor_name = donor[0].replace(' ', '_')
+        donor_file = filename_format.format(donor_name, file_date)
+        with open(donor_file, 'w+') as outfile:
+            outfile.write(
+                "Dear {},\n "
+                "Thanks for your recent gift of ${:>.2f}, "
+                "you great big gullible maroon.".format(donor[0], donor[1]))
+            outfile.close()
```
natasha-aleksandrova: you don't need to close the file, `with` statement does this for you.

PR #174, opened 2019-05-22
completed lesson08 files

jkrecek3:


students/jim_krecek/lesson08/test_circle.py
```diff
@@ -0,0 +1,79 @@
+#test_circle
+
+import pytest
+
+from circle import *
```
natasha-aleksandrova: avoid using start imports, instead -> `from circle import Circle`

natasha-aleksandrova:
don't forget to submit on the course website.

PR #182, opened 2019-05-24
added fizzbuzz exercise for lesson02

esokada:


students/esokada/lesson02/FizzBuzz.py
```diff
@@ -0,0 +1,11 @@
+x = 1
+while x <= 100:
```
natasha-aleksandrova: you can also use `range()` here

esokada: Right! Thanks for the tip!

PR #183, opened 2019-05-25
trigram.py

pkoleyni:


pkoleyni:
Trigram assignment took a lot of time, i feel that explanation for this assignment is not clear for online students like me which we dont have the opportunity to be in class and have discussion about the assignment. 

Regarding except_exercise.py
looking at except_test.py, looks like import antigravity is unused import statement 

def last_fun():
    print(cust, cheese)
    time.sleep(1)
    import antigravity

I think i finished except_exercise.py but i am not getting same result as what is in https://uwpce-pythoncert.github.io/PythonCertDevel/exercises/except_exercise.html

Can you please help me understand where am i doing it wrong ?

natasha-aleksandrova:
`import antigravity` does something magic on import 😉

students/pkoleyni/Lesson05/except_exercise.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+
+# Here is a try/except block. Add an else that prints not_joke
+try:
+    not_joke = fun(first_try[2])
+except SyntaxError:
+    print('Run Away!')
+else:
+    print('not_joke')
```
natasha-aleksandrova: `Add an else that prints not_joke`

you should be printing here `not_joke` variable -> `print(not_joke)`

students/pkoleyni/lesson04/trigram.py
```diff
@@ -0,0 +1,62 @@
...
+            word_pairs[pair].append(follower)
+    return word_pairs
+
+
+def new_text(trigrams_dic):
+    a_list = []
+    l = len(trigrams_dic.keys())
+    for i in range(10):
+        random_number = random.randint(0,l)
+        key = list(trigrams_dic.keys())[random_number]
```
natasha-aleksandrova: this throws an IndexError: if your length of dictionary is 4 and if your random_number is 4 it will throw that IndexError

natasha-aleksandrova: you need to start with a random key first: `key = random.choice(list(trigrams_dic))`
and then in your loop you need to look up the key (please check assignment requirements):
```
new_words.extend(list(key) + [word])

key = tuple(new_words[-2:])
```

students/pkoleyni/lesson04/trigram.py
```diff
@@ -0,0 +1,62 @@
...
+        a_list.append(" ".join(list(key)))
+        a_list.append(" ".join(trigrams_dic[key]))
```
natasha-aleksandrova: you need to use a random here too to choose *ONE* word from your words associated with a key:
`word = random.choice(trigrams_dic[key])`

PR #191, opened 2019-05-28
Session 9 - Final Mailroom submission

RandiPet:


students/randi_peterson/session09/mailroom.py
```diff
@@ -0,0 +1,79 @@
...
+            string_to_print += row_name_format.format(name) + row_data_format.format(*donor_report[name]) + '\n'
+
+        return string_to_print
+
+    def print_donors(self):
+        #Prints donor names
+        donor_names = ''
+        for donors in self.donor_list:
+            donor_names+= donors.name + '\n'
+        return donor_names
```
natasha-aleksandrova: shortcut: `return "\n".join([donor.name for donor in self.donor_list])`

students/randi_peterson/session09/test_mailroom.py
```diff
@@ -0,0 +1,111 @@
+import pytest
+from mailroom import *
+from cli_main import *
```
natasha-aleksandrova: avoid start imports

PR #194, opened 2019-05-29
added updated mailroom 4

Chieu1:


Chieu1:
Natasha,

For lesson 7 on Html Remder, there
should be 8 or 9 html output 
reports ? 
I see on the exercise 
“run.html.render” , you can generate
Up to 8 html output reports.

Thanks,

Chieu

PR #204, opened 2019-06-03
Added Exercise 2.2 - Grid Printer

ajaggarcia10:
Finished all three sections of Exercise 2.2.

Apologies for the extra commits, I was fiddling with getting git to work with uploading files.


students/andrew_garcia/lesson02_Grid/Grid_Printer_1.py
```diff
@@ -0,0 +1,30 @@
...
+        print_vertical()
+
+    def entire_grid():
+        print_horizontal()
+        print_multiple_vertical()
+        print_horizontal()
+        print_multiple_vertical()
+        print_horizontal()
+
+    entire_grid()
```
natasha-aleksandrova: there is no need for these to be nested function; better to define them individually

PR #207, opened 2019-06-05
Lesson 7 Assignment

brgalloway:
Seems to render decently, but have two tests in parts 8 and 9 that are failing.


students/brgalloway/Lesson_7/html_render.py
```diff
@@ -4,48 +4,139 @@
...
+
+    def __init__(self, link, content=None, **kwargs):
+        kwargs['href'] = link
+        super().__init__(content, **kwargs)
+        
+class H(OneLineTag):
+
+    tag = "h"
+    def __init__(self, num, content=None, **kwargs):
+        assert type(num) == int
```
natasha-aleksandrova: you don't typically use `assert` statements in production code. use `raise Exception` instead.

natasha-aleksandrova:
@brgalloway nice work!

PR #209, opened 2019-06-05
Lesson04 and Lesson05 updates

buttsjames:


natasha-aleksandrova:
@buttsjames something is off here with your PR...

were you using branches other than `master` or using different laptops?

there are also tons of lines like this `<<<<<<< HEAD` - those come from merge conflicts which can be caused by above reasons I mentioned.

Also, are `.idea` files yours? Please don't commit those.

Let us know if you need further help with this.

buttsjames:
Yes, I saw that too. You can cancel/delete and I’ll see if I can figure it out and create a new PR.

I got sideways when I tried to delete a very large file from the tree/cache that I checked in by accident. I’ll see if I can figure it out and resubmit.

natasha-aleksandrova:
@buttsjames ok will close for now.

Please reach out if you get stuck and we can help you out.

PR #218, opened 2019-06-08
Lesson 4 homeworks and exercises

KRTstudent:
Mailroom and trigrams exercise

### DIFF
--- a/students/kevin_t/lesson4/mailroom.py
+++ b/students/kevin_t/lesson4/mailroom.py
@@ -1,0 +1,117 @@
+#Opening prompt function
+def opening_screen():
+    #Tell the user what their options are. Ask for an input.
+    print('')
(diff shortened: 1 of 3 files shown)


students/kevin_t/lesson4/mailroom.py
```diff
@@ -0,0 +1,117 @@
...
+
+    #Prompt the user to do something. Stay here unless user selects a quit item.
+    while True:
+        #Call the opening screen function
+        user_selection = opening_screen()
+        #Analyze responses and call related function or quit.
+        if user_selection in switch_dict:
+            switch_dict[user_selection](donor_info)
+        elif user_selection == '4':
+            break
```
natasha-aleksandrova: I would recommend mapping option "4" to a quit function, for example:
```
def quit_program():
    print("Bye!")
    sys.exit()


```

PR #228, opened 2019-06-12
adding lesson01 python pushups

gnar5tar:
Completed break_me.py task and coding bat puzzles

### DIFF
--- a/students/zach_thomson/lesson02/fizzbuzz.py
+++ b/students/zach_thomson/lesson02/fizzbuzz.py
@@ -1,0 +1,14 @@
+def fizzBuzz():
+    i = 1
+    while i <= 100:
+        if i % 5 == 0 and i % 3 == 0:
+            print('FizzBuzz')
+        elif i % 5 == 0:
+            print('Buzz')
+        elif i % 3 == 0:
+            print('Fizz')
+        else:
+            print(i)
+        i = i + 1
+
+fizzBuzz()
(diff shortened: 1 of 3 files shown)


students/zach_thomson/lesson02/fizzbuzz.py
```diff
@@ -0,0 +1,14 @@
+def fizzBuzz():
```
natasha-aleksandrova: pep8 style violation: use snake_case style for var names and func names -> `fizz_buzz`

students/zach_thomson/lesson02/fizzbuzz.py
```diff
@@ -0,0 +1,14 @@
...
+    i = 1
+    while i <= 100:
```
natasha-aleksandrova: note that you could have used `range` here which is more appropriate

natasha-aleksandrova:
don't forget to submit on the course website with a link to this PR

students/zach_thomson/lesson02/fizzbuzz.py
gnar5tar: Ok, good to know. I will use that next time.

students/zach_thomson/lesson02/fizzbuzz.py
gnar5tar: So it'd be more appropriate to have my while statement look like:

while i in range(101)

PR #236, opened 2019-06-17
Lesson 9 Mailroom OOP

brgalloway:


students/brgalloway/Lesson_9/mailroom/donor_models.py
```diff
@@ -0,0 +1,220 @@
...
+      self.times_donated = 0
+      self.average_donation = 0
+      
+   def __repr__(self):
+      # Prints out the object in dictionary format
+      # TODO output Donor object as dictionary to match representation
+      return f"{{\"{self.fullname}\": {{\"donation_total\": {self.donation_total:.2f}, \"times_donated\": {self.times_donated}, \"average_donation\": {self.average_donation:.2f}}}"
+       
+   @property
+   def donate(self):
```
natasha-aleksandrova: properties are usually nouns which means this should be named `donation_total`.. however since you're not "hiding" the attribute or doing anything special you really don't need to use a property

students/brgalloway/Lesson_9/mailroom/donor_models.py
```diff
@@ -0,0 +1,220 @@
...
+      # returns the donation total amount
+      return self.donation_total
+   
+   @donate.setter
+   def donate(self, donation):
+      # calculates all donor attribute actions
+      self.donation_total = donation + self.donation_total
+      self.times_donated+=1
+      self.average_donation = self.donation_total / self.times_donated
+      self.donor_attributes = [self.fullname, self.donation_total, self.times_donated, self.average_donation]
```
natasha-aleksandrova: this should probably be a method

also, `self.donor_attributes` isn't a pythonic pattern - access those attributes directly

students/brgalloway/Lesson_9/mailroom/donor_models.py
```diff
@@ -0,0 +1,220 @@
...
+   @classmethod
+   def from_donor(cls, fullname, donation_total, times_donated=1):
+      # alternative constructor
+      cls.fullname = fullname
+      cls.donation_total = donation_total
+      cls.times_donated = times_donated
+      cls.average_donation = donation_total / cls.times_donated
+      cls.donor_attributes = [fullname, donation_total, cls.times_donated, cls.average_donation]
+   
+      return  cls
```
natasha-aleksandrova: is this function used anywhere? the pattern above isn't very pythonic - I could suggest alternate implementation but I am not clear on the intent

students/brgalloway/Lesson_9/mailroom/donor_models.py
```diff
@@ -0,0 +1,220 @@
...
+   
+   def __lt__(self, other):
+      return self.donation_total < other.donation_total
+
+   def send_thankyou(self, fullname, donation):
```
natasha-aleksandrova: this is an instance method which means full name and donations are attributes available to you via `self`

they should not be passed in as parameters.

students/brgalloway/Lesson_9/mailroom/donor_models.py
```diff
@@ -0,0 +1,220 @@
...
+               email_output.append(filename)
+               file.write(email_template)
+      else:
+         with open(filename, "w") as file:
+               email_output.append(filename)
+               file.write(email_template)
+
+      return email_output
+
+class DonorCollection(Donor):
```
natasha-aleksandrova: why is DonorCollection inherting from Donor? what is the relationship here?

hint: remember that DonorCollection would have reference to many Donor objects

PR #237, opened 2019-06-18
Added List_Lab.py

ajaggarcia10:


students/andrew_garcia/lesson_03/List_Lab.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+new_fruit3 = input("Add a fruit to the list: ")
+fruit_list.insert(0, new_fruit3.title())
+print(fruit_list)
+
+# Print fruits starting with P
+print('\n' + "The fruits in your list that start with P are:")
+for item in fruit_list:
+    if 'P' in item[0]:
```
natasha-aleksandrova: using `==` is more applicable since `in` tests for element inside of a collection and here you're testing character against character.

students/andrew_garcia/lesson_03/List_Lab.py
```diff
@@ -0,0 +1,110 @@
...
+    favorites_list = []
+    for item in list:
+        while True:
+            answer = input("Do you like {} ?: [Yes/No] ".format(item.lower()))
+            if answer.lower() == 'no':
+                break
+            elif answer.lower() == 'yes':
+                favorites_list += [item]
+                break
+            elif answer.lower() != 'yes' or 'no':
```
natasha-aleksandrova: this isn't going to do what you expect because of the order of operation - it will evaluate both sides of `or` which means `or 'no`' will always be true.

you will want either of the two below:
`elif answer.lower() in ('yes', 'no')`

`elif answer.lower() != 'yes' or answer.lower() != 'no'`

the first being the better option.

students/andrew_garcia/lesson_03/List_Lab.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+def reverse_fruits(list):
+    reversed_list = []
+    for item in list:
+        reversing = []
+        reversing += item
+        reversing.reverse()
+        backwards = ''.join(reversing)
+        reversed_list += [backwards]
```
natasha-aleksandrova: you can just reverse the string using slicing
```
r_fruits = []
for fruit in fruits:
    r_fruits.append(fruit[::-1])
```

PR #239, opened 2019-06-19
Updating Lesson 06 and adding Lesson 07

cholso:
(1) Updating mailroom4.py and test_mailroom4.py to run more successful tests 
(2) Adding Lesson 07 work


natasha-aleksandrova:
please disregard my previous comment (which I edited) about submissions - there was an error during your registration time which enrolled you in a different version of the site which means we have to go a different place to grade your work and sometimes creates confusion for us 🙂 - you're all good!

PR #241, opened 2019-06-19
Finished object oriented refactor of mailroom, included testing

gregdevore:
Is there a better way to create the donor collection objects in cli_main.py when running it from the command line? Currently, in the __name__ == "__main__" section, I create some sample donor data and then just call main(). All of the functions that use donor objects work, but it feels weird to be referring to variables that were created outside of the function (it's almost like they're global variables, but that's not quite right). Just wondering if there's a better/preferred way to do that!

Thanks,
Greg


students/gregdevore/lesson09/donor_models.py
```diff
@@ -0,0 +1,203 @@
...
+            name (str):
+                Donor's name
+            amount (float):
+                Donation amount
+        """
+        # Create donor object if it doesn't exist, otherwise retrieve from collection
+        d = self.getDonor(name)
+        # Update donor object and replace in collection
+        d.addDonation(amount)
+        self.donors[name] = d
```
natasha-aleksandrova: remember that you can always shortcut since you're using temp var `d`:
 `self.donors[name] = self.get_donor(name).add_donation(amount)`

PR #243, opened 2019-06-20
Added string formatting and mailroom

esokada:
Thanks for looking at this. I would especially appreciate any feedback on the mailroom exercise, as I wasn't sure if I organized it well, and I think I ended up with more code than I needed.  

I also have a question: occasionally when committing, I get a commit like the following below:
Merge branch 'master' of https://github.com/UWPCE-PythonCert-ClassRep… 

I am not sure what to do with this, as I didn't do anything intentionally to cause it.  Is it okay to ignore this?


natasha-aleksandrova:
that git message is usually due to running `merge upstream` commands - which is totally expected and ok!

students/esokada/lesson03/mailroom.py
```diff
@@ -0,0 +1,75 @@
...
+        dtotal = sum(d[1])
+        dnum = len(d[1])
+        davggift = mean(d[1])
+        reportlines.append([dname, dtotal, dnum, davggift])
+    for r in sorted(reportlines, key = itemgetter(1), reverse = True):
+        print(f"{r[0]:15} {r[1]:>20.2f} {r[2]:>10} {r[3]:>20.2f}")
+
+def quit_program():
+    print("Goodbye!")
+    sys.exit()
```
natasha-aleksandrova: 👍

students/esokada/lesson03/mailroom.py
```diff
@@ -0,0 +1,75 @@
...
+
+def write_report():
+    titles = ["Donor Name", "Total Given", "Num Gifts", "Average Gift"]
+    print(f"{titles[0]:15} {titles[1]:>20} {titles[2]:>10} {titles[3]:>20}")
+    reportlines = []
+    for d in donors:
+        dname = d[0]
+        dtotal = sum(d[1])
+        dnum = len(d[1])
+        davggift = mean(d[1])
```
natasha-aleksandrova: should use average here not mean

students/esokada/lesson03/mailroom.py
```diff
@@ -0,0 +1,75 @@
...
+        reportlines.append([dname, dtotal, dnum, davggift])
+    for r in sorted(reportlines, key = itemgetter(1), reverse = True):
```
natasha-aleksandrova: no spaces around keyword items -> `key=itemgetter(1)`

and nice use of `itemgetter` !

students/esokada/lesson03/mailroom.py
```diff
@@ -0,0 +1,75 @@
...
+def send_thankyou():
+    donorname = input("Please input a name:\n")
+    if donorname == "cancel":
+        return
+    if donorname == "list":
+        for i in donors:
+            print(i[0])
+        donorname = input("Please input a name:\n")
+        if donorname == "cancel":
+            return
```
natasha-aleksandrova: lines 24 to 26 should be de-dented - no need for these to only present on "list" selection as user may not start with that.

also mahy want to use while True loop at the beginning and loop until user selects "list" or enters a name and exit when user types cancel or "C"

students/esokada/lesson03/mailroom.py
```diff
@@ -0,0 +1,75 @@
...
+    donationamount = input("Please enter a donation amount:\n")
+    if donationamount == "cancel":
+        return
+    existingdonor = False
+    for d in donors:
+        if d[0] == donorname:
+            existingdonor = True
```
natasha-aleksandrova: use `is` with True/False

`==` is used for primitive objects like ints/strings

`is` is used for object comparison (like class objects, True/False, None)

students/esokada/lesson03/mailroom.py
```diff
@@ -0,0 +1,75 @@
...
+            break
+    if existingdonor == False:
+        donors.append([donorname, []])
```
natasha-aleksandrova: you can use `for / else` structure instead of using flag variable `existingdonor`

students/esokada/lesson03/mailroom.py
esokada: Can you give me any information about why? I tried to search online for things like "average vs mean python" and "average function python", but I couldn't find much.

students/esokada/lesson03/mailroom.py
esokada: Whoa, that's handy! Didn't know I could do that, thanks!

esokada:
Thanks so much for your feedback! It's the first time I've ever gotten such detailed feedback in my (very short) programming experience, and it's a huge help.

students/esokada/lesson03/mailroom.py
natasha-aleksandrova: Simply because assignment asks for an average 🙂 

```
Include Donor Name, total donated, number of donations, and average donation amount as values in each row.
```

natasha-aleksandrova: average and mean are different things - not a big deal but assignment asks for average

natasha-aleksandrova:
@esokada you're very welcome!

students/esokada/lesson03/mailroom.py
esokada: Right, thanks!

PR #247, opened 2019-06-21
Refactored Lesson 9 OOP Mailroom

brgalloway:
Took the feedback and recommendations that Natasha provided, and cleaned up the classes in donors_models.py


brgalloway:
Need to fix cli_main.py. Codes not ready.

PR #248, opened 2019-06-22
updating mailroom per Natasha's comments

esokada:
Thanks for your help! I believe I have addressed all your comments.

### DIFF
--- a/students/esokada/lesson03/mailroom.py
+++ b/students/esokada/lesson03/mailroom.py
@@ -15,24 +15,22 @@ donors = [["Usagi", [12.5, 11.17, 130.1]], ["Ami", [17, 230, 2115]],
 #functions
 
 def send_thankyou():
-    donorname = input("Please input a name:\n")
-    if donorname == "cancel":
-        return
-    if donorname == "list":
-        for i in donors:
-            print(i[0])
+    while True:
         donorname = input("Please input a name:\n")
-        if donorname == "cancel":
+        if donorname == "C":
             return
+        if donorname == "list":
+            for i in donors:
+                print(i[0])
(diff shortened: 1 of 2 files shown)


students/esokada/lesson03/mailroom.py
```diff
@@ -15,24 +15,22 @@
...
-    if donationamount == "cancel":
+    if donationamount == "C":
         return
-    existingdonor = False
     for d in donors:
         if d[0] == donorname:
-            existingdonor = True
             break
-    if existingdonor == False:
+    else:
```
natasha-aleksandrova: this is close.. but here is a more condensed version:
```
    while True:
        donorname = input("Please input a name, "list" command or "C" to cancel:\n")  # clarify what a user can input
        if donorname == "C":
            return
        elif donorname == "list":
            for donor in donors:
                print(donor[0])
        else:
            # here we will be adding donors
            donation_amount = input("Please enter a donation amount:\n")
            for donor in donors:
                if donor[0] == donorname:
                    donor[1].append(donation_amount)
            else:
                donors.append([donorname, [donation_amount]])

            print(f"Dear {donorname}, thank you for your generous donation of ${float(donation_amount):.2f}.\n")
```

natasha-aleksandrova: please review above ^^ let me know if you have questions.

esokada: Thanks, that makes sense!  I will incorporate these changes into next mailroom assignment.

natasha-aleksandrova: sounds good!

PR #250, opened 2019-06-22
Added unit tests for mailroom

bryancoates:


students/bcoates/lesson06/mailroom4.py
```diff
@@ -0,0 +1,182 @@
+#!/usr/bin/env python3
+
+# Starting list of donors and their donation amounts
+donor_history = {
+    "donor_1": {"name": "Lana Kane", "donations": [2999.99]},
+    "donor_2": {"name": "Cheryl Tunt", "donations": [150.20, 98192.10]},
+    "donor_3": {"name": "Cyril Figgis", "donations": [819.25, 998.62, 10.50]},
+    "donor_4": {"name": "Pam Poovey", "donations": [74926.10, 2675.87, 3289.33]},
+    "donor_5": {"name": "Ray Gillette", "donations": [3820.90]}
```
natasha-aleksandrova: this isn't ideal structure for this program - your key should benefit you but it does not in this case.

If you needed to find a value in your dictionary, how would you know what key to use? You program uses donor names to update your data in which case you should use donor name as a key.

students/bcoates/lesson06/mailroom4.py
```diff
@@ -0,0 +1,182 @@
...
+    # Build a prompt message based on prompt and actions
+    prompt_message = prompt
+    for item in prompt_actions:
+        prompt_message += "\n" + item[0] + " - " + prompt_actions[item]['message']
+    prompt_message += "\n> "
+    
+    # Prompt the user and set the selection
+    while True:
+        selection = input(prompt_message)
+        if selection not in prompt_actions:
```
natasha-aleksandrova: this should use a `KeyError` check

students/bcoates/lesson06/mailroom4.py
```diff
@@ -0,0 +1,182 @@
...
+            print("Please select a valid number...")
+        elif prompt_actions[selection]['action']() == 'quit':
+            print("Quitting...")
+            break
```
natasha-aleksandrova: you can map quit_program function to `sys.exit()`

students/bcoates/lesson06/mailroom4.py
```diff
@@ -0,0 +1,182 @@
...
+            zero_check = 1 / float_amount
+            break
+        except ValueError:
+            print("Invalid donation amount, please try again...")
+        except ZeroDivisionError:
+            print("Donation amount can't be zero, please try again...")
+
+    # Add the donation to the donor's list of donations
+    add_donation_amount(donor, float_amount)
+    return float_amount
```
natasha-aleksandrova: functions like this don't typically return values ("void" functions)

students/bcoates/lesson06/mailroom4.py
```diff
@@ -0,0 +1,182 @@
...
+
+def add_donation_amount(donor, float_amount):
+    """ Add donation for donor """
+
+    # Add the donation to the donor's list of donations
+    donor_history[donor]['donations'].append(float_amount)
+            
+    # Return the donation amount
+    return float_amount
```
natasha-aleksandrova: functions like this don't typically return values ("void" functions)

PR #251, opened 2019-06-23
Submitting Slicing.py

nlenssen2013:
I had issues with the lesson 2 exercises. I think I did this right.

### DIFF
--- a/students/Nick_Lenssen/lesson03/slicing.py
+++ b/students/Nick_Lenssen/lesson03/slicing.py
@@ -12,6 +12,7 @@ def change_order(seq):
     split = int(len(seq)/3)
     return (seq[2*split:]+seq[:split]+seq[split:2*split])
 
+
 a_string = "this is a string"
 a_tuple = (2, 54, 13, 12, 5, 32,2,6,1,0,'my',6)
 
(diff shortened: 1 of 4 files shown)


students/Nick_Lenssen/lesson03/slicing.py
```diff
@@ -0,0 +1,32 @@
+def first_last(seq):
+    return (seq[-1:]+seq[1:-1]+seq[:1])
+
+def every_other(seq):
+    return (seq[0::2])
+
+def remove_high_low(seq):
+    return (seq[4:-4])
+def reverse(seq):
+    return (seq[-1::-1])
```
natasha-aleksandrova: can be just `seq[::-1]`

students/Nick_Lenssen/lesson03/slicing.py
```diff
@@ -0,0 +1,32 @@
+def first_last(seq):
+    return (seq[-1:]+seq[1:-1]+seq[:1])
+
+def every_other(seq):
+    return (seq[0::2])
+
+def remove_high_low(seq):
+    return (seq[4:-4])
```
natasha-aleksandrova: "with the first 4 and the last 4 items removed, and then every other item in the remaining sequence."

should be `seq[4:-4:2]`

students/Nick_Lenssen/lesson03/slicing.py
natasha-aleksandrova: need to have two lines spaces between functions

PR #256, opened 2019-06-26
Adding lesson 2 files.

jineehan:
Adding the three practice files for lesson 2.

### DIFF
--- a/students/jinee_han/lesson02/FizzBuzz.py
+++ b/students/jinee_han/lesson02/FizzBuzz.py
@@ -1,0 +1,45 @@
+###############################
+# Fizz Buzz
+# 06/24/2019 Jinee Han
+# Python Programming Lesson 2
+#####################################
+
+for i in range(101):
+    if i == 0:
+        pass
+    elif (i%3) == 0 and (i%15) != 0:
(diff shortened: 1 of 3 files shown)


students/jinee_han/lesson02/FizzBuzz.py
```diff
@@ -0,0 +1,45 @@
+###############################
+# Fizz Buzz
+# 06/24/2019 Jinee Han
+# Python Programming Lesson 2
+#####################################
+
+for i in range(101):
+    if i == 0:
+        pass
```
natasha-aleksandrova: instead of special case for `0` - use `range(1, 101)`

students/jinee_han/lesson02/FizzBuzz.py
```diff
@@ -0,0 +1,45 @@
...
+    elif (i%3) == 0 and (i%15) != 0:
+        print ("Fizz")
+    elif (i%5) == 0 and (i%15) != 0:
+        print ("Buzz")
+    elif (i%15) == 0:
```
natasha-aleksandrova: you could have check against 15 first to avoid having to check 15 every time.

PR #257, opened 2019-06-27
Mailroom, part 1 submission

gnar5tar:
Pull request for mailroom part 1, ready to be graded.

### DIFF
--- a/students/zach_thomson/lesson03/mailroom.py
+++ b/students/zach_thomson/lesson03/mailroom.py
@@ -32,12 +32,7 @@ def number_donations(x):
 def average_gift(x):
     return sum_function(x)/len(donor_db[x][1])
 
-header = ('Donor Name', 'Total Given', 'Num Gifts', 'Average Gift')
 
-table_header = print("{:<20}| {} | {} | {}".format(*header) + '\n' + "-" * 60)
-
-print(table_header)
-print(donor_db[1])
 
 def create_table():
     report_table = []
@@ -52,9 +47,17 @@ def create_table():
 
 print(create_table())
 
+header = ('Donor Name', 'Total Given', 'Num Gifts', 'Average Gift')
+table_header = "{:<20}| {} | {} | {}".format(*header) + '\n' + "-" * 60
+
 def create_report():
     '''formats create_table into the create_report format'''
-    pass
+    table = create_table()
+    print(table_header)
+    line_format = ("{:<20}" + " $" + "{:>12.2f}" + "{:>11}" + "  $" + "{:>12.2f}" + '\n') * len(donor_db)
+    print(line_format.format(*table))
+
+#create_report()
 
 def exit_program():
     print('Bye')
--- a/students/zach_thomson/lesson03/mailroom.py
+++ b/students/zach_thomson/lesson03/mailroom.py
@@ -1,10 +1,10 @@
 import sys
 
-donor_db = [('donor 1', [10000.00, 20000.00, 4500.00]),
-            ('donor 2', [100.00, 500.00]),
-            ('donor 3', [25.00]),
-            ('donor 4', [100000.00, 50000.00, 125000.00]),
-            ('donor 5', [50.00])]
+donor_db = [('Eddie Vedder', [10000.00, 20000.00, 4500.00]),
+            ('Chris Cornell', [100.00, 500.00]),
+            ('Kurt Cobain', [25.00]),
+            ('Dave Matthews', [100000.00, 50000.00, 125000.00]),
+            ('Dave Grohl', [50.00])]
 
 
 prompt = '\n'.join(('Welcome to the mailroom',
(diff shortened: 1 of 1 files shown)


students/zach_thomson/lesson03/mailroom.py
```diff
@@ -0,0 +1,129 @@
...
+    """takes in a name and determines if it is in the donor database"""
+    i = 0
+    result = False
+    while i < len(donor_db):
+        if x == donor_db[i][0]:
+            result = True
+            break
+        else:
+            i = i + 1
+    return result
```
natasha-aleksandrova: it is simpler and more pythonic to loop over list directly:
```
for donor in donor_db:
    if donor[0] == name:
        return True
return False
```

students/zach_thomson/lesson03/mailroom.py
```diff
@@ -0,0 +1,129 @@
...
+
+prompt = '\n'.join(('Welcome to the mailroom',
+         'Please choose from the following options:',
+         '1 - Send a Thank You',
+         '2 - Create a Report',
+         '3 - Quit',
+         '> '))
+
+#send a thank you tasks
+def donor_in_list(x):
```
natasha-aleksandrova: don't use `x` (anonymous vars) as function input params

students/zach_thomson/lesson03/mailroom.py
```diff
@@ -0,0 +1,129 @@
...
+    return result
+
+def donor_idx(x):
+    """takes in a name and finds the index number in donor database"""
+    i = 0
+    while i < len(donor_db):
+        if x == donor_db[i][0]:
+            return i
+        else:
+            i += 1
```
natasha-aleksandrova: `enumerate` is more appropriate here

students/zach_thomson/lesson03/mailroom.py
```diff
@@ -0,0 +1,129 @@
...
+
+def donor_names():
+    """return a list of names in the donor database"""
+    i = 0
+    donors = []
+    while i < len(donor_db):
+        donors.append(donor_db[i][0])
+        i += 1
+    return donors
```
natasha-aleksandrova: this should also be a simpler for loop over list elements.

students/zach_thomson/lesson03/mailroom.py
```diff
@@ -0,0 +1,129 @@
...
+
+donation_email = "\nDear {},\nThank you for your generous donation of ${:.2f}!\n"
+
+def thank_you():
+    ty_prompt = input('Please enter a full name or type list to see all the current donors: ')
+    while ty_prompt.lower() == 'list':
+        print(donor_names())
+        ty_prompt = input('Please enter a full name or type list to see all the current donors: ')
+    else:
+        if donor_in_list(ty_prompt) == True:
```
natasha-aleksandrova: use `is` with objects instead of `==`

also, you can just have this as `if donor_in_list(ty_prompt):`

students/zach_thomson/lesson03/mailroom.py
```diff
@@ -0,0 +1,129 @@
...
+            donation_amount = float(input('Please enter a donation amount: '))
+            donor_db[donor_idx(ty_prompt)][1].append(float(donation_amount))
+        else:
+            donation_amount = float(input('Please enter a donation amount: '))
```
natasha-aleksandrova: the donation_amount line can be above this block (line 57) - since its duplicate on lines 59, 62

students/zach_thomson/lesson03/mailroom.py
```diff
@@ -0,0 +1,129 @@
...
+            donor_db.append(new_entry)
+    print(donation_email.format(ty_prompt, donation_amount))
+
+
+#create a report functions
+def sum_function(x):
+    total = 0
+    for i in donor_db[x][1]:
+        total = total + i
+    return total
```
natasha-aleksandrova: use `sum()` function directly

students/zach_thomson/lesson03/mailroom.py
```diff
@@ -0,0 +1,129 @@
...
+        total = i + total
+    return total
+
+def create_table():
+    """takes the donor database and sorts on total given. also makes a summary
+    table with total given, number of gifts and average amount of gift"""
+    donor_db.sort(key = sum_second, reverse = True)
+    report_table = []
+    i = 0
+    while i < len(donor_db):
```
natasha-aleksandrova: should also be a simple for loop over list elements

students/zach_thomson/lesson03/mailroom.py
```diff
@@ -0,0 +1,129 @@
...
+        report_table.append(donor_db[i][0])
+        report_table.append(sum_function(i))
+        report_table.append(number_donations(i))
+        report_table.append(average_gift(i))
+        i = i + 1
```
natasha-aleksandrova: typically, this report is created first and then you sort this summarized version.

right now your sort key is a bit complex for this reason.

students/zach_thomson/lesson03/mailroom.py
```diff
@@ -0,0 +1,129 @@
...
+    return report_table
+
+header = ('Donor Name', 'Total Given', 'Num Gifts', 'Average Gift')
+table_header = "{:<20}| {} | {} | {}".format(*header) + '\n' + "-" * 60
```
natasha-aleksandrova: these should go either into the appropriate function or at the top of the file as constants (all caps)

PR #261, opened 2019-06-29
Updated mailroom and unit tests per recommendations

bryancoates:


students/bcoates/lesson06/mailroom4.py
```diff
@@ -21,13 +23,14 @@ def main_menu(prompt, prompt_actions):
     
     # Prompt the user and set the selection
     while True:
-        selection = input(prompt_message)
-        if selection not in prompt_actions:
+        try:
+            selection = input(prompt_message)
+            if prompt_actions[selection]['action']() == 'quit':
+                print("Quitting...")
+                quit_program()
```
natasha-aleksandrova: in your switch dict you should map "4" to this function.

PR #262, opened 2019-06-29
made changes and added test_mailroom file to work with pytest

humbertojrg:


students/humberto_gonzalez/session06/test_mailroom.py
```diff
@@ -0,0 +1,66 @@
...
+                ['Jarvis Landry', 150.25, 1, 150.25],
+                ['Tom Brady', 50.0, 1, 50.0]]
+    assert create_report(donor_db) == expected
+    print('Test Passed')
+
+
+def test_create_letter_txt():
+    expected = '''Dear Tom Brady,\n    Thank you for your generous donation of $50.0. \n    Your donation will be put to great use. \n         Sincerely, \n          -The Organization'''
+    assert create_letter_txt('Tom Brady',50.0) == expected
+    print('Test Passed.')
```
natasha-aleksandrova: note that pytest should output the "test passed" information for you

PR #263, opened 2019-06-29
Added Mailroom Part 1

ajaggarcia10:


students/andrew_garcia/lesson_03/Mailroom.py
```diff
@@ -0,0 +1,112 @@
...
+The kind generosity that you and others give us helps us complete the great work we are doing for others.
+
+From,
+Your Local Charity
+\n''')
+    options_menu()
+
+def add_donation(name, all_donations = all_donations):
+    """Takes a new or existing donor and adds their donation to the list"""
+    amount = input(f'How much did {name} donate?: ')
```
charlesrobison: Nice use of f strings!

students/andrew_garcia/lesson_03/Mailroom.py
```diff
@@ -0,0 +1,112 @@
...
+    print(f'''\nDear {name.title()},
+
+Thank you so much for your generous donation of ${amount:.2f}!
+The kind generosity that you and others give us helps us complete the great work we are doing for others.
+
+From,
+Your Local Charity
+\n''')
+    options_menu()
+
```
charlesrobison: Add another blank row between functions - looks like you're doing that for most of these, but this one got missed?

PR #273, opened 2019-07-02
re-adding lessons 2-5 to git for submission

buttsjames:
Fixed Git sync errors, with the following commands:

git fetch upstream
git checkout master
git reset --hard upstream/master  
git push origin master --force

Reimported my changes since my last pull request. Will submit these assignments via the portal.


students/jbutts/Lesson02/series.py
```diff
@@ -7,12 +7,57 @@
 '''
 
 
+<<<<<<< HEAD
+def fibonacci(number):
+=======
 def fibonacci(n):
+>>>>>>> fedf8c86a2a2636875db022c6fb2dee41583f9a7
```
natasha-aleksandrova: looks like some git conflicts in here.

natasha-aleksandrova:
@buttsjames code is full of conflicts (lines with '<<') I would not recommend doing 'reset --hard'  - there is usually no need for that, not sure what issue you were trying to solve.

Have these assignment been reviewed previously? Which ones are new? Please let us know so that we know what we need to review.

PR #274, opened 2019-07-03
Assignment 1

nlenssen2013:


students/Nick_Lenssen/lesson03/mailroom.py
```diff
@@ -0,0 +1,69 @@
...
+            main_list[main_list[0].index(full_name)+1].append(amount)
+            print ('Thank you {} for your generous donation of {}'.format(full_name, amount))
+            main()
+def send_thanks():
+    full_name = input("\n\nPlease enter a name for a Thank You to go out: ")
+    while True:
+        while full_name == 'list':
+            print (main_list[0])
+            full_name = input("\n\nPlease enter a name for a Thank You to go out: ")
+        if full_name == 'c':
```
natasha-aleksandrova: is 'c' a quit option? should the user know about this option?

add `return` statement don't call main() directly

students/Nick_Lenssen/lesson03/mailroom.py
```diff
@@ -0,0 +1,69 @@
...
+            main()
+        if full_name in main_list[0]:
+            print ('in')
+            use_it(full_name)
+        elif full_name not in main_list[0]:
```
natasha-aleksandrova: your if test has only two options: donor exists or not in which case you should use `if/else` logic

students/Nick_Lenssen/lesson03/mailroom.py
```diff
@@ -0,0 +1,69 @@
...
+        print ("{:<20} {:>9} {:>9} {:>26}{:>11}{:>12}".format(*(i, '$', sum(main_list[val+1]), len(main_list[val+1]), '$',round(sum(main_list[val+1])/len(main_list[val+1]),1))))
+        
+def use_it(full_name):
+    while True:
+        amount = input("\nPlease enter in the amount you want to donate: ")
+        if amount == 'c':
+            main_list[0].pop()
+            main_list.pop()
+            print (main_list)
+            main()
```
natasha-aleksandrova: instead of adding donor and then asking amount and then potentially removing donor.. you can just ask amount first, ensure it's valid and then proceed with adding/updating donor.

students/Nick_Lenssen/lesson03/mailroom.py
```diff
@@ -0,0 +1,69 @@
...
+        num = get_option()
+        if num =='a':
+            print ('send')
+            send_thanks()
+        if num == 'b':
+            print ('create')
+            create_report()
+
+main_list = [['Dan C','Erika K','Mindy P','Mike H','Lovey G'], [1,200,20],[34000],[40,50,60],
+                        [13,60,42],[22]]
```
natasha-aleksandrova: per tutorial, recommended structure is:
```
donor_db = [("William Gates, III", [653772.32, 12.17]),
            ("Jeff Bezos", [877.33]),
            ("Paul Allen", [663.23, 43.87, 1.32]),
            ("Mark Zuckerberg", [1663.23, 4300.87, 10432.0]),
            ]
```

PR #275, opened 2019-07-03
added exception exercise and mailroom 3

esokada:
There didn't actually seem to be too many places to add exception handling and list comprehensions, but I hope this is okay.

### DIFF
--- a/students/esokada/lesson05/mailroom3.py
+++ b/students/esokada/lesson05/mailroom3.py
@@ -1,0 +1,84 @@
+#imports
+
+import sys
+
+from statistics import mean
+
+from operator import itemgetter
+
+
+# data structure
+
+donors = {"Usagi":[12.5, 11.17, 130.1], "Ami":[17, 230, 2115], "Rei":[510.1, 50], "Makoto": [324, 22.7], "Minako": [310] }
+
+#functions
+
+def send_thankyou():
+    while True:
+        donorname = input("Please input a name, 'list' command, or 'c' to cancel:\n")
+        if donorname == "c":
+            return
+        elif donorname == "list":
+            print(donors.keys())
+        else:
+            while True:
+                donationamount = input("Please enter a donation amount:\n")
+                try:
+                    if donationamount == "c":
+                        return
+                    if donorname in donors:
+                        donors[donorname].append(float(donationamount))
+                        break
+                    else:
+                        donors.update({donorname: [float(donationamount)]})
+                        break
+                except ValueError:
+                    print("Please enter the donation amount as a number.")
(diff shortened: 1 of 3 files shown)


students/esokada/lesson05/mailroom3.py
```diff
@@ -0,0 +1,84 @@
...
+                try:
+                    if donationamount == "c":
+                        return
+                    if donorname in donors:
+                        donors[donorname].append(float(donationamount))
+                        break
+                    else:
+                        donors.update({donorname: [float(donationamount)]})
+                        break
+                except ValueError:
```
natasha-aleksandrova: ideally exceptions should be around the code that throws it - with this being said I would move float() call to line 25 and just catch exception there.

esokada: Thanks!
I don't want to use try-except with the donationamount = input() line because I want to keep the option of returning out of the function if donationamount == "c".
So instead, what if I use a try-except-else as below? Do you think this is better? 
```
def send_thankyou():
    while True:
        donorname = input("Please input a name, 'list' command, or 'c' to cancel:\n")
        if donorname == "c":
            return
        elif donorname == "list":
            print(donors.keys())
        else:
            while True:
                donationamount = input("Please enter a donation amount:\n")
                if donationamount == "c":
                    return
                try:
                    amount = float(donationamount)
                except ValueError:
                    print("Please enter the donation amount as a number.")
                else:
                    if donorname in donors:
                        donors[donorname].append(amount)
                    else:
                        donors.update({donorname: [amount]})
                    letterdict = {"donor_name": donorname, "donation_amount": amount}
                    print("Dear {donor_name}, thank you for your generous donation of ${donation_amount:.2f}.\n".format(**letterdict))
                    return
```

natasha-aleksandrova: what you have suggested above looks great!

 my personal preference:
```
try:
    amount = float(donation_amount)
except ValueError:
    print("invalid")
    continue

if donorname in donors:
    donors[donorname].append(amount)
else:
    donors.update({donorname: [amount]})
letterdict = {"donor_name": donorname, "donation_amount": amount}
print("Dear {donor_name}, thank you for your generous donation of ${donation_amount:.2f}.\n".format(**letterdict))
return
```
two reasons:
* because of being inside of a while True loop I like to use `continue` to indicate that we should start the loop over form this point
* I like to keep `else` code portion concise 

however, the code you suggested works as well!

esokada: Thanks, that makes so much sense!  I didn't think of `continue`.

PR #279, opened 2019-07-06
Circle Class Exercise

franjaku:
may add step 8 optional code later...


students/franjaku/Lesson_8/circle.py
```diff
@@ -0,0 +1,84 @@
...
+    def diameter(self, diameter):
+        self._radius = diameter / 2
+
+    @property
+    def area(self):
+        return math.pow(self.radius, 2)*math.pi
+
+    @classmethod
+    def from_diameter(self, diameter):
+        return Circle(diameter/2)
```
natasha-aleksandrova: in classmethods, the first param is `cls` and instance is created via `return cls(diameter/2)`

PR #281, opened 2019-07-10
adding circle assignment

humbertojrg:


students/humberto_gonzalez/session08/circle.py
```diff
@@ -0,0 +1,80 @@
...
+        self._area = 2*pi*self._radius
+    
+    @property
+    def area(self):
+        return self._area
+    
+    @classmethod
+    def from_diameter(cls,diameter):
+        self = cls(diameter/2)
+        return self
```
natasha-aleksandrova: should just be `return cls(diameter/2)` you're not gaining anything by assigning it to `self` variable

PR #288, opened 2019-07-17
Lesson 09 Assignment Submission

mitch334:


students/mitch334/lesson09/mailroom.py
```diff
@@ -0,0 +1,244 @@
...
+
+        return (f'Dear {self.name},\n\nThank you for the generous donation of ${amount:.2f}.\n\n'
+            'Sincerely,\nMatthew Mitchell')
+
+
+class DonorCollection:
+    """Class responsible for donor collection data encapsulation"""
+
+    def __init__(self, donors=None):
+        if donors == None:
```
natasha-aleksandrova: use `is` with objects like `is None` `is True`

students/mitch334/lesson09/mailroom.py
```diff
@@ -0,0 +1,244 @@
...
+            self._donors = donors
+
+    @property
+    def donors(self):
+        return self._donors
+
+    def donors_list(self, list_output):
+        """Create donor list or database details"""
+
+        list = '\n'
```
natasha-aleksandrova: you can't use `list` as a variable name - you are breaking built-in `list()` function

students/mitch334/lesson09/mailroom.py
```diff
@@ -0,0 +1,244 @@
...
+# send_thankyou()
+# donors.create_report()
+# donors.send_letters()
+
+if __name__ == '__main__':
+
+    option = None
+    main_menu = {'1': send_thankyou, '2': donors.create_report, '3': donors.send_letters, '4': quit }
+
+    while option != '4':
```
natasha-aleksandrova: this is better in a `while True` loop that exists via option 4 on its own since you have the function defined.

PR #292, opened 2019-07-19
added mailroom 4

esokada:


students/esokada/lesson06/test_mailroom4.py
```diff
@@ -0,0 +1,69 @@
...
+    update_donor(donorname, donors, amount)
+    assert 1234 in donors["Poirot"]
+
+def test_write_oneletter():
+    donorname = "Roderick Allyn"
+    amount = 345
+    letter = write_oneletter(donorname, amount)
+    assert "Roderick Allyn" in letter
+    assert "generous donation" in letter
+    assert "345" in letter
```
natasha-aleksandrova: why not assert the full text?

students/esokada/lesson06/test_mailroom4.py
```diff
@@ -0,0 +1,69 @@
...
+
+#test for functions used in write_report()
+
+def test_create_rows():
+    testreport = create_rows(donors)
+    assert testreport.startswith("Donor")
+    assert "Poirot" in testreport
+    assert "1666" in testreport
```
natasha-aleksandrova: why not assert the full report?

students/esokada/lesson06/test_mailroom4.py
```diff
@@ -0,0 +1,69 @@
...
+    sample_letter = batch_letters_text(k, v)
+    assert "Usagi" in sample_letter
+    assert "3.00" in sample_letter
+    assert "6.00" in sample_letter
+
+def test_batch_letters_files():
+    for k, v in donors.items():
+        filetext = batch_letters_text(k, v)
+        batch_letters_files(k, filetext)
+        assert os.path.isfile("Poirot.txt") is True
```
natasha-aleksandrova: you can omit `is True` part

esokada: Right, thanks!

esokada:
Thanks for pointing this out - this was entirely due to my laziness, and being unsure how thorough to be with these tests.  (I guess you can never be too thorough!)  I will make sure to test for the full text in the future.

PR #295, opened 2019-07-21
adding mailroom assignment

humbertojrg:


students/humberto_gonzalez/session09/donor_models.py
```diff
@@ -0,0 +1,146 @@
...
+        self.donations = []
+    
+    def add_donation(self,donation):
+        """
+        adds donation to the donations list
+        """
+        try:
+            donation = float(donation)
+            self.donations.append(donation)
+        except:
```
natasha-aleksandrova: make sure to catch specific exception - I'm assuming should be around float() call?

students/humberto_gonzalez/session09/donor_models.py
```diff
@@ -0,0 +1,146 @@
...
+        """
+        return round(sum(self.donations)/len(self.donations),2)
+    
+    def generate_thank_you_note(self):
+        """
+        created the thank you note text for the donor
+        """
+        donor_name = self.name
+        donation = self.donations[len(self.donations)-1]
+        txt = f'Dear {donor_name},\n    Thank you for your generous donation of ${donation}'
```
natasha-aleksandrova: you can use `self.donor_name` directly in the f-string

students/humberto_gonzalez/session09/donor_models.py
```diff
@@ -0,0 +1,146 @@
...
+        return txt
+    
+    def __lt__(self,other):
+        return self.total_donations() > other.total_donations()
+    
+    def __eq__(self,other):
+        return self.total_donations() == other.total_donations()
+    
+
+class donorCollection:
```
natasha-aleksandrova: camel case for class names -> `DonorCollection`

PR #300, opened 2019-07-23
Lesson 8 hw

KRTstudent:
Had some trouble with the 'repr' and sort methods. Some help would be appreciated.

### DIFF
--- a/students/kevin_t/lesson8/test_circle.py
+++ b/students/kevin_t/lesson8/test_circle.py
@@ -1,0 +1,60 @@
+#!/usr/bin/env python
+import pytest
+from circle import *
+
+def test_step1_radius():
+    c = Circle(5)
+    assert c.radius == 5
+
+def test_step2_diameter():
+    c = Circle(5)
+    assert c.diameter == 10
+
+def test_step3_assign_diameter():
+    c = Circle(5)
+    assert c.diameter == 10
+    c.diameter = 2
+    assert c.diameter == 2
+    assert c.radius == 1
+
+def test_step4_area():
+    c = Circle(2)
+    assert c.area == 12.566370
+
+def test_step5_create_from_diameter():
+    c = Circle.from_diameter(8)
+    assert c.diameter == 8
+    assert c.radius == 4
+
+def test_step6_print_circle():
+    c = Circle(4)
+    assert str(c) == 'Circle with radius: 4.00000'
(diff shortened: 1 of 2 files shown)


natasha-aleksandrova:
@KRTstudent can you tell us a little more about your issues?

KRTstudent:
Natasha,

in the lines that i commented out, one of the assertions results in a
string that says 'Circle(5)' = 'Circle(5)' and results in a failure. I
checked into the failure and it's comparing strings (not memory location),
so I don't know why it's failing. Spelling is right, no hidden " " spaces.
Everything that was commented out either wasn't passing when it looked to
me that it was.

students/kevin_t/lesson8/test_circle.py
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/env python
+import pytest
+from circle import *
```
natasha-aleksandrova: it is a bad practice to use start imports

students/kevin_t/lesson8/test_circle.py
```diff
@@ -0,0 +1,60 @@
...
+def test_step8_compare_circles():
+    c1 = Circle(2)
+    c2 = Circle(4)
+    assert (c1 > c2) == False
+    assert (c1 < c2) == True
+    assert (c1 == c2) == False
+    c3 = Circle(4)
+    assert (c2 == c3) == True
+    circles = [Circle(5), Circle(2), Circle(9), Circle(6), Circle(4), Circle(1)]
+    #assert circles.sort() == [Circle(1), Circle(2), Circle(4), Circle(5), Circle(6), Circle(9)]
```
natasha-aleksandrova: the reason this fails is because `sort()` has not return value (returns `None`) so you are comparing a list of circle objects to None.

students/kevin_t/lesson8/test_circle.py
```diff
@@ -0,0 +1,60 @@
...
+
+def test_step5_create_from_diameter():
+    c = Circle.from_diameter(8)
+    assert c.diameter == 8
+    assert c.radius == 4
+
+def test_step6_print_circle():
+    c = Circle(4)
+    assert str(c) == 'Circle with radius: 4.00000'
+    #assert repr(c) == 'Circle(4)'
```
natasha-aleksandrova: I didn't have issues with this assert line

PR #302, opened 2019-07-23
Lesson09

pkoleyni:
Lesson9/cli_main.py
Lesson9/donor_models.py
Lesson9/test_mailroom_oo.py

### DIFF
--- a/students/pkoleyni/Lesson9/donor_models.py
+++ b/students/pkoleyni/Lesson9/donor_models.py
@@ -1,0 +1,105 @@
+class Donor:
+    def __init__(self, name, donation):
+
+        self.name = name.title()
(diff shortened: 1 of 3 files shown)


students/pkoleyni/Lesson9/donor_models.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+    @add_donation.setter
+    def add_donation(self, new_donation):
+        """
+        setter to add new donation value to the list of all the donations made by
+        specific donor
+        """
+        if Donor.donation_value_check(new_donation):
+            self.donation_list.append(new_donation)
```
natasha-aleksandrova: per my previous comment - this should be a method not property.

PR #306, opened 2019-07-25
Took a bit, and the pairing was a bit clunky

nlenssen2013:


students/Nick_Lenssen/lesson04/triagram.py
```diff
@@ -0,0 +1,88 @@
...
+        follower = words_list[i + 2]
+        if pair not in triagram.keys(): #add a new if the pair is the first time being added to the triagram
+            triagram[pair] = [follower]
+        else:
+            triagram[pair].append(follower)  #append to an existing
+    return triagram
+
+def build_text(triagram):
+    new_text = []
+    tri_keys = list(triagram.keys())
```
natasha-aleksandrova: note that calling `.keys()` is not required

PR #307, opened 2019-07-25
add mailroom1

Jaidjen:


students/Jaidjen/lesson03/Mailroom1.py
```diff
@@ -1,53 +1,60 @@
...
+                    else:
+                        donors.append(new_donors)
+                        donors.append(donation_amt)
+                        print('Dear', new_donors, ',')
+                        print('Thank you', new_donors, 'for donating', '{:.2f}'.format(int(donation_amt)))
+                        print('We appreciate your generosity.')
+                        print()
+                        print('Sincerely, Donation Team')
+                        print()
+                        main()
```
natasha-aleksandrova: remember that we don't want the `main()` call in here - it's already in a while True loop

students/Jaidjen/lesson03/Mailroom1.py
```diff
@@ -1,53 +1,60 @@
...
-        tot_ave= dnr_ave/ave_amt
-        donate.append(dnr_ave)
-        sorted(donate)
-        print("{:20s}{:^20.2f}{:^19d}{:^18.2f}".format(Mail[i][0], dnr_ave, len(Mail[i][1]), tot_ave))
+    summary_data = []
+
+    for donor in donors:
+        donations = donor[1]
+        summary_data.append([((len(donations))),sum(donations)])
+        sorted_data = sorted(summary_data, reverse=True)
```
natasha-aleksandrova: this line needs to be de-dented, meaning moved out of this for loop. there is no need to sort on each loop iteration, you should sort after you're done looping.

students/Jaidjen/lesson03/Mailroom1.py
```diff
@@ -1,53 +1,60 @@
...
+
+    for i in range(len(donors)):
```
natasha-aleksandrova: no need to use range here, you can just loop over donors directly.

Jaidjen:
I tried to remove main() but it won’t go back to the main screen. It keeps looping and asking to add more donors. Did I miss anything in my code that stops it from going back to main() without calling it? 

I will work on the other part when I get home. 
Thank you!

students/Jaidjen/lesson03/Mailroom1.py
natasha-aleksandrova: you need a `break` statement in order to stop your loop

students/Jaidjen/lesson03/Mailroom1.py
```diff
@@ -1,53 +1,60 @@
...
+            if new_donors == "list":
+                print("{:<20s}".format(donors[i][0]))
+                break
+            else:
+                donation_amt = int(input("Please enter the donation amount: "))
+                for donor in donors:
+                    if donor[0] == new_donors:
+                        donors.append()
+                        break
+                    else:
```
natasha-aleksandrova: also, for and else are on the same indentation. `else` should be part of your for loop not if statement.

PR #309, opened 2019-07-27
added lesson 07

esokada:
Thanks as always for your feedback! 

I also have a question (also found in comments):
`class A(OneLineTag):

    tagname = "a"

    def __init__(self, link, content=None, **kwargs):
        kwargs["href"] = link
        super().__init__(content, **kwargs)
        '''
        Question: am I able to call super().__init__ here without giving it the self argument
        because it "already knows" about self at this point?
        (This doesn't seem a very accurate way of explaining it,
        and I'm trying to clarify my understanding.)
        '''`

### DIFF
--- a/students/esokada/lesson07/html_render.py
+++ b/students/esokada/lesson07/html_render.py
@@ -1,0 +1,178 @@
+#!/usr/bin/env python3
+
+"""
+A class-based system for rendering html.
+"""
+
+
+#Step 1
+
+# This is the framework for the base class
+class Element(object):
+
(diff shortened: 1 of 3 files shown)


natasha-aleksandrova:
@esokada for your question above.

In python 2 signatured used to be:
```
super(ClassName, self).__init__()
```

in python3 you no longer need to supply parameters to super(), it uses call stack to find the class instead of you explicitly telling it.

esokada:
Thanks, that makes sense!

PR #315, opened 2019-07-28
Lesson 07 & Lesson 08 Assignments

trvvss:


students/travis_nelson/lesson07/html_render.py
```diff
@@ -0,0 +1,152 @@
...
+    tag = "html"
+    indent = ""
+
+    def __init__(self, content=None, **element_attribute):
+
+        if content is not None:
+            self.contents = [content]
+        else:
+            self.contents = [""]
+        if element_attribute:
```
natasha-aleksandrova: *attributes (should be plural)

students/travis_nelson/lesson07/html_render.py
```diff
@@ -0,0 +1,152 @@
...
+            self.attributes = element_attribute
```
natasha-aleksandrova: I don't think you want the `if` statement - otherwise you won't have `self.attributes` defined. By default if no kwargs are supplied it will be `{}`

students/travis_nelson/lesson07/html_render.py
```diff
@@ -0,0 +1,152 @@
...
+
+    def _open_tag(self):
+        try:
+            open_tag = [f"<{self.tag}"]
+            for k, v in self.attributes.items():
+                open_tag.append(f' {k}="{v}"')
+            open_tag.append(">")
+            return "".join(open_tag)
+        except AttributeError:
+            return f"<{self.tag}>"
```
natasha-aleksandrova: per my comment above, if `self.attributes` will be `{}` then you can remove except clause and your code should work as is

students/travis_nelson/lesson08/circle.py
```diff
@@ -0,0 +1,84 @@
+
+import math
+
+
+class Circle():
```
natasha-aleksandrova: note that in python3, you define a class as `class Circle:`

students/travis_nelson/lesson08/circle.py
```diff
@@ -0,0 +1,84 @@
...
+        if not self._is_valid_operand(other):
+            return NotImplemented
+        return Circle(self.radius + other.radius)
+
+    def __imul__(self, other):
+        return Circle(self.radius * other)
+
+    @classmethod
+    def from_diameter(cls, val):
+        self = cls((val/2))
```
natasha-aleksandrova: this can be simply: `return cls(val / 2)`

PR #319, opened 2019-07-29
Mailroom_2

nlenssen2013:


students/Nick_Lenssen/lesson04/mailroom_2.py
```diff
@@ -0,0 +1,83 @@
...
+                        len(donor_db[key]), '$',round(sum(donor_db[key])/len(donor_db[key]),1))))
+
+def send_thanks():
+    while True:
+        full_name = input("\n\nPlease enter a name for a Thank You to go out or c to go to orginal prompt: ")
+        if full_name == 'list':
+            for key in donor_db:
+                print (key)
+        elif full_name == 'c':
+            main()
```
natasha-aleksandrova: use `return` here - shouldn't call main() directly

students/Nick_Lenssen/lesson04/mailroom_2.py
```diff
@@ -0,0 +1,83 @@
+from sys import exit
+import pathlib
+from operator import itemgetter
+
+def get_donation_amount():
+    while True:
+        amount = input("\nPlease enter in the amount you want to donate or c to go to original prompt: ")
+        if amount == 'c':
+            main()
```
natasha-aleksandrova: also, use `return` here

PR #324, opened 2019-07-30
Added ex 2 assignments

babslee:


students/blee/lesson02/ex02.2_fbz.py
```diff
@@ -0,0 +1,12 @@
+def fbzz(x):
+    for i in range(x+1):
+#this line has to be here, or will have wrong answer at 3 or 5
+        if i%3==0 & i%5==0:
```
natasha-aleksandrova: use `and` instead of `&`

students/blee/lesson02/ex02.3_series.py
```diff
@@ -0,0 +1,83 @@
...
+"""
+
+
+def fibonacci(n):
+    if n == 0:
+        return 0
+    elif n == 1:
+        return 1
+    return fibonacci(n-1) + fibonacci(n-2)
+    pass
```
natasha-aleksandrova: `pass` is not needed here

PR #332, opened 2019-08-03
Lesson 6 Exercises (not graded)

brianminsk:
Intro to Unit Testing exercise and 
args and kwargs lab

__I'm not quite sure I understand the 2nd part of the args and kwargs lab - the part where it asks for a new function with the parameters as *args and **kwargs.  Maybe I don't understand what it is asking for - I feel like my solution might be too complex or I missing the point/concepts. It's my colors1 function in args_kwargs.py__    _Will you take a look, please?_


natasha-aleksandrova:
@brianminsk  your `colors1` function is correct.

the point here was to understand how complicated it may become if you decide to use `*args, **kwargs` signature; there are specific cases for a signature like that but typically you would have more defined params.

PR #333, opened 2019-08-03
Anthony McKeever - Lesson3 Assignment Submission

Snip3rM00n:
Submission for Mailroom assignment in lesson 3.

This was fun, I enjoyed it. :)

_**Squashed Commit History:**_

* Add beginnings of assignment 3

* more assignment 3 work

* Continue Assignment 3

* Finish assignment 1

Refactor the donor summary and donor list to align with assignment
Implement send thanks, mail printing, etc
Minor refactorings to reduce code bloat
Remove user stories.
Correct start date.

* Add DocString to email

### DIFF
--- a/students/anthony_mckeever/lesson3/assignment_1/mailroom.py
+++ b/students/anthony_mckeever/lesson3/assignment_1/mailroom.py
@@ -1,0 +1,295 @@
+"""
+Programming In Python - Lesson 3 Assignment 1: Mailroom Part 1
+Code Poet: Anthony McKeever
+Start Date: 07/30/2019
+End Date: 08/01/2019
+"""
+
+menu_opts = ["main", "return", "stop"]
+quit_opts = ["exit", "end", "quit"]
+list_opts = ["list", "l", "ls"]
+help_opts = ["help", "h", "?"]
+
+donors_list = [("Cresenta Starchelle", [99.99, 6000.00, 10345.23, 29.99]),
+               ("Delilah Matsuka", [199.99, 299.99, 2100.00]),
+               ("Astra Matsume", [599.99]),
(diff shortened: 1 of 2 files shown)


students/anthony_mckeever/lesson3/assignment_1/mailroom.py
```diff
@@ -0,0 +1,295 @@
...
+
+stay_on = True
+
+
+def main():
+    """
+    The main loop of the script.  Prompts user with a main menu.
+    """
+    print("\nStudio Starchelle Donor Appreciation System\n")
+    while stay_on:
```
natasha-aleksandrova: I would recommend simplifying this (especially will matter in lesosn4)

use `while True` and create a quit_program function to call `sys.exit()`

students/anthony_mckeever/lesson3/assignment_1/mailroom.py
```diff
@@ -0,0 +1,295 @@
...
+    :user_choice:   The str().lower value of the user's input selection.
+    """
+    if user_choice in help_opts:
+        print_help()
+    elif user_choice == "create a report":
+        create_report()
+    elif user_choice == "send a thank you":
+        send_thanks()
+    elif user_choice in quit_opts:
+        set_stay_on(False)
```
natasha-aleksandrova: as mentioned above, call `quit_program` here

students/anthony_mckeever/lesson3/assignment_1/mailroom.py
Snip3rM00n: I appreciate the feedback!  I'll definitely simplify the code for lesson 4.

PR #334, opened 2019-08-03
Added Except Exercise and Mailroom Part 3

ajaggarcia10:
I added try-except blocks to the mailroom assignment, and looked through to see if I could use comprehensions, but I didn't see any places where it could be used. If I'm wrong, please let me know that there are places where it is better to use comprehensions, but not exactly where.

### DIFF
--- a/students/andrew_garcia/lesson_05/Mailroom Part 3.py
+++ b/students/andrew_garcia/lesson_05/Mailroom Part 3.py
@@ -1,0 +1,176 @@
+'''
+Andrew Garcia
+Mailroom Part 3
+7/31/19
+'''
+
+# !/usr/bin/env python3
+
+all_donations = {'Jayson Black': [156.80, 207.32, 219.92],
+                 'Bryan Fultan': [1236.28],
+                 'Danica Dolores': [163.51, 100.42, 186.22],
+                 'Skylar Odell': [167.90, 151.62],
+                 'Roy Max': [137.97, 227.63]}
+
+
+def options_menu():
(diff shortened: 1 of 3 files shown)


students/andrew_garcia/lesson_05/Mailroom Part 3.py
```diff
@@ -0,0 +1,176 @@
...
+        else:  # adds donor if they do not exist in dictionary
+            add_donation(name)
+        break
+
+
+def sorting_donors(all_donations=all_donations):
+    """Sorts the all_donations, ordering them in terms of highest total donation amount"""
+
+    sorting_donors = []
+    donor_info = []
```
charlesrobison: This variable isn't used anywhere, so unnecessary.

students/andrew_garcia/lesson_05/Mailroom Part 3.py
```diff
@@ -0,0 +1,176 @@
...
+        all_donations[name.title()] += [float(amount)]
+    elif name.title() not in all_donations:  # adds donor to the dictionary if it does not exist
+        all_donations[name.title()] = [float(amount)]
+    format_thank_you(name, float(amount))
+
+
+def send_thank_you(all_donations=all_donations):  # edit for dict
+    """Gets donor name so that a thank you letter can be formatted"""
+    while True:
+        name = input('Which Donor would you like to send a note to?: ')
```
charlesrobison: Missing the option to list active donors.

PR #336, opened 2019-08-04
Added circle exercise for lesson08

esokada:
Thanks as always for your feedback!

### DIFF
--- a/students/esokada/lesson08/circle.py
+++ b/students/esokada/lesson08/circle.py
@@ -1,0 +1,69 @@
+import math
+
+from functools import total_ordering
+
+
+@total_ordering
+class Circle(object):
+
+    def __init__(self, radius):
+        self.radius = radius
+
+    @classmethod
+    def from_diameter(cls, diameter):
+        radius = diameter / 2
+        return cls(diameter / 2)
+
+    @property
+    def diameter(self):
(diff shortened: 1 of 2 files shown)


students/esokada/lesson08/circle.py
```diff
@@ -0,0 +1,69 @@
...
+
+@total_ordering
+class Circle(object):
+
+    def __init__(self, radius):
+        self.radius = radius
+
+    @classmethod
+    def from_diameter(cls, diameter):
+        radius = diameter / 2
```
natasha-aleksandrova: this can be removed since you're not using `radius`

students/esokada/lesson08/test_circle.py
```diff
@@ -0,0 +1,74 @@
+from circle import *
```
natasha-aleksandrova: avoid using `*` (star) imports; instead `from circle import Circle`

esokada: Right! I'll be more careful in the future.

students/esokada/lesson08/circle.py
esokada: Thanks for catching this! Must have been left over, I don't know what I was thinking.

PR #345, opened 2019-08-07
Lesson01 tasks

Reem0330:
Thank you for your time reviewing my submitted .py files for lesson 01.. I'm bit confused on how I can add the exceptions in if  __name__ =="__main__"

Regards 
Reem

### DIFF
--- a/students/ReemAlqaysi/Lesson01/Lesson01_Exceptions.py
+++ b/students/ReemAlqaysi/Lesson01/Lesson01_Exceptions.py
@@ -1,6 +1,7 @@
 
 #By Reem Alqaysi
 #lesson01
+""" script should calculate Lunch portion size for a reguler day, but it's Break me code """
 
 #This Function will cause NameError exception
 def lunch_NameError():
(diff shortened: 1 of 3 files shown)


students/ReemAlqaysi/Lesson01/Lesson01_Exceptions.py
```diff
@@ -0,0 +1,49 @@
+
+#By Reem Alqaysi
+#lesson01
+""" script should calculate Lunch portion size for a reguler day, but it's Break me code """
+
+#This Function will cause NameError exception
+def lunch_NameError():
```
charlesrobison: Make sure function names (and most other things except classes) are snake_case in Python.  So, def lunch_name_error():  Even though it's giving an example of a NameError.

students/ReemAlqaysi/Lesson01/Lesson01_Exceptions.py
```diff
@@ -0,0 +1,49 @@
...
+ fruit = 1
```
charlesrobison: Always indent with 4 spaces in Python.

students/ReemAlqaysi/Lesson01/Lesson01_Exceptions.py
```diff
@@ -0,0 +1,49 @@
...
+ Sandwitch = 1
+ vegetable = 2
+
+ lunch = fruit + Sandwitch + vegetable
+ print('The function is working fine')
+
+#This Function will cause SyntaxError exception
+def lunch_SyntaxError():
+ fruit = 1
+ Sandwitch = 1
```
charlesrobison: Same here, snake_case, so sandwich = 1

PR #358, opened 2019-08-10
Added Mailroom 4

ajaggarcia10:


students/andrew_garcia/lesson_06/mailroom_test.py
```diff
@@ -0,0 +1,66 @@
...
+
+def test_thank_all_files():
+    """Tests content in a file created"""
+    text = (f'Dear Skylar Odell, \nThank you for your 2 donations, totaling $319.52. \nFrom, Your Local Charity')
+    with open('Skylar Odell.txt', 'r+') as file:
+        fileinfo = ''
+        for line in file:
+            line.strip('\n')
+            fileinfo += line
+    assert fileinfo == text
```
natasha-aleksandrova: you aren't executing any of your mailroom code here - each test should execute the code you're testing.

students/andrew_garcia/lesson_06/mailroom_test.py
```diff
@@ -0,0 +1,66 @@
...
+                                'Mark Wright': [250.36]}
+
+if __name__ == '__main__':
+    test_sorting_donors()
+    test_thank_all()
+    test_thank_all_files()
+    test_format_thank_you()
+    test_add_donation()
+    test_add_donation2()
+    print('Working')
```
natasha-aleksandrova: all of these test functions should be executed with `pytest` instead of as regular python script.

PR #360, opened 2019-08-11
Assignment 7: The Circle Class

nam-vo:


students/nam_vo/lesson08/circle.py
```diff
@@ -0,0 +1,71 @@
...
+        self.radius = int(value / 2)
+
+    @property
+    def area(self):
+        return math.pi * math.pow(self.radius, 2)
+
+    @classmethod
+    def from_diameter(cls, value):
+        self = cls(int(value / 2))
+        return self
```
natasha-aleksandrova: note you can just return `return cls(int(value) / 2)`

PR #372, opened 2019-08-15
Added python pushups and lesson02 print_grid assignment

alexanderboone96:
To clarify on my print_grid assignment, a name wasn't specified for function 1 so I named it step_one_grid().


students/alexander_boone/lesson02/fizz_buzz.py
```diff
@@ -0,0 +1,16 @@
+# Fizz Buzz Program
+for i in range(100):
```
natasha-aleksandrova: should be ``range(1, 101)`

PR #374, opened 2019-08-15
Lesson04

Jaidjen:
Re-submitting Trigrams

### DIFF
--- a/students/Jaidjen/lesson04/Mailroom2.py
+++ b/students/Jaidjen/lesson04/Mailroom2.py
@@ -1,55 +1,73 @@
 import sys  # imports go at the top of the file
 
 
-Mail = {1:{'Name':'James Smith', 'Donation':[2000.25,500,27.50]},
-              2:{'Name': 'Dorothy Parker','Donation':[3175,475.45,92.30]},
-              3:{'Name':'Dohgyu Hwang','Donation':[10670,350.19,175]},
-              4:{'Name': 'Sally Yu', 'Donation':[100,925]},
-              5:{'Name':'Jenny Ramos','Donation':[25.14]}}
-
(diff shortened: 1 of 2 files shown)


students/Jaidjen/lesson04/Mailroom2.py
```diff
@@ -1,77 +1,103 @@
...
+        prompt = int(input("Select a menu option: "))
+
+        if prompt == 1:
+            add_donors()
+        elif prompt == 2:
+            view_donors(donors)
+        elif prompt == 3:
+            thankyou_donors()
+        elif prompt == 4:
+            exit_program
```
natasha-aleksandrova: you should be using a switch dict in lesson 4 - do you need help how to apply that concept to the mailroom?

students/Jaidjen/lesson04/Mailroom2.py
```diff
@@ -1,77 +1,103 @@
...
-          "4 - Quit",
-          ">>> "))
-
-
-def view_mail(Mail):
+donors = [{'Name':'James Smith', 'Donation':[21000.25,500,27.50]},
+          {'Name': 'Dorothy Parker','Donation':[3175,475.45,92.30]},
+          {'Name':'Dohgyu Hwang','Donation':[10670,350.19,175]},
+          {'Name': 'Sally Yu', 'Donation':[100,925]},
+          {'Name':'Jenny Ramos','Donation':[25.14]}]
```
natasha-aleksandrova: you were using a dictionary before, now you're using a list.

this is what your data structure should look like:
```
donors = {"James Smith": [100, 200, 300], "Sally Yu": [202, 22.3]}
```


once you convert this to above dictionary then you can update your look up and update code (it will get much simpler)

PR #376, opened 2019-08-16
Anthony McKeever - Lesson 5 Assignment & Exercise Submission

Snip3rM00n:
This is my submission for lesson 5.  It contains both the graded assignment and the graded exercise.

It took me a little bit to wrap my head around comprehensions, however once I was able to mentally equate them to C#'s Linq and lambda expressions it made it easier to comprehend.  <3  After that I had a lot of fun reducing many of my list and dictionary loops into comprehensions.  My favorite part was where I effectively reduced 9 lines of code down to one.

PS. Exercise 1 contains both the except_exercise.py (modified to handle exceptions) and except_test.py (unmodified from the course material and included as is).

---

**_Squashed Commit History_**

* Initial work for Lesson 5

Add safe input
Add try catch
Try catch exercises

* Bug fix and comprehensions

Fix bug where keyboard interupts and EOF were not exiting
	Replace Finally with Else in the safe_input method
Replace get_legnths with get_length using list comprehensions
Replace get_donor_summary logic with list comprehension
	Convert dictionary into a list of tuples.

* Add header to exceptions exercise

* pre-money conversion commit

* Finalize Assignment

Convert most dictionary and list looping into comprehensions.
Menu selections with multiple keys now expressed via comprehension

create_report logic almost entirely replaced with comprehensions
Formatting floats to money in the summary line instead of in multiple places
get_lengths replaced with get_length - entire lengths collection
expressed via comprehension

* Update read me and minor PEP8 alignment

* Fix header

### DIFF
--- a/students/anthony_mckeever/lesson5/assignment_1/mailroom.py
+++ b/students/anthony_mckeever/lesson5/assignment_1/mailroom.py
@@ -1,0 +1,397 @@
+#!/usr/bin/env python3
+
+"""
+Programming In Python - Lesson 5 Assignment 1: Mailroom Part 3
+Code Poet: Anthony McKeever
+Start Date: 08/13/2019
+End Date: 08/15/2019
+"""
+import os
+import sys
+import tempfile
+
+
+def main():
+    """
+    The main loop of the script.  Prompts user with a main menu.
+    """
+    print("\nStudio Starchelle Donor Appreciation System\n")
+    while True:
+        menu_system(main_menu_dict, "\nMain Menu:", "\nWhat do you want to do? > ")
+
+
+def safe_input(msg):
+    """
+    Return the input from the user or exit upon KeyboardInterupt or EOFError
+
+    :msg:   The message for the input prompt
+    """
+    output = None
+    try:
+        output = input(msg)
+    except (KeyboardInterrupt, EOFError):
+        print("Exiting...")
+        sys.exit()
+    else:
+        return output
+
+
+def menu_system(opts_dict, menu_text, prompt_text, include_main=False, include_donors=False, invalid_opt=None):
+    """
+    The menu system that handles most user multiple choice inputs.
+
+    :opts_dict:         The initial options dictionary.
+    :menu_text:         The flavor text of the menu.
+    :prompt_text:       What to prompt the user when accepting input.
+    :include_main:      Whether or not to include options for returning to the main menu (Default = False)
+    :include_donors:    Whether or not to include the donors with the options. (Default = False)
+                        Note: Donors will not print when menu options are printed.
+    :invalid_opt:       Override the action if the user provides an invalid selection. (Default = None)
+                        Note: Using None will prompt user to reattemt their choice until a correct choice is made.
+    """
+    show_opts = opts_dict.copy()
+    
+    if include_main:
+        show_opts.update(main_dict)
+
+    show_opts.update(quit_dict)
+
+    print(menu_text)
+    print_options(show_opts)
+    
+    # Don't include the list of donors in the help text so the user has to request them.
+    if include_donors:
+        show_opts.update(donor_dict)
+
+    while True:
+        user_choice = safe_input(prompt_text)
+        choice_key = key_from_lower(user_choice, show_opts.keys())
+
+        if choice_key is not None:
+            selection = show_opts[choice_key]
+
+            if selection[1]:
+                if selection[1] == accept_donation:
+                    selection[1](choice_key)
+                else:
+                    selection[1]()
+            
+            if selection[1] != print_donors:
+                break
+
+        else:
+            if invalid_opt is None:
+                print("Invalid choice.  Please select from available options.")
+            else:
+                invalid_opt(user_choice)
+                break
+
+
+def print_options(show_opts):
+    """
+    Print the options from a menu.
+
+    :show_opts:     The complete dictionary of the current menu's options.
+    """
+    skip = set([])
+    for key, value in show_opts.items():
+        if key not in skip:
+            opts_string, skippable = get_opts_string(key, show_opts)
+            print("\t" + opts_string + value[0])
+            skip.update(skippable)
+            
+
+def key_from_lower(user_choice, keys):
+    """
+    Return the user's lower case input to a properly cased dictionary key.
+
+    :user_choice:   The choice the user made at a menu.
+    :keys:          The keys from the dictionary the user is choosing from.
+    """
+    lower_keys = {k.lower(): k for k in keys}
+    return lower_keys[user_choice] if user_choice in lower_keys.keys() else None
+
+
+def get_opts_string(key, dictionary):
+    """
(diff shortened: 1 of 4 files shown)


students/anthony_mckeever/lesson5/assignment_1/mailroom.py
```diff
@@ -0,0 +1,397 @@
...
+
+def key_from_lower(user_choice, keys):
+    """
+    Return the user's lower case input to a properly cased dictionary key.
+
+    :user_choice:   The choice the user made at a menu.
+    :keys:          The keys from the dictionary the user is choosing from.
+    """
+    lower_keys = {k.lower(): k for k in keys}
+    return lower_keys[user_choice] if user_choice in lower_keys.keys() else None
```
natasha-aleksandrova: I don't really understand the purpose of this function, however this can be simplified to 
`return lower_keys.get(user_choice)`

students/anthony_mckeever/lesson5/assignment_1/mailroom.py
```diff
@@ -0,0 +1,397 @@
...
+        return
+    
+    write_dir = user_dir if user_dir is not None else default_dir
+
+    for k, v in donor_dict.items():
+        file_path = os.path.join(write_dir, f"{k}.txt")
+        email = get_email(k, v[0][-1])
+        write_file = open(file_path, "w")
+        write_file.write(email)
+        write_file.close()
```
natasha-aleksandrova: use `with` statement when working with files

students/anthony_mckeever/lesson5/assignment_1/mailroom.py
```diff
@@ -0,0 +1,397 @@
...
+
+    donor[0].append(donation_float)
+    email = get_email(donor_name, donation_float)
+
+    print("\n\n----- PLEASE SEND THIS EMAIL TO THE DONOR -----\n\n")
+    print(email)
+    print("\n\n----- PLEASE SEND THIS EMAIL TO THE DONOR -----\n\n")
+
+
+def send__to_all():
```
natasha-aleksandrova: single underscores in the function names.

students/anthony_mckeever/lesson5/assignment_1/mailroom.py
```diff
@@ -0,0 +1,397 @@
...
+
+quit_dict = {"exit": ("\t\tQuit the script.  Can be used at any input prompt.", sys.exit),
+             "end":  ("\t\tQuit the script.  Can be used at any input prompt.", sys.exit),
+             "quit": ("\t\tQuit the script.  Can be used at any input prompt.", sys.exit),
+             }
+
+            
+list_dict = {"list": ("\t\tPrint a list of available donors.", print_donors),
+             "l":    ("\t\tPrint a list of available donors.", print_donors),
+             "ls":   ("\t\tPrint a list of available donors.", print_donors)
```
natasha-aleksandrova: it's best for above to be number options:
```
1 - send thank you
2 - create report
...
```

having the user type out the selection is error prone and not very user friendly.

natasha-aleksandrova: note that per directions, the program should ask to "list" or enter a donor names - user shouldn't be required to execute "list" command each time they would like to make a donation.

students/anthony_mckeever/lesson5/assignment_1/mailroom.py
Snip3rM00n: That is much easier to read, I'll be sure to implement this if this function is still needed when I implement the numbering for the menu system.

students/anthony_mckeever/lesson5/assignment_1/mailroom.py
Snip3rM00n: I'll implement this.  :)

students/anthony_mckeever/lesson5/assignment_1/mailroom.py
Snip3rM00n: Good catch, I'll remove the double underscore.

students/anthony_mckeever/lesson5/assignment_1/mailroom.py
Snip3rM00n: > it's best for above to be number options:
> 
> ```
> 1 - send thank you
> 2 - create report
> ...
> ```
> 
> having the user type out the selection is error prone and not very user friendly.

I'll get this implemented. :)

Snip3rM00n: > note that per directions, the program should ask to "list" or enter a donor names - user shouldn't be required to execute "list" command each time they would like to make a donation.

@natasha-aleksandrova I'm a little confused by this.  I looked back through Mailroom parts 1 through 3 and the only instructions I for listing donor's is in Mailroom Part 1 under the "Send a Thank You" section which states:

> If the user types list show them a list of the donor names and re-prompt.

Mailroom Part 2 and Part 3 do not update this requirement.  I am absolutely happy to modify the code to prompt the user if they want to list before thanking the donor, but can you confirm that this is the correct and desired behavior?

Snip3rM00n:
@natasha-aleksandrova - Thank you for your feedback on my PR.  😄  I have submitted a new PR implementing the feedback you have given me.  You can find it here:

https://github.com/UWPCE-PythonCert-ClassRepos/SP_Online_PY210/pull/381

students/anthony_mckeever/lesson5/assignment_1/mailroom.py
natasha-aleksandrova: @Snip3rM00n I just saw this comment. 

The instructions are:

>  If the user types list show them a list of the donor names and re-prompt.
    If the user types a name not in the list, add that name to the data structure and use it.
    If the user types a name in the list, use it.


meaning, that those are all 3 valid options that a user can execute from "send a thank you" option. your code was requiring the user to always enter list first.

Snip3rM00n: > @Snip3rM00n I just saw this comment.
> The instructions are:
> 
> If the user types list show them a list of the donor names and re-prompt.
> If the user types a name not in the list, add that name to the data structure and use it.
> If the user types a name in the list, use it.
> 
> meaning, that those are all 3 valid options that a user can execute from "send a thank you" option. your code was requiring the user to always enter list first.

Ah, I see where the confusion is coming from.  My original code actually met these requirements, the user didn't have to enter a `list` command before being able to apply a donation to an existing donor.  The comment I had in my code on line 62:

> `# Don't include the list of donors in the help text so the user has to request them.`

was intended to mean the donor names do not appear in the list menu commands, so if they wanted to view them they could use a list command to see them.  The user could have entered `send a thank you` and then entered an existing donor like `Cresenta Starchelle` and apply a donation from that donor without listing them first.  I used this scenario for my golden path/build verification testing.

With that said, I actually really prefer the refactor makes things much more clear and removed a good amount of the complexity around it, which is always good.  Plus refactoring is always fun. 😄

PR #382, opened 2019-08-18
Re-committing mail_room_pt2_revised--3x's a charm?

CCSt130:
Thank you for reviewing. I have changed my code so that donors may now have more than one new donation added per day (any count is possible). Please note, given my revised data structure, a donation amount may not be duplicated on the same day for any donor, but I hope that you find that satisfactory.


CCSt130:
Roger, understood. I apologize if my unnecessary complication has caused you extra work.

PR #385, opened 2019-08-20
adding in lesson04

tsulgrove:


students/thomas_sulgrove/lesson04/mailroom.py
```diff
@@ -0,0 +1,130 @@
...
+
+##### Global Variables
+
+#Dictionaries
+donor_db = {
+            "William Gates, III": [653772.32, 12.17],
+            "Jeff Bezos": [877.33],
+            "Paul Allen": [663.23, 43.87, 1.32],
+            "Mark Zuckerberg": [1663.23, 4300.87, 10432.0],
+            }
```
natasha-aleksandrova: 👍

students/thomas_sulgrove/lesson04/mailroom.py
```diff
@@ -0,0 +1,130 @@
...
+    #import db
+    #global donor_db
+    #keep doing this until further notice
+    while True:
+        #ask for name or to print list of names
+        donor_name = input(thank_you_prompt)
+        #print out names
+        if donor_name == 'list':
+            for key, value in donor_db.items():
+                print(key)
```
natasha-aleksandrova: since you're not using "values" you can loop like this:
```
for name in donor_db:
    print(name)
```

students/thomas_sulgrove/lesson04/mailroom.py
```diff
@@ -0,0 +1,130 @@
...
+    
+    for key, value in donor_db.items():
+        
+        if response == 'no':
+            file_name = key + '.txt'
+        else:
+            file_name = response + key + '.txt'
+        
+        
+        with open(file_name, 'w') as file:
```
natasha-aleksandrova: avoid using built-in names for variable names (`file`) - notice github highlights it a different color.

students/thomas_sulgrove/lesson04/mailroom.py
```diff
@@ -0,0 +1,130 @@
...
+    #sum up donations for sortation
+    return sum(db[1])
+
+##Switchs
+main_menu = {
+        "1 - Create a report" : create_a_report,
+        "2 - Send thank you letter to one donor" : send_thank_you_one,
+        "3 - Send thank you letter to all donors" : send_thank_you_all,
+        "4 - Exit" : exit_program
+        }
```
natasha-aleksandrova: the idea was to use a switch dict, where a selection would map to a key in the dictionary and then the value would be executed as a function. 

you're not currently utilizing the keys in main_menu to create a switch dict - this might as well be a tuple instead of a dict.

PR #386, opened 2019-08-20
Added Circle

ajaggarcia10:


students/andrew_garcia/lesson_08/Circle.py
```diff
@@ -0,0 +1,64 @@
...
+    def diameter(self, diameter):
+        self.radius = diameter/2
+
+    @property
+    def area(self):
+        return round(math.pi * (self.radius ** 2), 3)
+
+    @classmethod
+    def from_diameter(cls, diameter):
+        cls.radius = diameter/2
```
natasha-aleksandrova: this is incorrect, should be `return cls(diameter / 2)`

this is typically used as an alternate constructor like:
```
circle = Circle.from_diameter(10)
```

the way you currently have it will not satisfy above use case.

PR #388, opened 2019-08-20
Lesson 9 hw

KRTstudent:
Object oriented is way better!

### DIFF
--- a/students/kevin_t/lesson9/cli_main.py
+++ b/students/kevin_t/lesson9/cli_main.py
@@ -1,0 +1,81 @@
+from mailroom_OO import Donor as D
+from mailroom_OO import DonorCollection as DC
+import sys
+
+#Opening prompt function
+def opening_screen():
+    #Tell the user what their options are. Ask for an input.
+    print('')
+    print('Welcome to the mailroom tool. Please select an action:')
+    print('1 - Send a Thank You')
+    print('2 - Create a Report')
+    print('3 - Send letters to all donors')
+    print('4 - Quit')
+    #Send user input back to main function to be analyzed.
+    return input()
+
+def list_donors(donor_collection):
+    #Print each existing donor's name
+    donor_collection.list_donors
+
+#'Send thank you' function
+def send_thank_you(donor_collection):
+    #Ask for a donor's name. If user requests 'list', display existing donors and request action agin.
+    while True:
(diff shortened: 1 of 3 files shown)


students/kevin_t/lesson9/cli_main.py
```diff
@@ -0,0 +1,81 @@
+from mailroom_OO import Donor as D
+from mailroom_OO import DonorCollection as DC
```
natasha-aleksandrova: package names should be all lower case `mailroom_oo` (file names)

also, I would avoid aliases generally (there are uses cases for this) - `Donor` is more clear than `D`

students/kevin_t/lesson9/mailroom_OO.py
```diff
@@ -0,0 +1,46 @@
+class Donor:
+    def __init__(self, donor_name, donations = []):
```
natasha-aleksandrova: using mutable objects for default arg value can be very dangerous:
https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html

students/kevin_t/lesson9/mailroom_OO.py
```diff
@@ -0,0 +1,46 @@
...
+
+class DonorCollection:
+    def __init__(self):
+        self.donor_list = []
+
+    def new_donor(self, new_donor: Donor):
+        self.donor_list.append(new_donor)
+
+    def new_donation(self, donor_name, donation_amount):
+        for donors in self.donor_list:
```
natasha-aleksandrova: should be `for donor in` (singular)

students/kevin_t/lesson9/test_mailroom.py
```diff
@@ -0,0 +1,100 @@
+import pytest
+from mailroom_OO import *
+from cli_main import *
```
natasha-aleksandrova: avoid star imports

students/kevin_t/lesson9/test_mailroom.py
```diff
@@ -0,0 +1,100 @@
...
+import os
+from os import listdir
+
+def test_Donor_initial():
```
natasha-aleksandrova: method names should be lower case

PR #389, opened 2019-08-21
added lesson 09 oo mailroom assignment

esokada:
Thanks as always for your feedback.  I also have a question: 

1.  I've added error handling to both the donor models and the client interaction code.  For example, I make sure the donation is a number in the send_thankyou() function in the client interaction code, and in the Donor class itself as well.  Is this too much? Is it preferable to do this in just one place or the other?

### DIFF
--- a/students/esokada/lesson09/donor_models.py
+++ b/students/esokada/lesson09/donor_models.py
@@ -1,0 +1,102 @@
+
+'''
+Donor models for the object-oriented mailroom assignment.
+'''
+
+from operator import itemgetter
+
+
+class Donor():
+    '''
+    Class for individual donors. Includes name, donations, and other information
+    pertaining to a single donor.
+    '''
+
+    def __init__(self, name, donations=None):
+        if isinstance(name, str):
+            self.name = name
+        else:
+            raise TypeError("name must be entered as a string")
+        if donations is None:
+            self.donations = []
+        else:
+            if isinstance(donations, list) is False:
+                raise TypeError("donations must be entered as a list")
+            for i in donations:
+                if isinstance(i, (int, float)) is False:
+                    raise ValueError("donation must be entered as a number")
+            self.donations = donations
+
+    def add_donation(self, donation):
+        '''Add a donation to the donor's list of donations.'''
+        if isinstance(donation, (int, float)) is False:
+            raise ValueError("donation must be entered as a number")
+        self.donations.append(float(donation))
+
+    @property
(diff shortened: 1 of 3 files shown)


natasha-aleksandrova:
@esokada great question. This will be case by case but in your case it isn't too much.

Your user interaction code should check for any invalid user input and provide feedback. Your "library" code should also have error handling to ensure that no "bad data" is stored because your library code may not know who is using it and how. Hopefully this makes sense.

students/esokada/lesson09/donor_models.py
```diff
@@ -0,0 +1,102 @@
...
+        '''Add a new donor to the collection.'''
+        self.collection.append(Donor(name, donations))
+
+    def get_donor(self, name):
+        '''Return the donor if present, otherwise return False.'''
+        for a in self.collection:
+            if a.name == name:
+                return a
+        else:
+            return False
```
natasha-aleksandrova: best to return `None` here since the function isn't expected to return a boolean (best not to mix types)

students/esokada/lesson09/donor_models.py
```diff
@@ -0,0 +1,102 @@
...
+        '''Generates a formatted report of donors, total given, num gifts, and average gift.'''
+        final_report = ""
+        titles = ["Donor Name", "Total Given", "Num Gifts", "Average Gift"]
+        formatted_titles = f"{titles[0]:15} {titles[1]:>20} {titles[2]:>10} {titles[3]:>20}\n"
+        final_report += formatted_titles
+        reportlines = []
+        for a in self.collection:
+            try:
+                reportlines.append([a.name, a.total_donations,
+                len(a.donations), a.total_donations/len(a.donations)])
```
natasha-aleksandrova: indentation here is off

students/esokada/lesson09/donor_models.py
```diff
@@ -0,0 +1,102 @@
...
+            except ZeroDivisionError:
+                reportlines.append([a.name, 0, 0, 0])
+        sortedlines = sorted(reportlines, key=itemgetter(1), reverse = True)
+        for r in sortedlines:
+          final_report += f"{r[0]:15} {r[1]:>20.2f} {r[2]:>10} {r[3]:>20.2f}\n"
```
natasha-aleksandrova: indentation here is off as well

esokada:
Thanks for your feedback and especially the clarification about error handling - really helps.

PR #399, opened 2019-08-25
Corrected the lesson name to lesson 2. Also added the exersise 4.3

firstzion:
I just realized that I had the lesson number work on the folder, so I fixed that. And added the sum_series for exercise 2.4


students/chrissp/lesson02/series.py
```diff
@@ -0,0 +1,65 @@
...
+    """ This function will compute the nth Fibonacci number. """
+    return sum_series(n, 0, 1)
+
+
+def lucas(n):
+    """ This function will compute the nth Lucas number. """
+    return sum_series(n, 2, 1)
+
+
+def sum_series(n, n0 = 0, n1 = 1):
```
natasha-aleksandrova: note that there should be no spaces aroun `=` operator
`def sum_series(n, n0=0, n1=1)`

PR #412, opened 2019-08-30
HTML Renderer take 3 & Circle Class

theseanhodges:
HTML Renderer: Address comments from [PR 409](https://github.com/UWPCE-PythonCert-ClassRepos/SP_Online_PY210/pull/409)
Circle Class submission


students/shodges/lesson07/html_render.py
```diff
@@ -26,34 +26,31 @@ def render(self, out_file, current_indent=''):
...
         out_file.write("{}</{}>\n".format(current_indent, self.tag))
 
 
-    def print_open_tag(self, out_file, newline=True):
+    def print_open_tag(self, out_file, newline=True, selfclose=False):
         out_file.write("<{}".format(self.tag))
         for k, v in self.elem_attributes.items():
             out_file.write(" {}=\"{}\"".format(k, v))
-        if isinstance(self, SelfClosingElement):
+        if selfclose is True:
```
natasha-aleksandrova: this is still the same - any "additional" functionality should be in it's own class that extends from `Element`

Here are the instructions:
```
Step 5:
Create a SelfClosingTag subclass of Element, to render tags like:

<hr /> and <br /> (horizontal rule and line break).
(See: https://www.w3schools.com/tags/tag_hr.asp)

For example you should be able to use this code:

Hr(width=400)
To get this result:

<hr width="400" />
You will need to override the render method to render just the one tag and attributes, if any.

Note that self closing tags can’t have any content. Make sure that your SelfClosingTag element raises an exception if someone tries to put in any content – probably a TypeError.

Create a couple subclasses of SelfClosingTag for <hr /> and <br />

Note that you now have maybe three render methods – is there repeated code in them?

Can you refactor the common parts into a separate method that all the render methods can call? And do all your tests still pass (you do have tests for everything, don’t you?) after refactoring?
```
which means that you should have a `SelfClosingTag` class that subclasses from `Element`. Then `Br` and `Hr` classes that subclass from `SelfClosingTag`. `SelfClosingTag` should have a `render` method that its subclasses can use.

students/shodges/lesson08/circle.py
```diff
@@ -0,0 +1,143 @@
+#!/usr/bin/env python3
+
+import math
+
+class Circle(object):
+    _radius = None
+    _diameter = None
```
natasha-aleksandrova: these should not be class attributes

students/shodges/lesson07/html_render.py
theseanhodges: @natasha-aleksandrova Sorry I'm not getting this!  Both SimpleElement and SelfClosingElement have render methods and the Element class isn't aware of either.  The Element class has a method print_open_tag which (now) has defaulted arguments to specify whether a newline is appended after the '>' or if the self closing tag '/' is appended.  Also per Step 5 common code should be refactored into methods render can call. :)

Initially I had this done with the isinstance call but per your initial comment that had the class aware of its subclasses.

If that's not how I should be doing this let me know and I can drop the code into each subclasses' render method.  Thanks!

students/shodges/lesson08/circle.py
theseanhodges: @natasha-aleksandrova Fixed, thanks.  Submitted [PR 419](https://github.com/UWPCE-PythonCert-ClassRepos/SP_Online_PY210/pull/419).

natasha-aleksandrova:
@theseanhodges thanks, will check the new PR

students/shodges/lesson07/html_render.py
natasha-aleksandrova: Hi @theseanhodges 

this line `if selfclose is True:` inside of `Element` class makes element aware that selfclose something exists. `Element` is base class and should not be aware of whatever could be subclassing it.

theseanhodges: @natasha-aleksandrova makes sense, thanks.  I submitted [PR 420](https://github.com/UWPCE-PythonCert-ClassRepos/SP_Online_PY210/pull/420) to address this.

PR #415, opened 2019-09-01
lesson02 exercises

oletos7j:


students/asotelo/lesson02/series.py
```diff
@@ -0,0 +1,80 @@
...
+Python 3 required
+Requirement: https://uwpce-pythoncert.github.io/PythonCertDevel/exercises/fib_and_lucas.html
+'''
+
+def fibonacci(n):
+    """ compute the nth Fibonacci number """
+    if n <= 1:
+        return n
+    else:
+        return(fibonacci(n-2) + fibonacci(n-1))
```
natasha-aleksandrova: `return` is a statement not a function, so no need for parens with it

PR #426, opened 2019-09-07
Lesson 3 labs exercises

calvinpy:
Lesson 3 labs exercises


students/calvin_fannin/lesson03/slicing_lab.py
```diff
@@ -0,0 +1,37 @@
+def exchange_first_last(seq):
+    return seq[-1:] + seq[1:-1] +seq[0:1]
```
natasha-aleksandrova: note that zero (0) is default and you don't need to specify it

PR #428, opened 2019-09-09
Assignment 5 - Lesson 6

bplanica:
Hi Charles - Please let me know where I can make any improvements. I had to restructure a bit when adding the test file so I feel like I am missing a test but cant tell for sure.

Also I had a random question - when creating functions that manipulate a data structure (like the donor dictionary), is it better to make the data structure an argument (and then work with it like a local in the function and then return it) or to have it how I currently do where it isn't an argument but is manipulated in the function. I suppose I realized I had difficulty testing anything that did not "return". Are either of these ways considered "best practice"?

### DIFF
--- a/students/bplanica/lesson06/mailroom4.py
+++ b/students/bplanica/lesson06/mailroom4.py
@@ -14,12 +14,7 @@
 
 # ----- DATA ----- #
 # ---------------- #
-donors = {"bill gates": ("Bill Gates", [326892.24, 326892.24]),
-"mark zuckerberg": ("Mark Zuckerberg", [5465.39, 5465.39, 5465.39]),
-"jeff bezo": ("Jeff Bezo", [877.33]),
-"paul allen": ("Paul Allen", [236.14, 236.14, 236.14]),
-"m bezo": ("M Bezo", [110463.25])} # holds original data set to start with
-
+donors = {}
 dict_menu = {} # holds menu options (dispatch)
 str_menu = ("""\n--------------------
 Choose an action:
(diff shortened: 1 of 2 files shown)


natasha-aleksandrova:
@bplanica great question!

having a function take an argument is not aways an option, it is normal to have a function that manipulates data and there are couple options for testing functions like that. For more advanced code you would use a library called `mock`, for our mailroom just asserting that donors dictionary has the new data is sufficient.

students/bplanica/lesson06/mailroom4.py
```diff
@@ -0,0 +1,226 @@
...
+    update_list("Paul Allen", 236.14)
+    update_list("M Bezo", 110463.25)
+
+
+def main():
+    """ main function """
+    while True: # display a menu of choices to the user
+        try:
+            response = int(input(str_menu)) # refers to preset menu declared under Data
+            if response == 4: raise SystemExit
```
natasha-aleksandrova: you should map this to a function:

```
def quit_program():
    sys.exit()
```

students/bplanica/lesson06/mailroom4.py
```diff
@@ -0,0 +1,226 @@
...
+            dict_menu.get(response)()
+        except TypeError:
+            print("\nPlease choose an action from the above list.")
+            print("Type the menu number associated with the action you would like to perform.")
+        except ValueError:
+            print("\nPlease choose an action from the above list.")
+            print("Type the menu number associated with the action you would like to perform.")
+
+
+def send_thank_you(name = ""):
```
natasha-aleksandrova: why does this function have an input param? I don't think it makes sense here

students/bplanica/lesson06/mailroom4.py
```diff
@@ -0,0 +1,226 @@
...
+    elif fq == 3 or (fq % 10 == 3 and fq != 13):
+        fq = (f"{fq}rd") # format with 'rd'
+    else:
+        fq = (f"{fq}th") # format with 'th'
+    return fq
+
+
+def total_given(sub_list):
+    """ calculates the total amount donated by a donor """
+    total_sum = 0
```
natasha-aleksandrova: no need for this to be initialized

PR #443, opened 2019-09-16
Object-oriented mailroom

theseanhodges:


students/shodges/lesson09/cli_main.py
```diff
@@ -0,0 +1,180 @@
+#!/usr/bin/env python3
+
+from donor_models import *
```
natasha-aleksandrova: avoid using star imports - it's a bad practice

students/shodges/lesson09/donor_models.py
```diff
@@ -0,0 +1,226 @@
...
+
+    @property
+    def count(self):
+        """
+        Returns the number of donations in the donor's ledger.
+        """
+        return len(self.ledger)
+
+
+    @property
```
natasha-aleksandrova: class methods only need one line space in between

students/shodges/lesson09/donor_models.py
```diff
@@ -0,0 +1,226 @@
...
+        return sum(self.ledger)
+
+
+    def process(self, donation):
+        """
+        Appends donation to the donor's ledger.  Returns True if successful.  No exceptions
+        raised are caught and are required to be caught by the implementing code.
+        """
+        self.ledger.append(float(donation))
+        return True
```
natasha-aleksandrova: why return True? not needed here

students/shodges/lesson09/donor_models.py
```diff
@@ -0,0 +1,226 @@
...
+
+    @property
+    def donors(self):
+        """
+        Returns a list of all donors in the database.
+        """
+        donorlist = []
+        for k in self.db.keys():
+            donorlist.append(k)
+        return donorlist
```
natasha-aleksandrova: can you use a comprehension here 😉

students/shodges/lesson09/donor_models.py
theseanhodges: You're right.  For some things I had been checking the return but not here (although I did in a few unit tests so those are fixed).

students/shodges/lesson09/donor_models.py
theseanhodges: Done!  Submitted changes in #450

PR #445, opened 2019-09-17
Committed Lesson 4 files

pbehrend:


students/philip_behrend/lesson04/mailroom_pt2.py
```diff
@@ -0,0 +1,101 @@
...
+                    [[50,40],[100,1000,10000],
+                     [2],[5,0.5,.05],[3,6,9]])
+
+# Avoided separate functions within else statements since there are nuanced differences
+def send_thanks(donor_dict):
+    name_response = input("Type full name: ")
+    while True:
+        donors = donor_dict.keys()
+        if name_response == "list":
+            print(list(donors))
```
charlesrobison: Getting same error here as in prior version - un-ending loop.

PR #459, opened 2019-09-23
Lesson5

calvinpy:
Could you point out some places to implement comprehensions, I did  not see any good candidates in my mailroom code.


students/calvin_fannin/mailroom/test_mailroom.py
```diff
@@ -0,0 +1,86 @@
...
+
+def test_get_next_key():
+    expected = 7
+    assert mr.get_next_key() == expected
+
+
+def test_search_fullname_not_found():
+    """ Search for fullname NOT in dataset """
+    donorname = "JeffY Kingbird"
+    assert mr.search_fullname(donorname) == False
```
natasha-aleksandrova: use `is` with object comparison like True/False, None

students/calvin_fannin/mailroom/test_mailroom.py
```diff
@@ -0,0 +1,86 @@
...
+    ['Calvin Fannin', 645.0, 1, 645.0]]
+    assert mr.create_summary() == expected
+
+
+def test_add_donation_new_donor():
+    """ Test adding a donation to new donor """
+    newdonor = "Vera Fannin"
+    amount = 745.0
+    expected = {'name':'Vera Fannin', 'amount':[745.0]}
+    newIndexdonor = 7
```
natasha-aleksandrova: make sure snake_case for var names -> new_index_donor

students/calvin_fannin/mailroom/mailroom.py
```diff
@@ -142,16 +145,22 @@ def create_report():
...
     main_menu = {1:send_thankyou, 2:create_report, 3:write_letters_all_donors, 4:quit_program}
     while True:
-        response = int(input(prompt))
-        main_menu.get(response,invalid_option)()
+        try:
+            response = int(input(prompt))
+        except ValueError:
+            print('Please enter a  int number')
+        else:
+            main_menu.get(response,invalid_option)()
```
natasha-aleksandrova: this can also be a good place to add try/except

natasha-aleksandrova:
it's ok if you don't find a place to use comprehensions

PR #472, opened 2019-10-06
lesson 2 exercises 

jlmts:
exercise 2.2 and 2.3

### DIFF
--- a/students/jlmts/lesson02/fizz_buzz.py
+++ b/students/jlmts/lesson02/fizz_buzz.py
@@ -1,0 +1,24 @@
+# lesson 2.3: fizz buzz exercise, J. Umetsu
+
+# program to print numbers from 1 to 100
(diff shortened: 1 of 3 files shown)


students/jlmts/lesson02/fizz_buzz.py
```diff
@@ -0,0 +1,24 @@
...
+        print('FizzBuzz')
+        return fizz_buzz(n+1)
+    elif n%3==0:
+        print('Fizz')
+        return fizz_buzz(n+1)
+    elif n%5==0:
+        print('Buzz')
+        return fizz_buzz(n+1)
+        
+fizz_buzz(1)
```
natasha-aleksandrova: this is better solved by using range() function:
```
for i in range(1, 101):
    # your if statements here
```

PR #475, opened 2019-10-08
Lesson 7 Assignment

calvinpy:


students/calvin_fannin/lesson07/html_render.py
```diff
@@ -0,0 +1,136 @@
...
+
+    def __init__(self, content=None, **kwargs):
+        if content is None:
+            self.content = []
+        else:
+            self.content = [content]
+        self.attrb = copy.deepcopy(kwargs)
+
+
+    def _create_open_tag(self,out_file,cur_ind=""):
```
natasha-aleksandrova: only need one line space between class methods

PR #479, opened 2019-10-08
adding lesson 3 files

Reem0330:
thank you for your time to review my work, still kind of struggling working with tuples

### DIFF
--- a/students/ReemAlqaysi/Lesson03/list_lab.py
+++ b/students/ReemAlqaysi/Lesson03/list_lab.py
@@ -1,0 +1,73 @@
+#!/usr/bin/env python3
+
+def series1():
+    mylist =  ['Apples', 'Pears', 'Oranges', 'Peaches']
+    print(mylist)
+    #Ask the user for another fruit and add it to the end of the list.
+    response = input("add another fruit please: ")
+    mylist.append(response)
+    print(mylist)
+    response2 = input("Provide a number between 1 and {:d}: ".format(len(mylist)))
+    print(response2 + " is the index number for : " + mylist[int(response2)-1])
+    #Add another fruit to the beginning of the list using “+” and display the list.
+    mylist =  mylist +['Banana']
+    print(mylist)
+    #Add another fruit to the beginning of the list using insert() and display the list.
+    mylist.insert(0,'Mango')
+    #Display all the fruits that begin with “P”, using a for loop.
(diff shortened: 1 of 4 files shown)


natasha-aleksandrova:
@Reem0330 can you let us know your challenges with tuples? be happy to clarify 🙂

students/ReemAlqaysi/Lesson03/list_lab.py
```diff
@@ -0,0 +1,73 @@
...
+    print(mylist)
+    #Add another fruit to the beginning of the list using insert() and display the list.
+    mylist.insert(0,'Mango')
+    #Display all the fruits that begin with “P”, using a for loop.
+    print("Display all the fruits that begin with “P”, using a for loop.")
+    for name in mylist:
+        if name[0].upper() == "P":
+            print(name)
+        else:
+            continue
```
natasha-aleksandrova: no need for this continue here - it will continue the loop on its own

Reem0330:
> @Reem0330 can you let us know your challenges with tuples? be happy to clarify 🙂

still confused on when I can use it and how to append to it
especially  if there is no specific instruction on what to use in the code list vs tuple.
to add values last time += didn't work for me, so I either change the list to list , then tuple again or as I did in one of the exercises I move it to a variable type list then move it back to the tuple in a dict.. not sure if what I'm doing is a best practice or not.

thank you for quick response
Reem

natasha-aleksandrova:
ah yes! great questions.

Compared to a list, tuple is immutable data structure which you means you *cannot* change it.
Once you define something as a tuple its size cannot be changed. You use a tuple to ensure that it cannot be changed and to in a way "communicate" that it cannot be changed to reduce bugs and to  avoid from people changing it (if someone is using your library).

For example, in the mailroom assignment there is a good place for a tuple.

```
donors = [('Bob', [100, 200])]
```
above is a hard definition of your donor data structure, you can add another donor, but you are ensuring donor is a two item tuple with a name and a list of donations. This will prevent anyone that wants to make changes to the code from breaking that contract.


So going forward when you need to choose a data structure - ask yourself "is this data expandable? can I append to it? or will it never change?"

hope this helps!

Reem0330:
Oh, that really makes sense now, I appreciate your time explaining the difference.. thank you so much 

Regards
Reem

natasha-aleksandrova:
glad to hear!

PR #482, opened 2019-10-10
Updated to Lesson 3 assignments

Eric74nyc:
Updated assignments based on your last feedback. Thank you for the guidance!


Eric74nyc:
Added list_lab updates per instructor request

Eric74nyc:
Includes updated mailroom and listlab assignments

PR #486, opened 2019-10-11
assignment 7 the circle class

calvinpy:


students/calvin_fannin/lesson08/ultimate_circle.py
```diff
@@ -0,0 +1,113 @@
+#!/usr/bin/env python3
+"""
+A Class to create a circle
+"""
+import math
+
+class Circle():
```
natasha-aleksandrova: note that in python3 you can just define class as `class Circle:`

PR #487, opened 2019-10-12
Assignment for Lesson 8

bplanica:
Hello - I included by file with the class/subclass and the test file. Please let me know if anything can be improved. Thanks! Breeanna

### DIFF
--- a/students/bplanica/lesson08/circle.py
+++ b/students/bplanica/lesson08/circle.py
@@ -24,12 +24,12 @@ class Circle(object):
 
     @property
     def area(self):
-    """calculates the area of a circle"""
+        """calculates the area of a circle"""
         return round(pi * (self._radius ** 2), 4)
 
     @classmethod
     def from_diameter(cls, diameter):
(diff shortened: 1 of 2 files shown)


students/bplanica/lesson08/circle.py
```diff
@@ -0,0 +1,93 @@
+from math import pi
+
+class Circle(object):
```
natasha-aleksandrova: note that in python3 you don't need to extend from object -> `class Circle:`

students/bplanica/lesson08/circle.py
```diff
@@ -0,0 +1,93 @@
...
+        return area
+
+    @property
+    def area(self):
+        """calculates the surface area of a sphere"""
+        area = round(4 * pi * (self._radius ** 2), 4)
+        return area
+
+    def __str__(self):
+        return (f"Sphere with a radius: {self._radius}")
```
natasha-aleksandrova: no need to use parens with `return` statement

PR #488, opened 2019-10-12
Added L07 assignement and revised L05 exercise

pbehrend:
I'm having trouble with the multiple_indent function in L07. There appears to be an extra </body> tag for some reason and I haven't been able to figure out why.


natasha-aleksandrova:
@pbehrend ok, will take a look at your issue.

students/philip_behrend/lesson07/html_render.py
```diff
@@ -0,0 +1,184 @@
...
+    def __init__(self, content=None, **kwargs):
+        self.content = content
+        if self.content is not None:
+            raise TypeError
+        self.kwargs = kwargs
+
+
+    def render(self,out_file,cur_ind=""):
+        out_file.write("<{} ".format(self.tag))
+        if self.kwargs != {}:
```
natasha-aleksandrova: no need for this check, just loop through the dict and if dict is empty then it won't do anything

students/philip_behrend/lesson07/html_render.py
```diff
@@ -0,0 +1,184 @@
...
+            out_file.write("{}=\"{}\"".format(key, value))
+        out_file.write(">\n")
+        for item in self.content:
+            try:
+                item.render(out_file, cur_ind+self.ind)
+            except AttributeError:
+                out_file.write(item)
+                out_file.write("\n")
+        out_file.write(cur_ind)
+        out_file.write("</{}>\n".format(self.tag))
```
natasha-aleksandrova: you should be able to use render from Element class

students/philip_behrend/lesson07/html_render.py
```diff
@@ -0,0 +1,184 @@
...
+    def render(self, out_file, cur_ind = ""):
+        out_file.write("<{}>\n".format(self.tag))
+        Meta(charset="UTF-8").render(out_file)
+        for item in self.content:
+            try:
+                item.render(out_file, cur_ind+self.ind)
+            except AttributeError:
+                out_file.write(cur_ind)
+                out_file.write(item)
+        out_file.write("\n</{}>\n".format(self.tag))
```
natasha-aleksandrova: this should be using parent render() method.

students/philip_behrend/lesson07/html_render.py
```diff
@@ -0,0 +1,184 @@
...
+
+
+class OneLineTag(Element):
+    tag = 'title'
```
natasha-aleksandrova: this class doesn't really have a tag

students/philip_behrend/lesson07/html_render.py
```diff
@@ -0,0 +1,184 @@
...
+class A(Element):
+    tag = 'a'
+    def __init__(self, link="",content=""):
+        self.link = link
+        self.content = content
+
+    def render(self, out_file, cur_ind = ""):
+        out_file.write(cur_ind)
+        out_file.write("<{} href={}>".format(self.tag,self.link))
+        out_file.write("{}</{}>\n".format(self.content,self.tag))
```
natasha-aleksandrova: this shouldn't have it's own definition of render() method

students/philip_behrend/lesson07/html_render.py
pbehrend: @natasha-aleksandrova 
Hi Natasha, the render function varies for each tag (for example, the HTML render has a DOCTYPE at the top while Head has a charset argument). How can these nuances be captured without fully rewriting each render function?

natasha-aleksandrova: @pbehrend great question and this is what makes inheritance tricky and great at the same time 🙂 

for `HTML` class in the instructions:
> Update the Html element class to render the “<!DOCTYPE html>” tag at the head of the page, before the html element.
You can do this by subclassing Element, overriding render(), but then calling the Element render from the new render.

which means your code could look something like this:
```
def render(self, file_out, cur_ind=""):
        file_out.write(cur_ind + "<!DOCTYPE html>\n")
        super().render(file_out, cur_ind=cur_ind)
```


`charset` is an attribute and should just be handled as part of the `Element` attribute handling.

PR #489, opened 2019-10-13
Completed HTML Rendering Assignement

alexanderboone96:
This one challenged me the most so far. The learning curve was more difficult to understand how things were working, but this was a great (and challenging) way to learn OOP. I'm excited to learn more.


students/alexander_boone/lesson07/html_render.py
```diff
@@ -0,0 +1,155 @@
...
+    indent = "    "
+
+    def __init__(self, content=None, **kwargs):
+        self.attributes = dict()
+        if content is None:
+            self.contents = []
+        else:
+            self.contents = [content]
+        for key, value in kwargs.items():
+            self.attributes[key] = value
```
natasha-aleksandrova: why not just assign kwargs to self.attributes? are you attempting a deep copy? if so, could you use a comprehension here? you can also use copy.deepcopy which makes the intent more clear

students/alexander_boone/lesson07/html_render.py
```diff
@@ -0,0 +1,155 @@
...
+
+    def append(self, new_content):
+        self.contents.append(new_content)
+
+    def render(self, out_file, cur_ind=""):
+        out_file.write(cur_ind)
+        out_file.write("<{}".format(self.tag))
+        if len(self.attributes.items()) > 0:
```
natasha-aleksandrova: this check is not needed, when you loop over a dictionary and it's empty it just wont' do anything which is the same effect

students/alexander_boone/lesson07/html_render.py
```diff
@@ -0,0 +1,155 @@
...
+class SelfClosingTag(Element):
+
+    def __init__(self, content=None, **kwargs):
+        self.attributes = dict()
+        if content is not None:
+            print("A SelfClosingTag cannot be initialized with contents.")
+            raise TypeError
+        for key, value in kwargs.items():
+            self.attributes[key] = value
+        pass
```
natasha-aleksandrova: call super init method in here to avoid repeating code

students/alexander_boone/lesson07/html_render.py
```diff
@@ -0,0 +1,155 @@
...
+    tag = "meta"
+
+
+class A(OneLineTag):
+    tag = "a"
+
+    def __init__(self, link, content):
+        self.attributes = dict()
+        self.attributes['href'] = link
+        self.contents = [content]
```
natasha-aleksandrova: same here, call super init method

students/alexander_boone/lesson07/html_render.py
```diff
@@ -0,0 +1,155 @@
...
+    tag = "ul"
+
+    def __init__(self, **kwargs):
+        self.attributes = dict()
+        self.contents = []
+        for key, value in kwargs.items():
+            self.attributes[key] = value
+
+    def append(self, content):
+        self.contents.append(content)
```
natasha-aleksandrova: no need to define append or init methods in this class - they can inherit from parent.

PR #490, opened 2019-10-15
submitting lesson04

Reem0330:
thank you for your review

### DIFF
--- a/students/ReemAlqaysi/Lesson04/mailroom.py
+++ b/students/ReemAlqaysi/Lesson04/mailroom.py
@@ -1,0 +1,110 @@
+"""
+MAIL ROOM PART 1
+this script has a data structure that holds a list of  donors and a history of the amounts they have donated. 
+This structure should be populated at first with at least five donors, with between 1 and 3 donations each.
+ data structure stored in the global namespace.
+The script prompts the user (you) to choose from a menu of 3 actions: 
+“Send a Thank You”, “Create a Report” or “quit”.
+"""
+
+from operator import itemgetter
+import sys  # imports go at the top of the file
+
+#initial donations ammount and givers names.
+donor_list = {
+    "Jan Balard": (600.00,250.00),
+    "Joe McHennry": (1500.00,1500.00),
+    "Jeff Hansen": (450.00,150.00),
+    "Scott Newman": (100.00,5000.00),
+    "Rabi Das": (500.00,950.00)
+    }
+# main menue prompt
+def main_menu():
+    print("\n".join(("Welcome to the MailRoom!",
+          "Please Choose an action:",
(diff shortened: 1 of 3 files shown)


students/ReemAlqaysi/Lesson04/mailroom.py
```diff
@@ -0,0 +1,110 @@
...
+from operator import itemgetter
+import sys  # imports go at the top of the file
+
+#initial donations ammount and givers names.
+donor_list = {
+    "Jan Balard": (600.00,250.00),
+    "Joe McHennry": (1500.00,1500.00),
+    "Jeff Hansen": (450.00,150.00),
+    "Scott Newman": (100.00,5000.00),
+    "Rabi Das": (500.00,950.00)
```
natasha-aleksandrova: this should not be a tuple since tuple are immutable and you do want to change this to add more donations.

students/ReemAlqaysi/Lesson04/mailroom.py
```diff
@@ -0,0 +1,110 @@
...
+#get the name and check if it exist, add the ammount, otherwise add the name and the amount to the donors list
+def add_name():
+    fullname = input("please enter full name : ")
+    if fullname == 'list':
+        for key in donor_list:
+            print(key)
+        add_name()
+    else:
+        amount = get_amount()
+        for doner, value in donor_list.items():
```
natasha-aleksandrova: here is what's cool about dict - you don't have to loop through the structure to find the key 🙂 

`if donor_list.get(donor):` will check if donor exists or not

students/ReemAlqaysi/Lesson04/mailroom.py
```diff
@@ -0,0 +1,110 @@
...
+
+def exit_program():
+    print("Bye!")
+    sys.exit()  # exit the interactive script
+
+#send letterto all givers
+def letter_to_all():
+        for k,v in donor_list.items():
+            amount = v[len(v) - 1]
+            fileName = k.replace(' ', '_').replace(',', '') + ".txt"
```
natasha-aleksandrova: make sure to use snake_case naming -> `file_name`

students/ReemAlqaysi/Lesson04/mailroom.py
```diff
@@ -0,0 +1,110 @@
...
+        if response == "1":
+            add_name()
+        elif response == "2":
+            create_report()
+        elif response == "3":
+            letter_to_all()
+        elif response == "4":
+            exit_program()
+        else:
+   	        print('That is not a valid answer!')
```
natasha-aleksandrova: this needs to use switch dict style (course materials go over this)

PR #495, opened 2019-10-17
added slicing.py

Dbaylor1222:


students/david_baylor/lesson3/slicing.py
```diff
@@ -0,0 +1,48 @@
...
+    seq[:1], seq[-1:] = last, first
+    return seq
+
+def remove_every_other(seq):
+    """remove every other item in the sequence"""
+    new_list = []
+    for i in range(len(seq)):
+        if not i % 2:
+            new_list.append(seq[i])
+    return new_list
```
natasha-aleksandrova: this function should use slice syntax

students/david_baylor/lesson3/slicing.py
```diff
@@ -0,0 +1,48 @@
...
+
+def remove_first_last_4(seq):
+    """remove the first and last 4 items in the lsit"""
+    del seq[:4]
+    del seq[-4:]
+    return seq
```
natasha-aleksandrova: you're kind of using slicing here

but the expected here is this `return seq[4:-4:2]`

PR #497, opened 2019-10-20
lesson 3, assignment 1

jlmts:
completed mailroom part 1

### DIFF
--- a/students/jlmts/lesson03/mailroom.py
+++ b/students/jlmts/lesson03/mailroom.py
@@ -1,0 +1,121 @@
+#!/usr/bin/env python
+import sys
+
+"""
+assignment 1: mailroom 
+joli umetsu
+python210
+"""
+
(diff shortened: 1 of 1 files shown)


students/jlmts/lesson03/mailroom.py
```diff
@@ -0,0 +1,121 @@
...
+    if amount.lower() in quit_request:
+        return 
+        
+    # if the name is already on the list, add donation amount to history    
+    if name in donors:
+        for i,j in donor_list:
+            if i == name:
+                j.append(float(amount))
+                email(name)
+            else: pass
```
charlesrobison: Note this is adding a new donor each time a new donation is created.  If I add donor Billy with a donation of $50, then do so again, the report shows 2 different donors named Billy with $50 donations.  It should show a single donor with total donations of $100.  See if you can address this and let us know if you need help.  You'll definitely want this fixed for future versions of this.

PR #499, opened 2019-10-21
submiting lesson 04 v2

Reem0330:
this is the version 2 of the mailroom , I change them according the previous pull request recommendations  , let me know if there is any thing else needed to be change .. thank you!

### DIFF
--- a/students/ReemAlqaysi/Lesson04/mailroom.py
+++ b/students/ReemAlqaysi/Lesson04/mailroom.py
@@ -12,11 +12,11 @@ import sys  # imports go at the top of the file
 
 #initial donations ammount and givers names.
 donor_list = {
-    "Jan Balard": (600.00,250.00),
-    "Joe McHennry": (1500.00,1500.00),
-    "Jeff Hansen": (450.00,150.00),
-    "Scott Newman": (100.00,5000.00),
-    "Rabi Das": (500.00,950.00)
+    "Jan Balard": [600.00,250.00],
+    "Joe McHennry": [1500.00,1500.00],
+    "Jeff Hansen": [450.00,150.00],
+    "Scott Newman": [100.00,5000.00],
+    "Rabi Das": [500.00,950.00]
     }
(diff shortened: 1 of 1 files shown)


students/ReemAlqaysi/Lesson04/mailroom.py
```diff
@@ -75,36 +72,35 @@ def exit_program():
...
     sys.exit()  # exit the interactive script
 
-#send letterto all givers
+#send letter to all givers, file name is snake style
 def letter_to_all():
         for k,v in donor_list.items():
-            amount = v[len(v) - 1]
+            amount = k[len(v) - 1]
             fileName = k.replace(' ', '_').replace(',', '') + ".txt"
+            fileName = fileName.lower()
```
natasha-aleksandrova: my last comment about fileName needing to be snake_style is in regards to your variable naming.

fileName -> file_name


per pep8 (python style guide), all variable and function names follow lower case snake_style naming convention.

Reem0330: currently the result I'm getting for example Reem Alqaysi >> reem_alqaysi is the file name .. not sure what else missing here.. can you advice please if I'm missing something?

natasha-aleksandrova: @Reem0330 per my previous comment we are not talking about name of your file.

We are talking about the variable name in your code. Pep8 is the style guide which advises code style to standardize code across developers. https://www.python.org/dev/peps/pep-0008/


your line of code should read:
` file_name = k.replace(' ', '_').replace(',', '') + ".txt"`

instead of 
`fileName = k.replace(' ', '_').replace(',', '') + ".txt"`

all variable names follow snake_case convention, for example:
`donorsList` vs `donors_list`

Reem0330:
Oh, got it now, sorry miss understood your comment, will fix the variables in my next assignment.. thank you

PR #504, opened 2019-10-25
Added mailroom file for lesson05 assignment.

firstzion:
I did not add any comprehensions to the mailroom.py file. I couldn't find a place where they made sense, but it is totally possible I just don't understand them.

### DIFF
--- a/students/chrissp/lesson05/mailroom.py
+++ b/students/chrissp/lesson05/mailroom.py
@@ -1,0 +1,134 @@
+import sys
+
+# Global Variables
+donor_db = [{"name": "William Gates, III", "donations": [653772.32, 12.17]},
+            {"name": "Jeff Bezos", "donations": [877.33]},
+            {"name": "Paul Allen", "donations": [663.23, 43.87, 1.32]},
+            {"name": "Mark Zuckerberg", "donations": [1663.23, 4300.87, 10432.0]},
+            ]
+
+prompt = "\n".join(("Welcome to the mailroom!",
+                    "Please choose from below options:",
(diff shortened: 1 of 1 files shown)


students/chrissp/lesson05/mailroom.py
```diff
@@ -0,0 +1,134 @@
...
+
+
+def main():
+    nav_dict = {1: send_thanks,
+                2: create_report,
+                3: send_thanks_all,
+                4: quit_program}
+    while True:
+        response = int(input(prompt))
+        if not response in nav_dict:
```
charlesrobison: Exception would be better here to catch invalid input, list strings

students/chrissp/lesson05/mailroom.py
```diff
@@ -0,0 +1,134 @@
...
+            main()
+        else:
+            try:
+                this_donation = float(this_donation)
+            except ValueError:
+                print("Donation must be a number.")
+                send_thanks()
+            else:
+                for person in donor_db:
+                    if user_input.lower() == person["name"].lower():
```
charlesrobison: Nice

students/chrissp/lesson05/mailroom.py
```diff
@@ -0,0 +1,134 @@
...
+            print(person["name"])
+        send_thanks()
+    else:
+        this_donation = input("What is the donation amount?")
+        if this_donation.title() == "Quit":
+            main()
+        else:
+            try:
+                this_donation = float(this_donation)
+            except ValueError:
```
charlesrobison: Good!

PR #519, opened 2019-10-29
Added list_lab.py

Dbaylor1222:


students/david_baylor/lesson3/list_lab.py
```diff
@@ -0,0 +1,79 @@
...
+Demonstrates basic list manipulation.
+"""
+
+def series1(fruit):
+    """series 1"""
+    print("Series 1")
+    print(fruit)
+    fruit.append(input("Name another fruit: "))
+    print(fruit)
+    step = int(input("Pick a number between 1 and {}: ".format(len(fruit))))
```
charlesrobison: Nice use of len()!

students/david_baylor/lesson3/list_lab.py
```diff
@@ -0,0 +1,79 @@
...
+    print(step, ": ", fruit[step - 1])
+    fruit = [input("name another fruit: "),] + fruit
+    print(fruit)
+    fruit.insert(0, input("Name another fruit: "))
+    print(fruit)
+    for i in range(len(fruit)):
+        if fruit[i][0] == "P":
```
charlesrobison: Use of upper() is a good idea here to catch case sensitive user input.

students/david_baylor/lesson3/list_lab.py
```diff
@@ -0,0 +1,79 @@
...
+            answer = input("Do you like {}? ".format(item.lower()))
+            if answer.lower() == "yes":
+                break
+            elif answer.lower() == "no":
+                del_list.append(item)
+                break
+    for i, item in enumerate(fruit):
+        if item in del_list:
+            del(fruit[i])
+    print(fruit)
```
charlesrobison: Works well

PR #521, opened 2019-10-30
Lesson 07 (Assignment 6 HTML Renderer)

zmeves:


students/zach_meves/lesson07/html_render.py
```diff
@@ -0,0 +1,165 @@
+#!/usr/bin/env python3
+
+"""
+A class-based system for rendering html.
+"""
+
+from typing import Union
```
charlesrobison: Union imported but not being used

PR #534, opened 2019-11-09
Lesson_03

amax1213:
Lesson 3 exercises and assignment


amax1213:
list_lab, slicing_lab, and strformat exercises. mailroom assignment

PR #535, opened 2019-11-10
Completed Lesson08

zmeves:
Lesson 8 (Assignment 7 - Circle) completed


students/zach_meves/lesson08/test_circle.py
```diff
@@ -0,0 +1,359 @@
+"""
+Test circle.py
+"""
+
+import unittest
+import os
+from circle import *
```
natasha-aleksandrova: it's typically bad practice to use star imports, import the whole module: `import circle` or specific classes.

PR #537, opened 2019-11-11
Assignment 6

conwaystern:
Assignment 6
Are there additional instructional materials to cover TDD?
I understand the concepts, but I don't feel I can effectively put it into practice.


students/rick_halsell/lesson06/test_mailroom_part4.py
```diff
@@ -0,0 +1,152 @@
+import pytest
+import os
+import ds.mailroom_part4 as mr4
```
charlesrobison: Did you mean to import this?

students/rick_halsell/lesson06/mailroom_part4.py
```diff
@@ -0,0 +1,307 @@
+#!/usr/bin/env python3
+import os
+import sys
+import operator
```
charlesrobison: Both of these operator modules are imported but not used?

natasha-aleksandrova:
@conwaystern can you explain in a little more details specific topics on TDD? I'll be happy to find and share some useful materials on that topic.

PR #538, opened 2019-11-12
add slicing lab

srmorehouse:


students/srmorehouse/Lesson03/slicing_lab.py
```diff
@@ -0,0 +1,71 @@
...
+"""
+def remove_first_last_four ( seq ):
+    return seq [4:-4:2]
+
+"""
+Write some functions that take a sequence as an argument, and return a copy of that sequence:
+with the elements reversed (just with slicing).
+"""
+def reverse_sequence ( seq ):
+    return seq [::-1]
```
natasha-aleksandrova: note that there is no space between your variable and slice syntax -> `return seq[::-1]`

PR #542, opened 2019-11-14
Lesson 03 Exercises and Assignment 

amax1213:


amax1213:
Recommit of Lesson 03 assignment and exercises

PR #546, opened 2019-11-16
lesson 5 - assignment 4

joli-u:
completed mailroom part 3

### DIFF
--- a/students/joli-u/lesson06/test_mailroom.py
+++ b/students/joli-u/lesson06/test_mailroom.py
@@ -1,0 +1,85 @@
+#!/usr/bin/env python
+
+import mailroom4 as mail
+import copy 
+import os 
+
+"""
+test code for mailroom part 4
+lesson 6
+joli umetsu
+py210
+"""
+
+# dictionary of donors and donation amounts that will be used in some tests 
+dict = {
+    'Name1' : [5.10,10.10],
+    'Name2' : [100.01,100.01],
+    }
+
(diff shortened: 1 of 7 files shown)


students/joli-u/lesson06/test_mailroom.py
```diff
@@ -0,0 +1,85 @@
...
+
+"""
+test code for mailroom part 4
+lesson 6
+joli umetsu
+py210
+"""
+
+# dictionary of donors and donation amounts that will be used in some tests 
+dict = {
```
natasha-aleksandrova: this is overwriting a built in `dict`

students/joli-u/lesson06/test_mailroom.py
```diff
@@ -0,0 +1,85 @@
...
+    name = 'Batman'
+    amount = 20.05
+    brdr = "-"*55
+    body = "\n".join(("Dear Batman,\n", 
+        "\tThank you for your generous donation of $20.05.\n", 
+        "\tIt will be put to very good use.\n",
+        "\t\tSincerely,",
+        "\t\t  -The Team"))
+    expected_text = print("\n".join((brdr,body,brdr))) 
+    assert mail.email(name,amount) == expected_text
```
natasha-aleksandrova: this doesn't really make sense for assertion, your mail() function returns `None` so does your `expected_text`

print() does not return a value (`None`)

PR #550, opened 2019-11-17
submiting Lesson07

Reem0330:
is there a video for Tutorial for the Html Render Assignment?, I'm still very confused with the subject.. the Tutorial helped a lot though.

### DIFF
--- a/students/ReemAlqaysi/Lesson07/html_render.py
+++ b/students/ReemAlqaysi/Lesson07/html_render.py
@@ -1,0 +1,117 @@
+#!/usr/bin/env python3
+
+"""
+A class-based system for rendering html. 
+"""
+
+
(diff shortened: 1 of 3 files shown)


students/ReemAlqaysi/Lesson07/html_render.py
```diff
@@ -0,0 +1,117 @@
...
+    def __init__(self, content=None, **kwargs):
+        self.kw_dict = kwargs
+
+        if content is None:
+            self.contents = []
+        elif isinstance(content, list) == False:
+            self.contents = [content]
+        else:
+            self.contents = content
+        pass
```
natasha-aleksandrova: no need for `pass` statements throughout

PR #552, opened 2019-11-17
Assignment 7 Attempt

conwaystern:
I could not follow this assignment even with the walk through.

An accompanying video for the tutorial would have be of immense help. 

Or, starting with a "crawl" through classes, then overlaying TDD, then moving to the large in-depth assignment. I read and tried the tutorial twice, went to other resources to learn about classes and TDD. The Lesson 7 assignment seemed too much for an Intro to OOP. I understand that the next course in this series contains assignment similar to this one, but we need a foundation built first before moving into a lesson 7 assignment.

### DIFF
--- a/students/rick_halsell/lesson06/mailroom_part4.py
+++ b/students/rick_halsell/lesson06/mailroom_part4.py
@@ -1,14 +1,13 @@
 #!/usr/bin/env python3
 import os
 import sys
-import operator
 import time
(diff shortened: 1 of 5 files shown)


natasha-aleksandrova:
@conwaystern appreciate your feedback!

we do agree and we have planned work for next revision to re-order the assignments so that the first one is much simpler.

PR #563, opened 2019-11-25
added string formating lab

Dbaylor1222:


students/david_baylor/lesson3/strformat_lab.py
```diff
@@ -0,0 +1,45 @@
...
+uses python 3
+
+"""
+
+def formatter(in_tuple):
+    form_string = "the {} numbers are: {}".format(len(in_tuple), "{}"+", {}"*(len(in_tuple)-1))
+
+    return form_string.format(*in_tuple)
+
+def tableMaker(row):
```
charlesrobison: Function names (as well as almost everything else) should be snake_case in Python: table_maker(row):

PR #564, opened 2019-11-26
updated mailroom2 and trigrams

fortucj:


students/fortucj/lesson04/mailroom_part2.py
```diff
@@ -20,11 +20,10 @@
 import sys
 
 # This was originally a list of lists.  I changed it based on the recommendation in the notes.
-data = dict([('Bob', [17.56]), ('Billy', [500.00, 1000.00]),
-            ('Joe Schmoe', [2.00, 0.03, 45.00]), ('This Guy', [1.00, 100000]),
-            ('That Gal', [9876.54])])
+Data = {'Bob': [17.56], 'Billy': [500.00, 1000.00], 'Joe Schmoe': [2.00, 0.03, 45.00],
+        'This Guy': [1.00, 100000], 'That Gal': [9876.54]}
```
natasha-aleksandrova: why did you change this to be upper case? all variable names and function names are lower case snake_case

fortucj: I mistakenly thought global variables were capitalized.  Thanks for straightening me out!

natasha-aleksandrova: Ah! 

quick recap:
* `data = ` && `def a_function` - will be lower case for var names and function names
* `MENU_CONSTANT` - all caps snake_case for constants (data _is_ global but not a constant)
* `class DonorCollection` - class names are capitalized camel case 
* `_` unused variable
* `_var_name` - private var name
* `__var_name` - protected var name

PR #566, opened 2019-11-27
Lesson 03: Mailroom Assignment, added sort function

stevetwalker:
Added sort function. I'd love to request an update to my grade for this assignment!

Thanks for letting me know I missed it and I apologize for not seeing that requirement the first time.

### DIFF
--- a/students/steve_walker/lesson03/mailroom.py
+++ b/students/steve_walker/lesson03/mailroom.py
@@ -55,6 +55,8 @@ def create_report():
     for record in donor_records:
         donor_summary.append([record[0], sum(record[1],0), 
             len(record[1]), sum(record[1],0)/len(record[1])])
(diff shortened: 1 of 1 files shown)


students/steve_walker/lesson03/mailroom.py
```diff
@@ -55,6 +55,8 @@ def create_report():
     for record in donor_records:
         donor_summary.append([record[0], sum(record[1],0), 
             len(record[1]), sum(record[1],0)/len(record[1])])
+    donor_summary.sort(key = lambda donor_summary: donor_summary[1], 
+        reverse = True)
```
natasha-aleksandrova: 👍 

couple other options you can use:
* define a sort function and pass for key= param
* use operator.itemgetter function


documentation:
https://docs.python.org/3/howto/sorting.html

natasha-aleksandrova:
grade adjusted!

PR #570, opened 2019-11-29
Lesson 04: Trigrams Assignment, v1

stevetwalker:
I wasn't fully clear what the assignment was asking for, but built a script that can print an ad hoc story using trigrams logic based on Dr. Seuss's Fox in Socks. Please let me know if there's anything specific I'm missing!


stevetwalker:
Also, I'd love to know if there's a more efficient way to handle the punctuation issue while treating select punctuation like independent words!

students/steve_walker/lesson04/trigrams.py
```diff
@@ -0,0 +1,76 @@
...
+# please let me know!
+
+
+import random
+
+def get_file(filename):
+# open file, then add space before and after all punctuation as mod_file
+    with open(filename,'r') as f:
+        mod_text = f.read()
+    return(mod_text)
```
natasha-aleksandrova: no need for parens with `return` statement, you can also return it in place:
```
with open(filename, 'r') as f:
    return f.read()
```

natasha-aleksandrova:
Don't forget to submit on the course website.

PR #571, opened 2019-11-30
Lesson07 through Lesson09

mgummel25:


students/mgummel/lesson07/html_render.py
```diff
@@ -0,0 +1,272 @@
...
+    indent = 3 * ' '
+
+    def __init__(self, content=None, **kwargs):
+        self.content = list()
+        if content:
+            self.content = [content]
+
+        if kwargs:
+            for key,value in kwargs.items():
+                self.attrs += f' {key}=\"{value}\"'
```
natasha-aleksandrova: since its a dictionary, you can safely loop through empty dictionary.

students/mgummel/lesson07/html_render.py
```diff
@@ -0,0 +1,272 @@
...
+A class-based system for rendering html.
+"""
+
+
+class Element(object):
+    """
+    An Element class used to render different html class types. 
+    """
+    tag = "html"
+    attrs = ""
```
natasha-aleksandrova: this shouldn't be a class variable.

Think of class variable as something that never changes with each instance of a class, like it will always be "html" tag. 

However, attrs can be different with each instance.

students/mgummel/lesson08/circle.py
```diff
@@ -0,0 +1,165 @@
...
+    @classmethod
+    def from_diameter(cls, diameter):
+        """
+        Creates a new Circle object based on diameter instead of
+        the radius.
+        """
+        if diameter % 2 == 0:
+            return cls(int(diameter / 2))
+        else:
+            return cls(diameter / 2)
```
natasha-aleksandrova: I don't really understand the need for special logic above ? if modulus returns zero then it's already an integer, no need for special cast to an int

PR #581, opened 2019-12-06
added mail room part one

Dbaylor1222:


students/david_baylor/lesson3/Mail_Room_part1.py
```diff
@@ -0,0 +1,67 @@
...
+    print("{:26}|${:12}|{:11}|${:12}".format(name, total, num_gift, ave_gift))
+
+def write_email(name, donation):
+    print("Here is the email:")
+    print(f"""
+Dear {name.capitalize()},
+
+Thank you for your generous donnation of ${donation}.""")
+
+main()
```
natasha-aleksandrova: you'll want this inside `if __name__ == "__main__":` 


this should be your entry point (and will cause issues in later lessons).

students/david_baylor/lesson3/Mail_Room_part1.py
```diff
@@ -0,0 +1,67 @@
...
+"""
+Mail_room_part1.py
+By David Baylor on 12/3/19
+uses python 3
+
+Automates writing an email to thank people for their donations.
+"""
+
+def main():
+    data = [["bill", 100, 2, 50],["john",75, 3, 25]]
```
natasha-aleksandrova: you can define this inside `if __name__ == "__main__"` block and use it globally without having to pass it into all of the functions.

students/david_baylor/lesson3/Mail_Room_part1.py
```diff
@@ -0,0 +1,67 @@
...
+        data += [[name, donation, 1, donation]]
+        write_email(name, donation)
+        return data
+
+def creat_report(data):
+    print("""
+Donor Name                | Total Given | Num Gifts | Average Gift
+--------------------------|-------------|-----------|-------------""")
+    for i in range(len(data)):
+        make_row(data[i][0].capitalize(), data[i][1], data[i][2], data[i][3])
```
natasha-aleksandrova: this should be sorted (please check assignment requirements)

PR #585, opened 2019-12-07
Lesson 5: mailroom.py

stevetwalker:
Adding mailroom.py assignment. Here, I'm not sure there's anything I need to do to improve the comprehensions. I added additional exception handling, revisited my documentation (added docstrings, improved or removed comments) and ran pycodestyle to ensure I stayed in line with PEP8. If there's anything else I can add, please let me know!


stevetwalker:
Adding lesson 6 mailroom assignments, which include mailroom.py and test_mailroom.py. If there's anything else I can add or clean up, please let me know!

students/steve_walker/lesson06/test_mailroom.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+if __name__ == "__main__":
+    test_store_donation()
+    test_update_record_1()
+    test_update_record_2()
+    test_write_letter()
+    test_thank_all_1()
+    test_thank_all_2()
+    test_create_report()
```
natasha-aleksandrova: you don't want to call these tests manually - did you run the tests using `pytest`? it auto discovers your tests and runs them.

natasha-aleksandrova:
@stevetwalker it's ok if you couldn't find a place for comprehensions - it doesn't always have an application as each student's code is unique 🙂

students/steve_walker/lesson06/mailroom.py
```diff
@@ -0,0 +1,178 @@
...
+
+
+def thank_all():
+    """Thank each donor for their most recent donations."""
+
+    for donor in donor_records:
+        f = donor + ".txt"
+        with open(f, 'w') as f:
+            f.write(write_letter(donor))
+            f.close()
```
natasha-aleksandrova: no need to call `close()` - the `with` statement handles that for you.

stevetwalker:
Thanks! I did use pytests to run all the tests, though I ran it directly
from the command line. Is there a way I can embed pytest into the program
itself?

Thanks for the heads up on with / close(). I remember having trouble with
that in the for loop at first and think I had tried adding close() as a
solution, but never removed it after I figured out the correct solution.
I'll pull it out now!

natasha-aleksandrova:
@stevetwalker unit tests are typically run via pytest as a command line.

There are automated CI tools that are typically used for real projects that can run your tests on each commit. 
Pycharm also has tools to run pytest directly from IDE as an option if you don't want to use a command line.

stevetwalker:
Ok, perfect. Thank you! Also, I often poke around to see other students'
work after I'm done with my first version to see what I might improve, but
if there's ever any specific examples you'd recommend I check out of always
appreciate hearing them!

stevetwalker:
Also, I just wanted to check in about my trigrams assignment from Lesson 4.
It still looks like it hasn't been graded. Is there anything more I need to
do to make this happen?

natasha-aleksandrova:
@stevetwalker based on my comment here: https://github.com/UWPCE-PythonCert-ClassRepos/SP_Online_PY210/pull/570#issuecomment-560034137

it wasn't submitted on the course website at the time of your code review so we couldn't submit the grade. Unfortunately, the course website doesn't have notifications for new submissions, but went ahead and put the grade in so you're should be all set 🙂

stevetwalker:
Perfect. Thank you!

PR #599, opened 2019-12-15
 Completed Lesson04, mailroom.py, assignment

lmburu-uw:
Could you please advice me if, using letter_validator(), is a good approach to programming, especially when using temp files, that don't persist on the filesystem? I was trying to show that the files actually do exist, for debugging reasons. I suspect the end user might not care for the function and it could be considered as unnecessary code, because it's only for development reasons.

### DIFF
--- a/students/larry_mburu/Lesson04/mailroom.py
+++ b/students/larry_mburu/Lesson04/mailroom.py
@@ -1,0 +1,233 @@
+#!/usr/bin/env python3
+
+import sys
+import tempfile
+import os
+from operator import itemgetter
+
+donations_per_individual = { 
+    'William Gates III' : [50000.0, 500.12], 
+    'Mark Zuckerberg' : [20000.0, 8500.99], 
+    'Jeff Bezo': [100000.0, 40000.0, 700.99], 
+    'Paul Allen': [200000.0, 1440.0, 300.00], 
+    'Bill Gates': [30000.0, 70000.0, 450.0]
+}
+
+def display_menu():
+    """
+    Display menu option to user
+    """
+    print("Welcome to Mail Automator (MA)!")
+    print()
+    print("[1] Send a Thank You!")
+    print("[2] Create a report!")
+    print("[3] Send letters to all donors")
+    print("[4] Quit!")
+    print()
+
+def letter_validator(file_tmp_dir):
+    """
+    Function just prints the content of a single letter in file, for 
+    validation only, since the send_thank_you_letter_to_all() utilizes 
+    tempfile.TempDirectory(), which does not persist files in the file system
+
+    param: file_tmp_dir: temporary directory full path name. 
+
+    return: None
+
+    """
+
+    for file in os.listdir(file_tmp_dir): 
+        print(f"[+] Listing contents of file: {file}")
+        with open(os.path.join(file_tmp_dir, file)) as fobj:
+            for line in fobj:
+                if line != '':
+                    print(line)
+        break # just one file. Avoids reading all files.
+
+def send_thank_you_letter_to_all(validate=False): 
+    """
+    Function, sends a thank you letter to all donor's in the donor DB list
+
+    param: validation: If set to True, prints the contents of single donor file, 
+    since tempfile.TempDirectory() does not persist files, in the file system. 
+    It's purely to validate that the letter / formatting was written to file
+    correctly.
+
+    return: None 
+    """
+
+    with tempfile.TemporaryDirectory() as file_tmp_dir: 
+        for donor in donations_per_individual:  
+
+            # -1, the latest donation in the donor "list"
+            email_template = f"""
+            Dear {donor}, 
+                Thank you for your kind donation of ${donations_per_individual[donor][-1]} 
+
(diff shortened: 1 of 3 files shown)


students/larry_mburu/Lesson04/mailroom.py
```diff
@@ -0,0 +1,233 @@
...
+import os
+from operator import itemgetter
+
+donations_per_individual = { 
+    'William Gates III' : [50000.0, 500.12], 
+    'Mark Zuckerberg' : [20000.0, 8500.99], 
+    'Jeff Bezo': [100000.0, 40000.0, 700.99], 
+    'Paul Allen': [200000.0, 1440.0, 300.00], 
+    'Bill Gates': [30000.0, 70000.0, 450.0]
+}
```
natasha-aleksandrova: 👍

students/larry_mburu/Lesson04/mailroom.py
```diff
@@ -0,0 +1,233 @@
...
+    validation only, since the send_thank_you_letter_to_all() utilizes 
+    tempfile.TempDirectory(), which does not persist files in the file system
+
+    param: file_tmp_dir: temporary directory full path name. 
+
+    return: None
+
+    """
+
+    for file in os.listdir(file_tmp_dir):
```
natasha-aleksandrova: avoid using `file` for var name since it's a built in variable (notice different color on github)

students/larry_mburu/Lesson04/mailroom.py
```diff
@@ -0,0 +1,233 @@
...
+
+                it will be put to very good use. 
+
+                Sincerely,
+                    - The Cloud Squad 
+            """
+            with open(file_tmp_dir + os.sep + donor + '.txt', 'w') as fobj: 
+                fobj.write(email_template)
+
+        if validate == True:
```
natasha-aleksandrova: with objects you want to use `is` instead of `==` 

you would use `is` with True/False/None and objects (later in lessons), otherwise for primitive types use `==`

students/larry_mburu/Lesson04/mailroom.py
```diff
@@ -0,0 +1,233 @@
...
+    }
+
+    if response == 1: 
+        dispatch_dict.get(response)()
+    if response == 2: 
+        dispatch_dict.get(response)()
+    if response == 3: 
+        dispatch_dict.get(response)(validate=True)
+    if response == 4: 
+        dispatch_dict.get(response)()
```
natasha-aleksandrova: your above code needs a bit of refactor - it doesn't satisfy assignment requirements on how to implement switch dict but in general has some style issues.


here is an example
```
dispatch_dict = { 
        1 : prompt_for_donation_amount, 
        2 : create_report, 
        3 : send_thank_you_letter_to_all,
        4 : quit_program
    }

while True:
    response = input("...")
    if response not in dispatch_dict:
        print('invalid')
    dispatch_dict[response]()
```

let me know if you have any questions on the above code or why it's written this way.

students/larry_mburu/Lesson04/mailroom.py
```diff
@@ -0,0 +1,233 @@
...
+def send_thank_you_email(name, amount): 
+    """
+    composes an email to a selected donor on 
+    the donor's list 
+    """
+    print(f"Thank you {name} for your generous donation of $ {amount:.2f} dollars")
+    print()
+
+    #return to the original prompt 
+    main()
```
natasha-aleksandrova: no need to manually call the main loop - this should not be controlled via your feature functions.

students/larry_mburu/Lesson04/mailroom.py
```diff
@@ -0,0 +1,233 @@
...
+    return: if donor is in the DB list or not, function returns the donor name 
+    for further use
+    """
+
+    donor = prompt_for_donor_name()
+    donor_list = []
+
+    #extract donor names from main donation DB list for in checks. 
+    for key in donations_per_individual: 
+        donor_list.append(key)
```
natasha-aleksandrova: this is now how you should be adding or checking when using a dictionary.

what's great about a dictionary is that you can easily find keys, or look up a key without any for loops

`list(donations_per_individual)` will return a list of names

to check if donor name already exists -> `if donations_per_individual.get(donor)`

students/larry_mburu/Lesson04/mailroom.py
```diff
@@ -0,0 +1,233 @@
...
+    and proceeds to send a thank you to the donor
+    """
+
+    donor_name = add_donar_to_list()
+    response = input(f"How much would {donor_name} like to donate? ")
+    donation_amount = float(response) 
+
+    #add donation to donor history, in donation DB list
+    if donor_name in donations_per_individual: 
+        donations_per_individual[donor_name].append(donation_amount)
```
natasha-aleksandrova: same comment here, no need to loop - just find donor item by accessing the key in the dictionary.

lmburu-uw:
Natasha,

Thank you for the great pointers. I believe i should be able to fix most of
it. I'll refactor and re-submit.

PR #600, opened 2019-12-16
submission of lesson07

cjfortu:


students/cjfortu/lesson07/html_render.py
```diff
@@ -0,0 +1,124 @@
...
+"""
+A class-based system for rendering html.
+"""
+
+
+# This is the framework for the base class
+class Element(object):
+    indent = 4
+    tag = 'html'
+    kwargs = {}
```
natasha-aleksandrova: this should not be a class variable  - class variables values don't change between instances but in reality your kwargs would change on each instance created.

students/cjfortu/lesson07/html_render.py
```diff
@@ -0,0 +1,124 @@
...
+        self.contents.append(new_content)
+
+    def render_opening(self, out_file, tag_close, cur_ind=0):
+        out_file.write((cur_ind * ' ' + '<{}').format(self.tag))
+        for key, value in self.kwargs.items():
+            out_file.write(' {}="{}"'.format(key, value))
+        if type(self.contents) == str:
+            out_file.write('>{}<'.format(self.contents))
+        else:
+            pass
```
natasha-aleksandrova: shouldn't have `pass` statements

PR #602, opened 2019-12-17
Lesson 5 assignment 4

Eric74nyc:
Added exception handling, list comps, and used dict() where applicable.


students/eric_grandeo/lesson05/mailroom3.py
```diff
@@ -0,0 +1,171 @@
...
+
+
+#switch between users selections, using a dict
+def menu_selection(prompt, dispatch_dict):
+    while True:
+        #handle KeyError
+        try:
+            response = input(prompt)
+            dispatch_dict[response]()
+        except KeyError:
```
natasha-aleksandrova: note that ideally this should just be around line 14

PR #603, opened 2019-12-17
Add lesson03 assignment for choltzman

asemlef:


students/choltzman/lesson03/mailroom.py
```diff
@@ -0,0 +1,98 @@
...
+        print(intro)
+        inp = input("Select: ")
+        print("")
+        if inp == "1":
+            send_thanks()
+        elif inp == "2":
+            create_report()
+        elif inp == "q":
+            print("Exiting.")
+            exit(0)
```
natasha-aleksandrova: ideally you should put this into a separate function (this will make a difference in the lesson 04)  - for now it's ok.

students/choltzman/lesson03/mailroom.py
```diff
@@ -0,0 +1,98 @@
...
+
+            Thank you for your generous donation of ${:,.2f} dollars!
+            """.format(name, gift))
+            print("\n" + emailstr)
+            input("Press Enter to continue...")
+            break
+
+
+def create_report():
+    """Output a list of donors sorted by total donation amount."""
```
natasha-aleksandrova: your comment here says "sorted" but I'm not seeing any sorting implemented.

PR #604, opened 2019-12-18
mailroom4

srmorehouse:
mailroom4 to include unit tests


students/srmorehouse/Lesson06/test_mailroom4.py
```diff
@@ -0,0 +1,58 @@
...
+    return_val = create_report(test_donor)
+    assert return_val == expected
+
+
+def test_send_letters ():
+    cur_dir = os.getcwd()
+    donor_count = len(donor_db)
+    print_report (donor_db)
+    donor_file_count = len([x for x in os.listdir(cur_dir) if x.endswith('.txt')])
+    assert donor_file_count == donor_count
```
charlesrobison: Getting an assertion error here - but not a big deal.  You may want to test a bit more, add new donors, test again, etc.

PR #606, opened 2019-12-18
edited mailroom part 2

Dbaylor1222:
Is there another way to sort a dictionary without using lambda? In the mail room part one assignment page you said not to use lambda but I could not find a way passed it. 
Thanks for the help -David

### DIFF
--- a/students/david_baylor/lesson4/Mail_Room_part2.py
+++ b/students/david_baylor/lesson4/Mail_Room_part2.py
@@ -8,21 +8,21 @@ uses python 3
 Automates writing an email to thank people for their donations.
 """
 
+import sys
 
 def send_thank_you(data):
     name = input("Enter a full name or type 'list' to vew current list: ")
     if name.upper() == "LIST":
-        for i in range(len(data)):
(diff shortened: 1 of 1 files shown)


natasha-aleksandrova:
https://uwpce-pythoncert.github.io/PythonCertDevel/exercises/mailroom_tutorial.html?highlight=sort#sorting

Please review the tutorial section on sorting

students/david_baylor/lesson4/Mail_Room_part2.py
```diff
@@ -8,41 +8,49 @@
...
-        for i in range(len(data)):
-            print(data[i]["name"].capitalize())
+        for i in data:
+            print(i.capitalize())
         return data
     else:
-        for i in range(len(data)):
-            if name.upper() == data[i]["name"].upper(): 
+        for i in data:
+            if name.upper() == i.upper():
```
natasha-aleksandrova: This should be updated to utilize efficient dictionary lookups (hint: you should use `get()` method on the dictionary here.

PR #607, opened 2019-12-19
Adding lesson_03 exercises.

N0vA4:
Added slicing_lab, list_lab, and strformat_lab files.  ***** I made a mistake and accidentally submitted my file -- strformat.py on the course website with the bug still in it.  It should be fixed here.

### DIFF
--- a/students/N0vA/lesson_03/mailroom.py
+++ b/students/N0vA/lesson_03/mailroom.py
@@ -1,0 +1,106 @@
+#!/usr/bin/env python3
+import sys
+#### Mailroom Project Part 1 ####
+
+# Create list of donors and their donation history
+database = [('Bill Gates', [2000000, 250000000]), ('Jeff Bezos', [2000000]), ('Elon Musk', [50000000, 10000000]), ('Howard Schultz', [1000000]), ('Paul Allen', [450000000])]
+
+# Send a thank you
+def thank_you():
+    
+    donor_list = database
+    print('Alright.  Which donor would you like to send a thank you card?')
+    
+    while True:
+        person = input('Enter their name here or type "list" to see a list of donors: ')
+        past_donor = False
+
+        # Print list option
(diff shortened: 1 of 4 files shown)


students/N0vA/lesson_03/mailroom.py
```diff
@@ -0,0 +1,106 @@
...
+            for name in donor_list:
+                if person.lower() == name[0].lower():
+                    past_donor = True
+
+            # Adding a new donor
+            if past_donor == False:
+                amount_donated = input('How much was their donation? ')
+                amount_donated = int(amount_donated)
+                new_record = (person, [amount_donated])
+                donor_list = donor_list.append(new_record)
```
natasha-aleksandrova: the assignment asks to add or update existing donations. Your code is written to only add donation on new donation.

students/N0vA/lesson_03/mailroom.py
```diff
@@ -0,0 +1,106 @@
...
+                    'We wish you all the best,\n\n'
+                    'Local Charity Persident\n')
+
+            email
+
+            # Exit to main menu
+            exit = 'fill'
+            while exit != 'quit':
+                exit = input('Press any button to return to main menu... ')
+                main()
```
natasha-aleksandrova: you shouldn't call main() from here, just have `return` statement.

students/N0vA/lesson_03/mailroom.py
```diff
@@ -0,0 +1,106 @@
...
+    if task == '1':
+        thank_you()
+    elif task == '2':
+        report()
+    elif task == '3':
+        print('Enjoy the rest of your day!')
+        sys.exit()
+    else:
+        print("Not a valid option!")
+        main()
```
natasha-aleksandrova: you shouldn't have all the various calls to main() - this should be controlled via simple while True loop.

Please review the tutorial.

PR #608, opened 2019-12-20
Lesson 3 exercises and assignments

Freemabe:
This is my lesson three work, there is an issue with the assignment where i used  .sorted and dictionaries before they were introduced. This wasn't intentional, I just didn't see last line of the guidelines page. If I need to rework my script to not use those lemme know. Thanks.

### DIFF
--- a/students/Ben_freeman/lesson03/Slicing_lab.py
+++ b/students/Ben_freeman/lesson03/Slicing_lab.py
@@ -1,0 +1,24 @@
+def swap_fir_la(seq):
+    """Swaps the first and last elements of a finite sequence"""
+    fir=seq[0:1]
+    la=seq[len(seq):len(seq)-2:-1]
+    mid=seq[1:len(seq)-1]
+    return la+mid+fir
+
+def remove_mod2(seq):
+    """removes every other item"""
+    return seq[::2]
+
+def fourfour_mod2(seq):
+    """removes first and last 4 elements and then every other"""
+    return seq[4:-4:2]
+
+def uno_reverse(seq):
+    """reverses the sequence"""
+    return seq[::-1]
+
+def third(seq):
+    f=seq[0:len(seq)//3]
+    m=seq[len(seq)//3:2*len(seq)//3]
+    l=seq[2*len(seq)//3:len(seq)]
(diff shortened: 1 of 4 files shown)


students/Ben_freeman/lesson03/Slicing_lab.py
```diff
@@ -0,0 +1,24 @@
+def swap_fir_la(seq):
+    """Swaps the first and last elements of a finite sequence"""
+    fir=seq[0:1]
+    la=seq[len(seq):len(seq)-2:-1]
+    mid=seq[1:len(seq)-1]
+    return la+mid+fir
```
natasha-aleksandrova: note that this can be just one line: `return seq[-1:] + seq[1:-1] + seq[:1]`

students/Ben_freeman/lesson03/mailroom.py
```diff
@@ -0,0 +1,68 @@
...
+donorlist={
+"Ahmed":[123,456,789],
+"Sally": [10,11],
+"Bob": [12],
+"Vladamir": [13.14,1516.17],
+"Santa": [181920.21,22,23.24],
+}
+
+#send thanks funct
+def sndthx():
```
natasha-aleksandrova: it is preferred to spell out function names: `def send_thank_you`

students/Ben_freeman/lesson03/mailroom.py
```diff
@@ -0,0 +1,68 @@
...
+    while True:
+        qry=input("Please enter a Full Name or enter 'list' to see a list of current donors: ")
+        if qry in donorlist:
+            Newdonate=float(input("How much would you like to donate: \n"))
```
natasha-aleksandrova: variable and function names follow snake_case convention -> 'new_donation`

Freemabe:
Thank you for the feedback regarding lesson 3 and variable names!

students/Ben_freeman/lesson03/Slicing_lab.py
Freemabe: Oh that is much cleaner, nice! Thanks for the feedback.

PR #611, opened 2019-12-22
Fixed bugs and added section for past donors new donations

N0vA4:
I had some issues with global vs local variable and appending it which is why I create copies for local use.  When I would append a new donor onto my existing list it would change the local variable and erase everything from the global.  I couldn't figure out why this was happening.

*Also do I need to resubmit the file on the course website or is the pull request ok?

### DIFF
--- a/students/N0vA/lesson_03/mailroom.py
+++ b/students/N0vA/lesson_03/mailroom.py
@@ -1,9 +1,14 @@
 #!/usr/bin/env python3
-import sys
 #### Mailroom Project Part 1 ####
 
+import sys
+
 # Create list of donors and their donation history
-database = [('Bill Gates', [2000000, 250000000]), ('Jeff Bezos', [2000000]), ('Elon Musk', [50000000, 10000000]), ('Howard Schultz', [1000000]), ('Paul Allen', [450000000])]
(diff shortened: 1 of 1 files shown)


natasha-aleksandrova:
@N0vA4 you couldn't use `database` as a global variable because there are couple issues in your code.

`donor_list = donor_list.append(new_record)` I'm assuming you were assigning this to `database` variable? 

1. append() method returns None so you would not assign its output to anything
2. you don't want to re-write your `database` variable under any conditions in your code.


Update your code instances where you overwriting database variable and then you should have no problems using it directly.

natasha-aleksandrova:
Correct, no need to resubmit on the course website.

students/N0vA/lesson_03/mailroom.py
```diff
@@ -25,12 +30,15 @@ def thank_you():
             for name in donor_list:
                 if person.lower() == name[0].lower():
                     past_donor = True
+                    amount_donated = input('How much was their donation? ')
+                    amount_donated = int(amount_donated)
+                    name[1].append(amount_donated)
```
natasha-aleksandrova: this would be a great place for a `break` statement

PR #612, opened 2019-12-22
Updated version of mailroom.py that uses lists.

Freemabe:
This version uses lists and should have all of the wonky variable names changed.

### DIFF
--- a/students/Ben_freeman/lesson03/mailroom.py
+++ b/students/Ben_freeman/lesson03/mailroom.py
@@ -1,52 +1,51 @@
 #Donor dict
-donorlist={
-"Ahmed":[123,456,789],
-"Sally": [10,11],
-"Bob": [12],
-"Vladamir": [13.14,1516.17],
-"Santa": [181920.21,22,23.24],
-}
+donor_list=[["Ahmed",123,456,789],
+["Sally", 10,11],
+["Bob", 12],
+["Vladamir", 13.14,1516.17],
+["Santa", 181920.21,22,23.24],
+]
 
 #send thanks funct
-def sndthx():
+def send_thanks():
     while True:
-        qry=input("Please enter a Full Name or enter 'list' to see a list of current donors: ")
-        if qry in donorlist:
-            Newdonate=float(input("How much would you like to donate: \n"))
-            donorlist[qry].append(Newdonate)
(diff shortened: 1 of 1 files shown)


natasha-aleksandrova:
@Freemabe when testing your code there is a bug:
* add a new donor with donation 100
* add a donation again for the same donor 200
* this donor is now listed twice in the list and will not update as part of existing donor

students/Ben_freeman/lesson03/mailroom.py
```diff
@@ -1,52 +1,51 @@
...
-        mrgl[i]=[sum(dlist[i]),len(dlist[i]),sum(dlist[i])/len(dlist[i])]
-    flist=[]
-    for keys, values in mrgl.items():
-        values.append(keys)
-        flist.append(values)
-    return sorted(flist, key=lambda flist:flist[0], reverse=True)
+#making a sorted list
+def sorting_function(dlist):
+    sorted_list=[]
+    for i in range(0,len(dlist)):
```
natasha-aleksandrova: you should loop over the list item directly instead of using `range()` - there are uses for this but not very common.

```
for donor in dlist:
    
```

students/Ben_freeman/lesson03/mailroom.py
```diff
@@ -1,52 +1,51 @@
...
+            elif query=="list":
+                    print(donor_list[i][0])
+            elif query.lower()=="quit":
+                return
+            elif query not in donor_list[i][0]:
+                new_donate=float(input("\nHow much would you like to donate: \n"))
+                donor_list.append([str(query),new_donate])
+                return print(f"\nThank you {query} for your donation of ${new_donate}, your donations make the work of the 'American society for taking donations' possible.\n\n"
+                    "Sincearly,\n\n"
+                    "A low paid intern\n")
```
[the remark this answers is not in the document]
Freemabe: Hello! I tried this but ended up running into the same problems as before. After some testing it seems like using for then if/else makes the machine look at each donor individually instead of as a group like I want. 

I.E.
donors=[donor1,donor2]
for donors in donors:
     if donor=response:
          update
     else:
           add new donor

unless response is equal to donor1 this will add a new donor, even if response==donor2.
Let me know if I'm thinking about this correctly.

students/Ben_freeman/lesson03/mailroom.py
Freemabe: Oh, that's much nicer looking and more straight forward, thank you!

PR #617, opened 2019-12-24
trigrams assignment

shervs:


students/shervs/lesson04/trigrams.py
```diff
@@ -0,0 +1,77 @@
...
+        if line[0:3] == '***':
+            in_lines.remove(line)
+        #Remove punctuation
+        else:
+            new_line = line.translate(str.maketrans(
+                    '''0123456789!"#$%&()*+,-./:;<=>?@[\]^_`{|}~'''
+                    ,'                                         ' ))
+            #Break each line to wrods and feed into the word list
+            for word in new_line.split():
+                if word is not "I":
```
natasha-aleksandrova: `is` is used with objects comparison like True/False/None and other objects.

for primitive type comparison use `==`

students/shervs/lesson04/trigrams.py
```diff
@@ -0,0 +1,77 @@
...
+        elif count == word_limit-1:
+            new_list.append(first_pair[0])
+            count += 1
+        elif count == word_limit-2:
+            new_list.append(first_pair[0])
+            new_list.append(first_pair[1])
+            count += 2
+        else:
+            new_list.append(first_pair[0])
+            new_list.append(first_pair[1])
```
natasha-aleksandrova: note that you can use `extend()` here -> `new_list.extend(first_pair)`

PR #620, opened 2019-12-25
Added dict_lab and mailroom 2 files to directory

N0vA4:
I had a question that I'm not sure why I couldn't figure out.  For my mailroom2.py I have a chunk of code that sets up the directory for the filepaths I am going to write.  It looks something like:

for donor in database.keys():
        first_name = donor.split()[0]
        last_name = donor.split()[1]
        file_format = "{0}_{1}.txt".format(first_name, last_name)
        file_path = folder / file_format

My issue was that if for file_format instead I put
 file_format = "letters/{0}_{1}.txt".format(first_name, last_name)**

to create a subfolder in lesson_04 for the letters it would throw and error saying the didn't exist.  Was wondering what I needed to change to fix that?  Thanks for the help.

### DIFF
--- a/students/N0vA/lesson_04/mailroom2.py
+++ b/students/N0vA/lesson_04/mailroom2.py
@@ -1,0 +1,129 @@
+#!/usr/bin/env python3
+#### Mailroom Project Part 2 ####
+
+import sys
+import pathlib
+
+# Create list of donors and their donation history
+database = {'Bill Gates': [2000000, 250000000], 
+            'Jeff Bezos': [2000000],
+            'Elon Musk': [50000000, 10000000], 
+            'Howard Schultz': [1000000], 
+            'Paul Allen': [450000000]}
+
+# Send a thank you
+def thank_you():
+    
+    print('Alright.  Which donor would you like to send a thank you card?')
+    
+    while True:
+        person = input('Enter their name here or type "list" to see a list of donors: ')
+        past_donor = False
+
+        # Print list option
+        if person.lower() == 'list':
+            past_donor = True
+            for per in database.keys():
+                print(per)
+
+        else:
+            for name in database.keys():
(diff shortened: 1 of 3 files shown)


natasha-aleksandrova:
> I had a question that I'm not sure why I couldn't figure out. For my mailroom2.py I have a chunk of code that sets up the directory for the filepaths I am going to write. It looks something like:
> 
> for donor in database.keys():
> first_name = donor.split()[0]
> last_name = donor.split()[1]
> file_format = "{0}_{1}.txt".format(first_name, last_name)
> file_path = folder / file_format
> 
> My issue was that if for file_format instead I put
> file_format = "letters/{0}_{1}.txt".format(first_name, last_name)**
> 
> to create a subfolder in lesson_04 for the letters it would throw and error saying the didn't exist. Was wondering what I needed to change to fix that? Thanks for the help.

@N0vA4 you would need to create the `letters` directory first before using it.

Please take a look at this article that does a great job explaining how to do that and also catch errors:
https://stackabuse.com/creating-and-deleting-directories-with-python/

students/N0vA/lesson_04/mailroom2.py
```diff
@@ -0,0 +1,129 @@
...
+        past_donor = False
+
+        # Print list option
+        if person.lower() == 'list':
+            past_donor = True
+            for per in database.keys():
+                print(per)
+
+        else:
+            for name in database.keys():
```
natasha-aleksandrova: This should not longer be a for loop. Dictionary look ups are very efficient.

Utilize `if database.get(name)` check in your logic. 

I would recommend code similar to this:
```
amount = input('...')
if database.get(name):
    database[name].append(amount)
else:
    database[name] = [amount]
```

as you can see it is much simpler than your current code.
```

PR #621, opened 2019-12-25
lesson03: list_lab, slicing, strformat_lab, mailroom

bromleysj:


students/scott_bromley/lesson03/mailroom.py
```diff
@@ -0,0 +1,142 @@
...
+    """
+    prints a formatted table of donors ranked in descending order by Total Given
+    :return: None
+    """
+    names, totals, num_gifts, avg_gift = get_donor_summary(donors)
+    print(f"Donor Name{'':<20} | Total Given{'':>0} | Num Gifts{'':>0} | Average Gift{'':>0}")
+    print(f"-" * 72)
+    for name, total, num_gift, avg_gift in zip(names, totals, num_gifts, avg_gift):
+        print(f"{name:<32}${total:>11}{num_gift:>12}  ${avg_gift:>13}")
+    return None
```
natasha-aleksandrova: another unneeded return statement.

students/scott_bromley/lesson03/mailroom.py
```diff
@@ -0,0 +1,142 @@
...
+    """
+    :param donors: optional list of donors and donations
+    :return: lists of names, totals, number of gifts, avg gift sorted by highest total
+    """
+    sorted_donors = sorted(donors, key=lambda x: sum(x[1]), reverse=True)
+    names = tuple(map(lambda x: x[0], sorted_donors))
+    totals = tuple(map(lambda x: float(format(sum(x[1]), '.2f')), sorted_donors))
+    num_gifts = tuple(map(lambda x: len(x[1]), sorted_donors))
+    avg_gift = tuple(map(lambda x: float(format(statistics.mean((x[1])), '.2f')), sorted_donors))
+    return names, totals, num_gifts, avg_gift
```
natasha-aleksandrova: above is actually pretty inefficient - there are a total of 4 loops when all of this can be done in one loop and put into a list structure which you create later anyway.

students/scott_bromley/lesson03/mailroom.py
```diff
@@ -0,0 +1,142 @@
...
+
+
+def get_donors(donors=None):
+    """
+    :param donors: optional list of donors
+    :return: tuple of donor names
+    """
+    if donors is None:
+        donors = []
+    return tuple(map(lambda x: x[0], donors))
```
natasha-aleksandrova: there is no other use of this function but to print the global donors list, so I would recommend using it directly.

The use map/lambdas functions is an overkill.

```
def print_donors():
    for donor in donors:
        print(donor[0])
```
this can be done as one liner with join() function but we haven't learned comprehensions yet.

bromleysj:
Natasha,

Thanks so much for your comments, the feedback was very helpful.

I've made improvements to my code in this PR:
https://github.com/UWPCE-PythonCert-ClassRepos/SP_Online_PY210/pull/624

Thanks,

Scott

natasha-aleksandrova:
sounds great!

PR #623, opened 2019-12-25
FizzBuzz-exercise.py

ikempcode33:
FizzBuzz Exercise is ready for grading! Still working on others.

### DIFF
--- a/students/Isabella_Kemp/lesson02/FizzBuzz-exercise.py
+++ b/students/Isabella_Kemp/lesson02/FizzBuzz-exercise.py
@@ -1,0 +1,15 @@
+#Isabella Kemp
+#12/23/2019
+#Fizz Buzz
+
+for x in range (1,101):
+    if x % 3 == 0 and x % 5 == 0:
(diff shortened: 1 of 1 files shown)


students/Isabella_Kemp/lesson02/FizzBuzz-exercise.py
```diff
@@ -0,0 +1,15 @@
+#Isabella Kemp
+#12/23/2019
+#Fizz Buzz
+
+for x in range (1,101):
```
charlesrobison: You'll want to remove the space after range, and add one after the comma.  So: for x in range(1, 101):

students/Isabella_Kemp/lesson02/FizzBuzz-exercise.py
```diff
@@ -0,0 +1,15 @@
...
+    if x % 3 == 0 and x % 5 == 0:
+        print ("Fizz Buzz")
```
charlesrobison: No space after print, as well as below.

PR #626, opened 2019-12-27
Made changes to mailroom2.py based on comments from last PR

N0vA4:
Made changes to mailroom2.py off comments from last pull request.  THank you for the feedback.  Also added txt files that were used in exercises.


students/N0vA/lesson_04/mailroom2.py
```diff
@@ -11,46 +11,37 @@
             'Howard Schultz': [1000000], 
             'Paul Allen': [450000000]}
 
+# List of donors
+donor_list = list(database.keys())
```
natasha-aleksandrova: this introduced a bug - if you add a new donor and then type "list" command, the new donor is not printed.

this should not be in the global space, just add `print(list(database.keys()))` on line 26

PR #628, opened 2019-12-28
mailroom part 2: return of the mailman

Freemabe:
Couple of questions:

1)How would I put a "quit" command into my menu dictionary? I tried putting a function that did
`return print("quitting scripted now") `
type statement, but that did not break out of the while loop I have in my  if __int__==__main__: block. Also tried just putting a break as the value for the quitting option in my dictionary, but that didn't work either. Is there a way to do this cleanly?

2)Does python always try to execute your code while reading it? I have run into minor issues because of where in my code I put a dictionary (for example) and was just wondering if there is a way for me to make sure that when executed my code is read first, and all of the variables initialized then my code runs. Hopefully this question makes sense, I don't quite know what the right words to use for this are. 

Cheers.

### DIFF
--- a/students/Ben_freeman/lesson04/mailroom2.py
+++ b/students/Ben_freeman/lesson04/mailroom2.py
@@ -1,0 +1,88 @@
+#Donor dict
+donor_list={
+"Ahmed":[123,456,789],
+"Sally": [10,11],
+"Bob": [12],
+"Vladamir": [13.14,1516.17],
+"Santa": [181920.21,22,23.24],
+}
+
+
+#send thanks funct
+def send_thanks():
+    while True:
+        query=input("Please enter a Full Name, enter 'list' to see a list of current donors, or quit to exit: ")
+        if query in donor_list:
+            new_donate=float(input("How much would you like to donate: \n"))
+            donor_list[query].append(new_donate)
+            print(f"\nThank you {query} for your donation of ${new_donate}, your donations make the work of the 'American society for taking donations' possible.\n\n"
+                "Sincearly,\n\n"
+                "A low paid intern\n")
+            break
+        elif query=="list":
+            print('\n'.join(donor_list))
+        elif query.lower()=="quit":
+            return
+        elif query not in donor_list:
+            donor_list[query]=[]
+            new_donate=float(input("\nHow much would you like to donate: \n"))
+            donor_list[query].append(new_donate)
+            print(f"\nThank you {query} for your donation of ${new_donate}, your donations make the work of the 'American society for taking donations' possible.\n\n"
+                "Sincearly,\n\n"
+                "A low paid intern\n")
+            break
+        
+
+#making a sortable list
+def sorting_function(dlist):
+    mrgl={}
+    for i in dlist:
+        mrgl[i]=[sum(dlist[i]),len(dlist[i]),sum(dlist[i])/len(dlist[i])]
+    flist=[]
+    for keys, values in mrgl.items():
+        values.append(keys)
+        flist.append(values)
+    return sorted(flist, key=lambda flist:flist[0], reverse=True)
+#creating a report
+def create_report(): 
+    listy=sorting_function(donor_list)
+    title=["Donor Name", "Total Given", "Num Gifts", "Average Gift"]
+    print(f"{title[0]:<20} | {title[1]:>10} | {title[2]:<} | {title[3]:<}")
+    print(f"-------------------------------------------------------------")
+    for i in range(0,len(listy)):
+        print(f"{listy[i][3]:<21} ${listy[i][0]:>12.2f}{listy[i][1]:>11}  ${listy[i][2]:>12.2f}")
+
+#writing thank you to all donors function
+def mass_send_thanks():
+    for donor in donor_list:
+        with open(f"{donor}.txt","w+") as letter:
+            letter.write(f"\nThank you {donor} for your donation of ${sum(donor_list[donor]):.2f}, your donations make the work of the 'American society for taking donations' possible.\n\n"
+                    "Sincearly,\n\n"
+                    "A low paid intern\n")
+    print("\nSending out thank you letters\n")
+
+#Menu dict
+menu={
+"A":send_thanks,
+"B":create_report,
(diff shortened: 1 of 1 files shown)


natasha-aleksandrova:
@Freemabe question 1 was covered in the mailroom tutorial: https://uwpce-pythoncert.github.io/PythonCertDevel/exercises/mailroom_tutorial.html

```def exit_program():
    print("Bye!")
    sys.exit()  # exit the interactive script
```

you would write a function that either calls `sys.exit()` or `quit()` function.


For your question # 2, I don't fully understand - is it possible you could provide some examples of the issues you've run into?
Perhaps you're asking about Python difference to Java? Interpreted vs Compiled language difference?

students/Ben_freeman/lesson04/mailroom2.py
```diff
@@ -0,0 +1,88 @@
...
+        elif query.lower()=="quit":
+            return
+        elif query not in donor_list:
+            donor_list[query]=[]
+            new_donate=float(input("\nHow much would you like to donate: \n"))
+            donor_list[query].append(new_donate)
+            print(f"\nThank you {query} for your donation of ${new_donate}, your donations make the work of the 'American society for taking donations' possible.\n\n"
+                "Sincearly,\n\n"
+                "A low paid intern\n")
+            break
```
natasha-aleksandrova: you can re-organize this a bit to reduce some duplicated code:
```
while True:
    query = input('...')
    if query == "list":
        # print donors
    elif query == "quit":
        return
    else:
        donation = input('...')
        if donor in donor_list:
            # update donation
        else:
            # new donor
```

students/Ben_freeman/lesson04/mailroom2.py
```diff
@@ -0,0 +1,88 @@
...
+#making a sortable list
+def sorting_function(dlist):
+    mrgl={}
+    for i in dlist:
+        mrgl[i]=[sum(dlist[i]),len(dlist[i]),sum(dlist[i])/len(dlist[i])]
+    flist=[]
+    for keys, values in mrgl.items():
+        values.append(keys)
+        flist.append(values)
+    return sorted(flist, key=lambda flist:flist[0], reverse=True)
```
natasha-aleksandrova: I'm having a hard time understanding above function. 

Typically, you would just create a list of tuples, each tuple containing the summary info for the report.

students/Ben_freeman/lesson04/mailroom2.py
```diff
@@ -0,0 +1,88 @@
...
+        prompt=input(
+            "What would you like to do?:\n"
+            "A - Send a thank you\n"
+            "B - Create a report\n"
+            "C - Send a thank you to all donors\n"
+            "D - Quit\n")
+        if prompt.upper() in ["A","B","C"]:
+            menu[prompt.upper()]()
+        elif prompt.upper() == "D":
+            break
```
natasha-aleksandrova: map your option "D" to a quit() or sys.exit() function to avoid special casing.

natasha-aleksandrova:
don't forget to submit assignment on the course website.

Freemabe:
> @Freemabe question 1 was covered in the mailroom tutorial: https://uwpce-pythoncert.github.io/PythonCertDevel/exercises/mailroom_tutorial.html
> 
> ```
> def exit_program():
>     print("Bye!")
>     sys.exit()  # exit the interactive script
> ```
> 
> you would write a function that either calls `sys.exit()` or `quit()` function.

Excellent, thank you!

> For your question # 2, I don't fully understand - is it possible you could provide some examples of the issues you've run into?
> Perhaps you're asking about Python difference to Java? Interpreted vs Compiled language difference?

This
`demo_list=[1,2,3,4]
def demo_function():
	print(demo_list)
demo_function()
`
and this
`
def demo_function():
	print(demo_list)
demo_list=[1,2,3,4]
demo_function()
`
work, but this
`def demo_function():
	print(demo_list)
demo_function()

demo_list=[1,2,3,4]`
does not. 

It makes perfect sense why you cannot call a function before everything in it has been been defined, but I would imagine that when working on some like old code that has had like 20 people working on it, it might be useful to be able to define functions wherever. (This is a convoluted scenario but hopefully you get what I'm saying)

students/Ben_freeman/lesson04/mailroom2.py
Freemabe: This is my hamfisted way of turning my dictionary into a list which is then sorted. 
```
    mrgl={}
    for i in dlist:
        mrgl[i]=[sum(dlist[i]),len(dlist[i]),sum(dlist[i])/len(dlist[i])]
```
collects all of the information that we want for each donor and assigns it to that donor

```    flist=[]
    for keys, values in mrgl.items():
        values.append(keys)
        flist.append(values)
    return sorted(flist, key=lambda flist:flist[0], reverse=True)
```
this takes the temporary dictionary that we just produced and merges the numerical values and donor name together to form a list which we then sort. 

Looking at it again it seem like there is a lot of unnecessary code here and it could probably be shortened to something like 

```
def sorting_function(dictionary):
    temp_list=[]
    for keys, values in dictionary.items():
        temp_list.append([sum(values),keys,len(values),sum(values)/len(values)])
    return sorted(temp_list, reverse=True)
```

Is there a benefit to using a list of tuples over a list of lists or would this be sufficient?

students/Ben_freeman/lesson04/mailroom2.py
Freemabe: Cool cool, thank you, this was bugging me.

students/Ben_freeman/lesson04/mailroom2.py
Freemabe: Nice, I'll get on that!

students/Ben_freeman/lesson04/mailroom2.py
natasha-aleksandrova: your condensed version is exactly what I was suggesting.

The main reason you would use a tuple over a list is if you want to be clear and sure that a list item will not be modified by the consumer of the data. You can use either but a list of tuples is technically more correct.

PR #629, opened 2019-12-28
Lesson 9: Mailroom Assignment

stevetwalker:
First submission of the final mailroom assignment.


students/steve_walker/lesson09/cli_main.py
```diff
@@ -0,0 +1,152 @@
...
+from donor_models import Donor, DonorCollection
+
+# Initial donations
+initial_donations = {"Rhianna": [747, 3030303, 1968950],
+                     "Grumps": [5.99],
+                     "EatPraySlay": [100000, 200000, 300000],
+                     "Muir": [469503, 50000, 186409],
+                     "Spacewalker": [4406, 342]}
+
+donors = DonorCollection.from_dict(initial_donations)
```
natasha-aleksandrova: nice use of alternate constructor here!

students/steve_walker/lesson09/donor_models.py
```diff
@@ -0,0 +1,73 @@
...
+
+    def add(self, donor):
+        """Add donor to the DonorCollection."""
+
+        self.donors.append(donor)
+
+    def list(self):
+        """Create list of donor names."""
+
+        donor_list = [self.donors[i].name for i in range(len(self.donors))]
```
natasha-aleksandrova: using a range() here is completely unnecessary - instead:
```
return [donor.name for donor in self.donors]
```

PR #630, opened 2019-12-28
Lesson04, code changes. Lesson05, completed Exceptions exercise. 

lmburu-uw:
@natasha-aleksandrova  - I adjusted the code for mailroom.py with the following changes. 

1. In letter_validator(), change the for loop not to use 'file' as a var as it's builtin 
2. In send_thank_you_to_all(), changed validate '==' True operator, to validate 'is' True. True/False/None are singleton objects
3. Refactored main function to use a while loop for the prompt and the 'in' dictionary operator, rather than a bunch of if statements
4. Removed main() function from send_thank_email() function. Unnecessary looping back to main prompt. 
5. In add_donar_to_list() function, removed the use of for loops to grab dictionary keys. Refactored code to use the dict.get() method.  
6. In prompt_for_donation_amount() function, removed 'if key in dict, append statements' and used the concise 'dict.get(key).append()' syntax

@charlesrobison - i added the 'sherlock.txt' full text version for use with the Trigram.py app in Lesson04. It looks like a response via email with an attachment, extracts the attachments content into the response window of GitHub, which is horrible. Sorry, i'm trying to get used to the GitHub system.

### DIFF
--- a/students/larry_mburu/Lesson04/mailroom.py
+++ b/students/larry_mburu/Lesson04/mailroom.py
@@ -45,7 +45,7 @@ def letter_validator(file_tmp_dir):
                     print(line)
         break # just one file. Avoids reading all files.
 
-def send_thank_you_letter_to_all(validate=False): 
+def send_thank_you_letter_to_all(validate=True): 
     """
     Function, sends a thank you letter to all donor's in the donor DB list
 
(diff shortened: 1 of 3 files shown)


lmburu-uw:
Charles,

I'll revise the code and see if i can fix the bug. If not, i'll reach out.
Also, are temp files not a requirement? It was part of the lesson guide.
I'll shift the code to use permanent files. Thanks.

PR #634, opened 2019-12-30
submission of lesson 08

cjfortu:


students/cjfortu/lesson08/Circle_Class.py
```diff
@@ -0,0 +1,95 @@
+#!/usr/bin/env python
+
+from math import *
```
natasha-aleksandrova: using "star" imports is bad practice, it is better to import specific modules.

PR #640, opened 2019-12-31
Lesson 6 assignment 5

Eric74nyc:
Created test suite, refactored mailroom4.py


students/eric_grandeo/lesson06/test_mailroom4.py
```diff
@@ -0,0 +1,76 @@
...
+#tests that files are created with proper name
+def test_send_letters():
+    mailroom4.send_letters(donors2)
+    for name in [*donors2]:
+        name2 = name.replace(" ", "_")
+        assert os.path.isfile(name2 + ".txt") is True
+
+#test content of file
+
+def test_get_letter_text():
```
natasha-aleksandrova: each one of your tests should execute your code and then assert on it - this particular test already assumes that the file was created earlier.

PR #641, opened 2020-01-01
Lesson 08 Circle

srmorehouse:
circle class and tests

### DIFF
--- a/students/srmorehouse/Lesson08/circle.py
+++ b/students/srmorehouse/Lesson08/circle.py
@@ -31,3 +31,9 @@ class Circle (object):
     def using_diameter (cir, diameter):
         return cir (diameter / 2)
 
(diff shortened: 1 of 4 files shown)


students/srmorehouse/Lesson08/circle.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+    @diameter.setter
+    def diameter (self, diameter):
+        self.radius = diameter / 2
+
+
+    @classmethod
+    def using_diameter (cir, diameter):
+        return cir (diameter / 2)
```
natasha-aleksandrova: (minor issue) the directions have `from_diameter` for this function name.

also, the convention is to use `cls` for the first variable in a classmethod.

PR #642, opened 2020-01-01
Lesson04, mailroom.py, bug fix!

lmburu-uw:
-fixed 'list' option error in prompt_for_donor_name()
-removed the use of tempfiles
-adjusted program to create local files in send_thank_you_letter_to_all()


lmburu-uw:
On Lesson05, Mailroom.py:
- Modified create_report() to use list comprehension 
- Added exception handling to the main() 

@charlesrobison, question for you. If a user selects option 1, and types in a value that's not a representation of a valid name, how you write code to handle that? For example, if i enter 'aninvalidname', it will show up in my database of contributors, which is not good. Thanks.

students/larry_mburu/Lesson05/mailroom.py
```diff
@@ -0,0 +1,203 @@
+#!/usr/bin/env python3
+
+import sys
+import os
```
charlesrobison: This import isn't being used?

charlesrobison:
@charlesrobison, question for you. If a user selects option 1, and types in a value that's not a representation of a valid name, how you write code to handle that? For example, if i enter 'aninvalidname', it will show up in my database of contributors, which is not good. Thanks.

Good question - but really beyond the scope of this assignment.  There really is no way to get around invalid input without having some type of login/validation.

PR #647, opened 2020-01-03
Adding files for lesson 6 "Unit Testing"

nskvarch:
adding the files for the lesson 6 exercises "Unit Testing"


students/nskvarch/Lesson6/mailroom4.py
```diff
@@ -0,0 +1,153 @@
...
+    # the Open operator so I can request input again rather than a try loop around the 'with open'
+    if custom != "":
+        while True:
+            customdir = input("Please specify a custom directory path:  ")
+            tested = os.path.exists(customdir)
+            if tested == True:
+                break
+            elif tested == False:
+                print("That directory or path does not exist, please specify a valid directory")
+        cwd = customdir
```
natasha-aleksandrova: the code can be cleaned up a bit here:
```
while True:
    custom_dir = input("specify custom dir...")
    if not os.path.exists(custom_dir):
        print("Invalid...")
    else:
        break
```

students/nskvarch/Lesson6/mailroom4.py
```diff
@@ -0,0 +1,153 @@
...
+            sizes[2]) + "    \t    " + line[3].rjust(sizes[3]))
+    print("\n")
+
+
+def run_report():
+    """take in the donor database, create organize and sort a table in the form of a list and then hand the table off
+    to the printer function """
+    data_lines = []
+    for name in donor_db:
+        data = donor_db[name]
```
natasha-aleksandrova: you can avoid this line altogether by using .items() function (the most common way to loop over dictionary.

```
for name, data in donor_db.items():
```

students/nskvarch/Lesson6/mailroom4.py
```diff
@@ -0,0 +1,153 @@
...
+    """use the sys module to clean-exit the script"""
+    print("Good bye")
+    sys.exit()
+
+
+def main():
+    """"main loop for the program to occupy while running, includes the main menu"""
+    while True:
+        response = input(menu)
+        switch_dict = {"1": lambda db=donor_db: thanks(db), "2": run_report, "3": create_thankyou_all}
```
natasha-aleksandrova: shouldn't need any lambda here, you also don't need to pass in donor_db as it is a global data structure.

`switch_dict` also should be defined outside of while True loop..

students/nskvarch/Lesson6/mailroom4.py
```diff
@@ -0,0 +1,153 @@
...
+        if response in switch_dict:
+            switch_dict[response]()
+        elif response == "4":
+            exit_program()
```
natasha-aleksandrova: this shouldn't be a special case here - map exit_program() function in the switch dict.

students/nskvarch/Lesson6/mailroomtest.py
```diff
@@ -0,0 +1,126 @@
...
+        output = io.StringIO()
+        with redirect_stdout(output):
+            print_report(test_datalines)
+        output.seek(0)
+        output1 = output.read()
+        assert output1 == test_report
+
+
+class test_case_two(unittest.TestCase):
+    def test_run_report(self):
```
natasha-aleksandrova: your tests should all live in a single class, and have multiple functions for each test case:
```
class TestMailroom(unittest.TestCase):
    def test_report(self):
        pass

    def test_add_donor(self):
        pass
```

PR #649, opened 2020-01-04
Added mailroom part 3 exercise and edits to except_exercise from last PR

N0vA4:
I got this to work but at the end was wondering if I should have used KeyError for a key not found in the arg_dict.keys()

### DIFF
--- a/students/N0vA/lesson_05/except_exercise.py
+++ b/students/N0vA/lesson_05/except_exercise.py
@@ -20,10 +20,10 @@ try:
     joke = fun(first_try[0])
 except NameError:
     print('There is no joke for topic - spam')
-try:
-    joke = fun(first_try[1])
-except NameError:
-    print('There is no joke for topic - cheese')
+    try:
+        joke = fun(first_try[1])
+    except NameError:
+        print('There is no joke for topic - cheese')
 
 # Here is a try/except block. Add an else that prints not_joke
 try:
--- a/students/N0vA/lesson_05/mailroom3.py
+++ b/students/N0vA/lesson_05/mailroom3.py
@@ -1,0 +1,133 @@
+#!/usr/bin/env python3
+#### Mailroom Project Part 2 ####
+
+import sys
+import pathlib
+
+# Create list of donors and their donation history
+database = {'Bill Gates': [2000000, 250000000], 
+            'Jeff Bezos': [2000000],
+            'Elon Musk': [50000000, 10000000], 
+            'Howard Schultz': [1000000], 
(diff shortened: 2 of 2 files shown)


natasha-aleksandrova:
@N0vA4 in reality either dict key check or KeyError are valid solutions, however - given that this lesson was about exception handling we would expect to see KeyError being caught. 

Not a huge deal, especially given that you do realize that the option is there.

students/N0vA/lesson_05/except_exercise.py
```diff
@@ -20,10 +20,10 @@
...
 except NameError:
     print('There is no joke for topic - spam')
-try:
-    joke = fun(first_try[1])
-except NameError:
-    print('There is no joke for topic - cheese')
+    try:
+        joke = fun(first_try[1])
+    except NameError:
+        print('There is no joke for topic - cheese')
```
natasha-aleksandrova: this would be considered a separate try/except block.

This should look like:
```
try:
    joke = fun(first_try[0])
except NameError:
    joke = fun(first_try[1])
```

students/N0vA/lesson_05/mailroom3.py
```diff
@@ -0,0 +1,133 @@
...
+        if name == 'list':
+            print(list(database.keys()))
+    
+    # Enter donation amount
+    amount = input('How much was their donations? ')
+    try: 
+        amount = int(amount)
+    except ValueError:
+        amount = input('Enter a valid donation amount: ')
+        amount = int(amount)
```
natasha-aleksandrova: what if they enter invalid number for the second time? 

your program will crash, ideally you want this prompt to be inside a while True loop that exists once a valid number is entered (int() cast passes)

students/N0vA/lesson_05/mailroom3.py
```diff
@@ -0,0 +1,133 @@
...
+        # Run functions for tasks based on user's response
+        try:
+            task = int(task)
+        except ValueError:
+            task = int(input('Please enter a valid option from 1-4: '))
+        else:
+            # I tried except KeyError here and that didn't work
+            if task not in arg_dict.keys():
+                task = int(input('Please enter a valid option from 1-4: '))
+        arg_dict.get(task)()
```
natasha-aleksandrova: if you used `arg_dict[task]` then you could catch KeyError

using `arg_dict.get()` will never throw a KeyError, even if key does not exist. It can throw a TypeError though, if key does not exist then `get()` will return `None` and executing `None()` will return `TypeError`

PR #650, opened 2020-01-04
lesson04 exercises and assignments

bromleysj:
I am stuck on the trigrams/kata assignment: when running the text "I wish I may I wish I might", I cannot get the values to update with additional matches e.g. ('wish', 'I'): ['may', 'might']; currently, I only get 'may'. 
Also, I am not sure how to process the trigram with the instruction "Remember that after adding a word to a pair to make a three-word text, the next pair is the last two words in that three-word text."  
Is the goal to just select any random tuple of words from the keys in my dict and then select a random word from the values of that pair?  Or, is the task to process the dict from the beginning, selecting only a random value for each tuple key sequentially?


students/scott_bromley/lesson04/trigrams.py
```diff
@@ -0,0 +1,66 @@
...
+
+def build_trigram(words):
+    """
+    build trigram out of words
+    :param words: list of words
+    :return: dict using tuple of word pairs as key and dict as values
+    """
+    trigram = {}
+    for i in range(len(words) - 2):
+        trigram.setdefault(tuple(words[i:i + 2]), words[i + 2])
```
natasha-aleksandrova: >  I cannot get the values to update with additional matches

you're never appending values to your dictionary here; the signature should be: 
```
newdata.setdefault(<key>, []).append(<value>)
```
I prefer to use `defaultdict`:
```
from collections import defaultdict
trigrams = defaultdict(list)

....
trigrams[key].append(value)
```

natasha-aleksandrova:
@bromleysj for

> Also, I am not sure how to process the trigram with the instruction "Remember that after adding a   word to a pair to make a three-word text


from instructions:
> "I wish" => ["I", "I"]
"wish I" => ["may", "might"]
"may I"  => ["wish"]
"I may"  => ["I"]

above should represent your dictionary, then

> To generate new text from this analysis, choose an arbitrary word pair as a starting point. Use this pair of words to look up a random next word (using the table above) and append this new word to the text so far.


so:
1. You're first going to pick a random key from your dictionary as a starting point. Let's say we start with "I wish".
2. Grab a word from a list associated with the key "I wish", let's grab "I" and create a text that should have a total of 3 words => "I wish I"
3. Now grab the last 2 words from the new generated text => "wish I", this is going to be your new key to use.
4. Repeat from step 2. In step 1 we picked a random key as a starting point, but now you have a key to use ("wish I").

Hopefully this clarifies.

natasha-aleksandrova:
@bromleysj I'm going to go ahead and close this PR - both assignments have issues.

Please test your mailroom by adding a new donor and adding donation to existing donor and run the report.
Both cases produce a crash.

bromleysj:
Natasha,

I've resubmitted both the mailroom2 and trigrams .py files.

-Scott

PR #651, opened 2020-01-04
lesson 09 submission

cjfortu:


students/cjfortu/lesson09/mailroom_00/donor_models.py
```diff
@@ -0,0 +1,93 @@
...
+    def lifetime_donations_sum(self):
+        return sum(self.donations)
+    
+    @property
+    def average_donations(self):
+        return self.lifetime_donations_sum / self.number_donations
+        
+    def add_donations(self, new_donations):
+        """Append new donations to the existing donations attribute."""
+        [self.donations.append(donation) for donation in new_donations]
```
natasha-aleksandrova: this is an invalid use of list comprehension

a list comprehension should produce an output that would get stored in a variable, right now you are generation a list of None values and not store it anywhere.

students/cjfortu/lesson09/mailroom_00/donor_models.py
```diff
@@ -0,0 +1,93 @@
...
+
+class Donor:
+    """Class that holds and accesses information about a single donor."""
+    new_donations = []
+    
+    def __init__(self, donor_name=None, *donations):
+        """Initiate a donor by name and donations."""
+        self.name = donor_name
+        self.donations = [*donations]
+        self.new_donations = [*donations]
```
natasha-aleksandrova: what is the purpose of the two different donations types?

students/cjfortu/lesson09/mailroom_00/donor_models.py
```diff
@@ -0,0 +1,93 @@
...
+    def add_donations(self, new_donations):
+        """Append new donations to the existing donations attribute."""
+        [self.donations.append(donation) for donation in new_donations]
+        self.new_donations = new_donations
+        
+    @property
+    def compose_letter(self):
+        """Generate text for a single letter."""
+        text = "\n\nHi {},\n\nThank you for your total donation of ${:,.2f}.\n\n\nVR\n\nThe Mailroom\n(555) 555-5555".\
+            format(self.name, sum(self.new_donations))
```
natasha-aleksandrova: you should only print the most recent donation here `self.donations[-1]` should satisfy this.

students/cjfortu/lesson09/mailroom_00/donor_models.py
```diff
@@ -0,0 +1,93 @@
...
+    def mass_letter(self):
+        """Generate text for letters for all donors."""
+        text = "\nHi {},\n\nThank you for your lifetime donations of ${:,.2f}.\n\n\nVR\n\nThe Mailroom\n(555) 555-5555".\
+            format(self.name, self.lifetime_donations_sum)
+        return text
+        
+    
+class DonorCollection:
+    """Class that holds all donor objects, as well as methods for multiple donors."""
+    data = {}
```
natasha-aleksandrova: your data should not be a class variable, it should be tied to an instance only

students/cjfortu/lesson09/mailroom_00/donor_models.py
```diff
@@ -0,0 +1,93 @@
...
+        self.data = {key: Donor(key, *value) for key, value in start_data.items()}
+        
+    def add_donor(self, donor_name = None, *donations):
+        """Add a donor using the Donor.__init__ method."""
+        self.data[donor_name] = Donor(donor_name, *donations)
+        
+    @property
+    def names(self):
+        """Build a list of donor names."""
+        return [self.data[donor_name].name for donor_name in self.data]
```
natasha-aleksandrova: the keys in `self.data` represent donors names so `return list(self.data)` should be sufficient

students/cjfortu/lesson09/mailroom_00/donor_models.py
```diff
@@ -0,0 +1,93 @@
...
+        """Build a new data structure for use by send_all and create_a_report."""
+        key1 = 'Donor Name'
+        key2 = '# Gifts'
+        key3 = 'Total Given($)'
+        key4 = 'Average Gift'
+        # leveraged comprehensions and exceptions here!
+        try:
+            report_data = [{key1: self.data[donor_name].name, key2: self.data[donor_name].number_donations,\
+                key3: self.data[donor_name].lifetime_donations_sum, key4: self.data[donor_name].average_donations}
+                for donor_name in self.data]
```
natasha-aleksandrova: this is bit hard to read- also you can just loop over donor objects directly and reduce length of code:
```
keys = ('Donor Name', '# Gifts')
report_data = [dict(zip(keys, (donor.name, donor.number_donations)) for donor in self.data]
```

students/cjfortu/lesson09/mailroom_00/donor_models.py
```diff
@@ -0,0 +1,93 @@
...
+        except ZeroDivisionError:
```
natasha-aleksandrova: I think your average donation should return zero instead of throwing an exception.

students/cjfortu/lesson09/mailroom_00/donor_models.py
cjfortu: I figured there was a chance the donor could make several donations at once.  I would log the individual donations in the data, but send a single letter addressing all donations.  So **donations** is the record of all donations, while **new_donations** is whatever was just donated, either a single donation or group of donations submitted simultaneously.

students/cjfortu/lesson09/mailroom_00/donor_models.py
cjfortu: Commenting this line out does cause my tests to fail, including "test_add_donations".  I looked at that test again, and it seems valid.  

Also, I manually run this comprehension in command line, and it achieves the intended effect of looping to append the list represented by **self.donations**.  

Could you help me understand more?  It seems to work, so is it bad Pythonic practice but functional?  Thank you.

students/cjfortu/lesson09/mailroom_00/donor_models.py
cjfortu: This goes with my comment above about **donations** vs **new_donations**

students/cjfortu/lesson09/mailroom_00/donor_models.py
natasha-aleksandrova: Correct it is bad practice.

```
>>> l = []
>>> [l.append(i) for i in range(3)]
[None, None, None]
```
 you are allocating an array of 3 elements with None values.

that code should be converted to a regular for loop OR you don't need a loop there at all:
```
self.donations.extend(new_donations)
```

PR #654, opened 2020-01-06
Added a try/except block to handle non numerical donation values

Freemabe:
I think this is the only place an error would show up, please let me know if I missed something however. As always thanks for looking over my work.

### DIFF
--- a/students/Ben_freeman/lesson05/mailroom3.py
+++ b/students/Ben_freeman/lesson05/mailroom3.py
@@ -1,0 +1,84 @@
+import sys
+#Donor dict
+donor_list={
+"Ahmed":[123,456,789],
+"Sally": [10,11],
+"Bob": [12],
+"Vladamir": [13.14,1516.17],
+"Santa": [181920.21,22,23.24],
+}
+
+
+#send thanks funct
+def send_thanks():
+    while True:
+        query=input("Please enter a Full Name, enter 'list' to see a list of current donors, or quit to exit: ")
+
+        if query=="list":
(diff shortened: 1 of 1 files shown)


students/Ben_freeman/lesson05/mailroom3.py
```diff
@@ -0,0 +1,84 @@
...
+        if query=="list":
+            print('\n'.join(donor_list))
+        elif query.lower()=="quit":
+            return
+        else:
+            tracker=0
+            while tracker==0:
+                try:
+                    new_donate=float(input("\nHow much would you like to donate: \n"))
+                    tracker=1
```
natasha-aleksandrova: you can avoid having a "flag" variable (tracker)

just a add a `break` on line 26 instead of `tracker=1`

students/Ben_freeman/lesson05/mailroom3.py
```diff
@@ -0,0 +1,84 @@
...
+#Script
+if __name__ == '__main__':
+    while True:
+        prompt=input(
+            "What would you like to do?:\n"
+            "A - Send a thank you\n"
+            "B - Create a report\n"
+            "C - Send a thank you to all donors\n"
+            "D - Quit\n")
+        if prompt.upper() in ["A","B","C","D"]:
```
natasha-aleksandrova: this can be `if prompt.upper() in menu` - no need to hardcode those selections.

also, you should simply catch a `KeyError` if they do enter invalid option.

PR #655, opened 2020-01-06
Add lesson04 assignment for choltzman

asemlef:


students/choltzman/lesson04/mailroom.py
```diff
@@ -0,0 +1,125 @@
...
+        },
+        {
+            "name": "Jeff Bezos",
+            "gifts": [1234567.89]
+        },
+        {
+            "name": "Paul Allen",
+            "gifts": [1, 2, 3, 4, 5, 6, 7, 8, 9]
+        }
+]
```
charlesrobison: Note the instructions for the assignment, "
Convert your main donor data structure to be a dict."  You have a list of dictionaries here.  Can you make it one dictionary, with donors as keys?

PR #661, opened 2020-01-09
Updated except exercises and mailroom3 after comments from last PR.  

N0vA4:
For mailroom3 I could not get the main function when the program starts to work properly if I converted the task input to an integer.  I guess it seems like an unnecessary step for this assignment.  Basically if I would do task = int(input(...)) and I would enter a string I could get it to give me my exception error catch but if I entered a string a second time the program would crash.


natasha-aleksandrova:
@N0vA4 do you mean convert to integer the main selections? I don't think it makes sense since they are just selections and don't really need to be integers.

PR #667, opened 2020-01-12
slicing.py

ikempcode33:
Ready for grading!


students/Isabella_Kemp/lesson03/slicing.py
```diff
@@ -0,0 +1,58 @@
+# Isabella Kemp
+# 1/12/2020
+# Slicing Lab
+
+# Exchanges first and last character in a string
+def exchange_first_last(seq):
+    first = seq[0:1]
```
natasha-aleksandrova: note that its common to omit default slice values (like zero)

PR #675, opened 2020-01-16
Fixed multiple prints in create_report

bromleysj:
Fixed multiple prints in create_report; however, I'd like some direction on creating a wrapper for my get_user_input function (I've commented out the relevant code).

### DIFF
--- a/students/scott_bromley/lesson05/mailroom3.py
+++ b/students/scott_bromley/lesson05/mailroom3.py
@@ -133,13 +133,11 @@ def create_report():
                                  statistics.mean(donors.get(donor))])
         except IndexError as donor_summary_err:
             raise donor_summary_err
-        else:
-            donor_summary.sort(key=lambda x: float(x[1]), reverse=True)
-            print(f"Donor Name{'':<20} | Total Given{'':>0} | Num Gifts{'':>0} | Average Gift{'':>0}")
-            print(f"-" * 72)
-        finally:
-            for name, total, num_gift, avg_gift in donor_summary:
-                print(f"{name:<32}${total:>11.2f}{num_gift:>12}  ${avg_gift:>13.2f}")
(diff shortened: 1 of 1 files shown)


natasha-aleksandrova:
@bromleysj your wrapper is an overkill for your use case - you can just wrap the input() call directly with try/except without the need for wrapper

natasha-aleksandrova:
when I uncommented your wrapper function the only error I got was:
```
  File "student.py", line 1, in <module>
    @safe_input
NameError: name 'safe_input' is not defined
```

Which means it cannot find your safe_input function, this is because you are using before defining it. If you move the definition above its usage then it should work.

bromleysj:
Natasha,

I have a question re:lesson06 unittest: the examples provided by
https://nedbatchelder.com/text/test0.html all have the Portfolio as its own
standalone class; how are we to change our mailroom code into a class that
it too can be instantiated even though we haven't covered this material
yet?

Thanks,

Scott

bromleysj:
One more question: what is the final deadline for all work to be submitted?

I want to make sure I'm on track to have everything done.

Scott

natasha-aleksandrova:
> One more question: what is the final deadline for all work to be submitted? I want to make sure I'm on track to have everything done. Scott


your due date is 3/18

natasha-aleksandrova:
@bromleysj mailroom will be converted into a class in Lesson 09.


For now you will be unit testing just the individual functions, the tutorial in Lesson 06 should have some tips on what parts to unit test:
https://uwpce-pythoncert.github.io/PythonCertDevel/exercises/mailroom-part4.html#guidelines


Two common libraries to use for unit testing are `pytest` and `unittest`

`pytest` uses functions style to write tests (and more preferred in the community)
`unittest` is very similar to Java unittest and is a class style


I would suggest `pytest` for Lesson 06.

PR #676, opened 2020-01-16
Finished mailroom4 and unit testing for mailroom4

N0vA4:
I left a note for one of my unit testing functions that I had trouble with or maybe understand the best way to tackle the problem might be a better way of putting it.


students/N0vA/lesson_06/test_mailroom4.py
```diff
@@ -0,0 +1,87 @@
...
+    report_data = format_report(sort_database(database))
+    # Format a row to test
+    data = sort_database(database)
+    member = []
+    member.append(data[1])
+    
+    # Test donor with multiple donations to ensure total and avg columns work properly
+    line_format = '{:<24}{:^5} ${:>14,}{:^5} {:^5}{:^5} ${:>14,.2f}'
+    member_row = line_format.format(member[0][0], ' ', int(sum(member[0][1])), ' ', round(len(member[0][1]),2), ' ', round(sum(member[0][1])/len(member[0][1]),2))
+    assert member_row == report_data[1]
```
natasha-aleksandrova: you have two options here:
1. generate a list of tuples that contains your summary data, then assert against this list
```
assert [('Bob', 100 <avg>, 100<total>,1<numof>)] == generate_report()
```
2. return a fully formatted table and assert against that
assert generate_report_table() == """
your table data here
"""

PR #680, opened 2020-01-17
Think I got everything working for project 7, did not enjoy the directions for this.

Freemabe:
The directions did not seem super helpful, and at times I was lost on the tutorial as well. Mostly it was because of issues like:

``Create a `Header` class – this one should take an integer argument for the header level. ...``

Unlike all of the other instances of class names before this in this assignment, the name given, `Header` is not what the run_html_render.py file is expecting, and instead wants `H`. Its easy to understand what the issues were once you get it, but its kind of frustrating in the beginning.

Over all (once I was done) I liked the assignment, and this might be because of some knowledge gaps on my part, but it seems like the vagueness of the directions jumped significantly between this assignment and the last few.

ALSO:

I'm getting this message when i run pytest, and I can't identify the problem.

`DeprecationWarning: invalid escape sequence \h`

Cheers hope this update wasn't too whiny.

### DIFF
--- a/students/Ben_freeman/lesson07/html_render.py
+++ b/students/Ben_freeman/lesson07/html_render.py
@@ -7,12 +7,116 @@ A class-based system for rendering html.
 
 # This is the framework for the base class
 class Element(object):
+    tag = "html"
+    indent = "   "
 
-    def __init__(self, content=None):
-        pass
+    def __init__(self, content=None, **kwargs):
+        self.attributes = kwargs
+        if content is None:
+            self.contents = []
+        else:
+            self.contents = [content]
 
     def append(self, new_content):
-        pass
+        self.contents.append(new_content)
 
-    def render(self, out_file):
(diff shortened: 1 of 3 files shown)


natasha-aleksandrova:
@Freemabe thanks for the feedback - would you say this was due not being familiar with HTML concepts? I can see that making this assignment a lot more challenging.

Freemabe:
> @Freemabe thanks for the feedback - would you say this was due not being familiar with HTML concepts? I can see that making this assignment a lot more challenging.

Yeah, that definitely could have been it. I haven't really interacted with html in like 15 years. Mostly I just was confused as to what was being asked of me, as most of the coding was pretty straight forward once I figured out what I was supposed to do. Looking at the directions again they don't seem as arcane as they did the first time around so I suspect I was just getting frustrated with not knowing what some of the words meant.

PR #683, opened 2020-01-19
lesson09 assignment

joli-u:
completed oo mailroom assignment


students/joli-u/lesson09/donor_models.py
```diff
@@ -0,0 +1,110 @@
...
+import pathlib
+
+"""
+donor_models.py: module containing donor classes
+lesson 9
+joli umetsu
+python210
+"""
+
+class Donor(object):
```
natasha-aleksandrova: note that in python3 its just `class Donor:`

PR #684, opened 2020-01-19
ganyoa adding mailroom.py

ganyoa:
Hi, please see mailroom.py added to my student/lesson03 folder. Thanks, Anthony Ganyo

### DIFF
--- a/students/ganyoa/lesson03/mailroom.py
+++ b/students/ganyoa/lesson03/mailroom.py
@@ -1,0 +1,95 @@
+import sys
+
+#main menu; selection will guide the 'main' function
+prompt = "\n".join(("\n""Mailroom options",
+          " Please choose from below options:",
+          " 1 - Send a Thank You",
+          " 2 - Create a Report",
+          " 3 - Quit",
+          ">>> "))
+
+#list of current donor and donation amounts
+donor_list = [("Mouth Devereaux", [4562.58, 3817.39]),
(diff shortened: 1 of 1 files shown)


students/ganyoa/lesson03/mailroom.py
```diff
@@ -0,0 +1,95 @@
...
+    for i in range(len(donor_list)):
+        if donor_name == donor_list[i][0]:
+            match = i
+            break
+    else:
+        donor_list.append((donor_name,[]))
+        match = i + 1
+    donation_amt = input("\n""How much did " + donor_name + " donate? > ")
+    amt = round(float(donation_amt),2)
+    donor_list[match][1].append(amt)
```
natasha-aleksandrova: you can move the amount prompt above your for loop and the donation right in place then you don't need to keep track of index.

students/ganyoa/lesson03/mailroom.py
```diff
@@ -0,0 +1,95 @@
...
+
+    #if user does not enter a name or 'list' they will be brought back to the main menu
+    if donor_name == '':
+        main()
+
+    #displays previous donor list
+    if donor_name == 'List':
+        for i in range(len(donor_list)):
+            print(donor_list[i][0])
+        main()
```
natasha-aleksandrova: you should have a `return` statement here instead of calling back to main() (same comment for line 33)

PR #691, opened 2020-01-21
Lesson06 Assignment, partially complete

lmburu-uw:
Please take a look at my Unit Test, on Lesson06. If you are asked for an amount, please enter 100. I'm stuck and not sure why the last 2 tests are failing. There is an extra line or tab somewhere, that i can't seem to find.

### DIFF
--- a/students/larry_mburu/Lesson06/mailroom.py
+++ b/students/larry_mburu/Lesson06/mailroom.py
@@ -1,0 +1,213 @@
+    #!/usr/bin/env python3
+
+import sys
+import os
+from operator import itemgetter
+
+donations_per_individual = { 
+    'William Gates III' : [50000.0, 500.12], 
+    'Mark Zuckerberg' : [20000.0, 8500.99], 
+    'Jeff Bezo': [100000.0, 40000.0, 700.99], 
+    'Paul Allen': [200000.0, 1440.0, 300.00], 
+    'Bill Gates': [30000.0, 70000.0, 450.0]
+}
+
+def display_menu():
+    """
+    Display menu option to user
+    """
+    print("Welcome to Mail Automator (MA)!")
+    print()
+    print("[1] Send a Thank You!")
+    print("[2] Create a report!")
+    print("[3] Send letters to all donors")
+    print("[4] Quit!")
+    print()
+
+def send_thank_you_letters(): 
+    """
+    Function, sends a thank you letter to all donor's in the donor DB list,
+    by writing thank you letters to the local filesystem.
+    """
+    print('[+] Generating thank you letters..')
+
+    for donor_name in donations_per_individual:  
+        #split and join,to prevent file names with spaces.
+        with open('_'.join(donor_name.split()) + '.txt', 'w') as fobj: 
+            fobj.write(get_letter_text(donor_name))
+
+    print('[+] Letter generation complete.')
+    print('[+] Letters are in your current working directory')
+
+def get_letter_text(donor_name):
+    """
+    returns formatted string, for content to be written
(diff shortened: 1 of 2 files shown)


students/larry_mburu/Lesson06/test_mailroom.py
```diff
@@ -0,0 +1,98 @@
...
+        ]
+    assert actual_summary == mailroom.get_report()
+
+#test send_thank_you_letters() function
+
+def test_send_thank_you_letters(): 
+    donor_names = list(mailroom.donations_per_individual)
+    expected_list = [ '_'.join(donor.split()) for donor in donor_names ] 
+    for expected in expected_list: 
+        assert os.path.isfile(expected + '.txt') == True
```
natasha-aleksandrova: what failure do you get on this function?

you're not actually calling any of your mailroom code here, each unit test should execute a single function to test.

students/larry_mburu/Lesson06/test_mailroom.py
```diff
@@ -0,0 +1,98 @@
...
+    for donor_name in donor_names: 
+        expected = f"""Dear {donor_name},
+        Thank you for your kind donation of ${mailroom.donations_per_individual[donor_name][-1]}. 
+            
+        It will be put to very good use. 
+            
+        Sincerely, 
+            
+        - The Cloud Squad"""
+        assert mailroom.get_letter_text(donor_name) == expected
```
natasha-aleksandrova: yes, it is a white space issue.

if you copy the same template construct as in your get_letter_text() to here it will work.

students/larry_mburu/Lesson06/test_mailroom.py
```diff
@@ -0,0 +1,98 @@
...
+
+    assert actual_sorted == expected_sorted
+
+    # test the create_report() returned results.
+    padding = 20
+    actual_summary = [ 
+        f"{name:{padding}} $ {total:10.2f} {num_gifts:14} $ {average:10.2f}" 
+        for name, total, num_gifts, average in expected_sorted
+        ]
+    assert actual_summary == mailroom.get_report()
```
natasha-aleksandrova: there is quite a bit in this unit test.. your first two asserts are asserting on just some code in this function, not testing your mailroom code.

your unit test functions should execute your mailroom code function and then assert on result or expected changes from the function call.

students/larry_mburu/Lesson06/test_mailroom.py
```diff
@@ -0,0 +1,98 @@
...
+
+def test_add_donor_to_list(): 
+    '''
+    test new user not in DB
+    '''
+    mailroom.add_donar_to_list('James Comey')
+    assert mailroom.donations_per_individual['James Comey'] == [100.0]
+
+def test_add_to_donation_database(): 
+    assert mailroom.add_to_donation_database('James Comey', 100) == None
```
natasha-aleksandrova: this doesn't really test anything, right? a function that has not return value can't really be tested by asserting that it returns nothing.

students/larry_mburu/Lesson06/mailroom.py
```diff
@@ -0,0 +1,213 @@
...
+    
+    add_donar_to_list(donor_name)
+    
+def get_donor_list(donor_name): 
+    """
+    Gets a current listing of donors 
+    Param: donor_name: if list, return a list of donors
+    """
+    if donor_name == 'list': 
+        return list(donations_per_individual) #list of donors
```
natasha-aleksandrova: why is this function taking in param? it's only purpose should be to return donor names

PR #692, opened 2020-01-21
Updated lesson_07 with files for html_render assignment

N0vA4:
This assignment was pretty challenging but there were some great directions to help out.


students/N0vA/lesson_07/html_render.py
```diff
@@ -0,0 +1,159 @@
...
+        self.attributes = kwargs
+        if content is None:
+            self.contents = []
+        else:
+            self.contents = [content]
+        
+
+    def append(self, new_content): # Add another string to content
+
+       return self.contents.append(new_content)
```
natasha-aleksandrova: you would not have a return statement here, append() has not return value.

students/N0vA/lesson_07/html_render.py
```diff
@@ -0,0 +1,159 @@
...
+    tag = 'ul'
+
+
+class Header(OneLineTag):
+
+    tag = 'header'
+
+    # Add level to init call
+    def __init__(self, level, content=None, **kwargs):
+        self.tag = 'header{}'.format(level)
```
natasha-aleksandrova: this should be `h{level}`

PR #693, opened 2020-01-21
Lesson 9 Exercise

amax1213:


students/allen_maxwell/Lesson_09/donor_models.py
```diff
@@ -0,0 +1,152 @@
...
+    '''Checks the name entry for a Full Name (First and Last).'''
+    try:
+        name.split()[1]
+        name = name.title()
+        return name.title()
+    except TypeError:
+        raise ('Name must contain a first and last name')
+
+
+def check_donation(donation):
```
natasha-aleksandrova: note that these methods would be good as staticmethods

students/allen_maxwell/Lesson_09/donor_models.py
```diff
@@ -0,0 +1,152 @@
...
+            return donation
+        else:
+            raise AttributeError('Donations must be positive values, greater than zero')
+
+    @property
+    def num_donations(self):
+        '''Returns number donations.'''
+        return self._donations[1]
+
+    @staticmethod
```
natasha-aleksandrova: this wouldn't be a static method if you're using an instance via `self`

students/allen_maxwell/Lesson_09/donor_models.py
```diff
@@ -0,0 +1,152 @@
...
+
+    @avg_donations.setter
+    def avg_donations(self, value):
+        '''Blocks setting average donations for the donor.'''
+        raise AttributeError('Average donation cannot be set')
+
+    def new_donation(self, donation):
+        '''Adds a new donation to the donors donation data.'''
+        self.add_total_donations(self, donation)
+        self.add_num_donations(self)
```
natasha-aleksandrova: are you never storing the raw list of donations?

students/allen_maxwell/Lesson_09/donor_models.py
```diff
@@ -0,0 +1,152 @@
...
+    @property
+    def donors(self):
+        '''Returns a collection of donors.'''
+        return self._donors
+
+    def is_donor_present(self, donor_name):
+        '''Checks if the donor is present in the collection.'''
+        for donor in self.donors.values():
+            if donor_name == donor.name:
+                return True
```
natasha-aleksandrova: why would you loop through the dictionary? the key is the name? how could you utilize a dictionary here?

students/allen_maxwell/Lesson_09/donor_models.py
amax1213: These 2 functions update the raw data to the donations list. Perhaps these are redundant but I decided to make the 2 calls more modular.

students/allen_maxwell/Lesson_09/donor_models.py
amax1213: Updated changes in Lesson 9 Assignment Update

students/allen_maxwell/Lesson_09/donor_models.py
amax1213: Updated changes in Lesson 9 Assignment Update

students/allen_maxwell/Lesson_09/donor_models.py
amax1213: Updated changes in Lesson 9 Assignment Update

PR #695, opened 2020-01-23
lesson08 assignment,

andzak10:
Adding circle class and test circle class, lesson 08 assignment

### DIFF
--- a/students/A_Zach/lesson08/circle_class.py
+++ b/students/A_Zach/lesson08/circle_class.py
@@ -1,0 +1,84 @@
+"""
+Create a class for circles
+"""
+
+from math import pi
+
+class Circle():
+    def __init__(self, radius):
+        self.radius = radius
+
+
+    def __str__(self):
+        return f"Circle with radius: {self.radius:.6f}"
+
+
+    def __repr__(self):
+        return f"Circle({self.radius})"
+
+
+    @classmethod
+    #Can create a circle from diameter instead
+    def from_diameter(cls,diameter):
+        self = cls(diameter/2)
+        return self
+
+
+    @property
+    def diameter(self):
+        return self.radius * 2
+
+
+    @diameter.setter
+    def diameter(self, value):
+        self.radius = value/2
+
+
+    @property
+    def area(self):
+        return pi * self.radius**2
+
+
+    def __add__(self, other):
+        sum = self.radius + other.radius
+        return Circle(sum)
+
+
+    def __mul__(self,other):
+        prod = self.radius * other
+        return Circle(prod)
+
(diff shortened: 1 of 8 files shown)


students/A_Zach/lesson08/circle_class.py
```diff
@@ -0,0 +1,84 @@
...
+"""
+
+from math import pi
+
+class Circle():
+    def __init__(self, radius):
+        self.radius = radius
+
+
+    def __str__(self):
```
natasha-aleksandrova: note that in classes, there is only one line of space between methods.

students/A_Zach/lesson09/donor_models.py
```diff
@@ -0,0 +1,71 @@
...
+        # define a dictionary for donors and their donations
+        self.donor_list = []
+    
+    
+    def new_donor(self, donor: Donor):
+        # Add a new donor
+        self.donor_list.append(donor)
+    
+    
+    def sort_donation(self, name, donation):
```
natasha-aleksandrova: a bit of a confusing name for this method

students/A_Zach/lesson09/donor_models.py
```diff
@@ -0,0 +1,71 @@
...
+                return
+        new_donor = Donor(name.title(), [donation])
+        self.new_donor(new_donor)
+    
+    def list_donors(self):
+        # Generate text list of donors name to be printed
+        donor_list_text = ''
+        for donor in self.donor_list:
+            donor_list_text += donor.name + "\n"   
+        return donor_list_text
```
natasha-aleksandrova: you could make this one line: `return "\n".join([donor.name for donor in self.donor_list])`

students/A_Zach/lesson09/donor_models.py
```diff
@@ -0,0 +1,71 @@
...
+    
+    def email_all(self):
+        for donor in self.donor_list:
+            self.compose_email(donor.name)
+    
+    
+    def build_report(self):
+        report = []
+        [report.append((donor.name, sum(donor.donation), len(donor.donation), sum(donor.donation)/len(donor.donation))) for donor in self.donor_list]
+        return report
```
natasha-aleksandrova: this is not the correct way to use comprehensions; fixed:
```
return [(donor.name, sum(donor.donation), len(donor.donation), sum(donor.donation)/len(donor.donation)) for donor in self.donor_list]
```

natasha-aleksandrova: also.. the sum, total and average values should be provided as methods/properties of Donor class

PR #700, opened 2020-01-24
Lesson 9 assignment

Eric74nyc:
Final assignment, OO version of mailroom. 3 files included, mailroom_oo.py (logic), test_mailroom_oo.py (pytest cases), and cli_main.py (main program and command line interface).


students/eric_grandeo/lesson09/mailroom_oo.py
```diff
@@ -0,0 +1,199 @@
...
+            Generate text for email thanking donor for their last donation
+
+    """
+    def __init__(self, name, donations=[]):
+        self.name = name
+        self.donations = donations
+
+
+    @property
+    def donor(self):
```
natasha-aleksandrova: A common pattern is to have a method named `to_dict()` (makes it really clear what to expect

students/eric_grandeo/lesson09/mailroom_oo.py
```diff
@@ -0,0 +1,199 @@
...
+        thank_you_letter():
+            Generate text for email thanking donor for their last donation
+
+    """
+    def __init__(self, name, donations=[]):
+        self.name = name
+        self.donations = donations
+
+
+    @property
```
natasha-aleksandrova: note that in a class methods only need one line space between definitions

students/eric_grandeo/lesson09/mailroom_oo.py
```diff
@@ -0,0 +1,199 @@
...
+    def sum_donations(self):
+        '''Return total donations made by donor'''
+        self._sum_donations = sum(self.donations)
+        return self._sum_donations
+
+    @property
+    def num_donations(self):
+        '''Return the number of donations made by donor'''
+        self._num_donations = len(self.donations)
+        return self._num_donations
```
natasha-aleksandrova: if you have a property then no need to actually store the value in an attribute

students/eric_grandeo/lesson09/mailroom_oo.py
```diff
@@ -0,0 +1,199 @@
...
+                donors.add_donation(amount)
+                return
+        
+        new_donor = Donor(name, amount)
+        self.donor_list.append(new_donor)
+
+    @property
+    def donor_names(self):
+        '''Returns a list of donor names'''
+        donor_name = []
```
natasha-aleksandrova: make sure that your data structures (which are plural) are represented by its name, so in this case you're holding `donor_names`

PR #705, opened 2020-01-26
lesson 9

srmorehouse:
lesson 9, new o-o mailroom

### DIFF
--- a/students/srmorehouse/Lesson09/donor_models.py
+++ b/students/srmorehouse/Lesson09/donor_models.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python3
 
+import operator
 import os
 import sys
 
@@ -51,12 +52,7 @@ class Donor (object):
         return (self.name, sum_donations, len_donations, average_donation)
     
     
-    def sort_key(item):
-        return item[1]
-    
-    
 class DonorCollection():
-
 
     def __init__(self):
         self.donors={}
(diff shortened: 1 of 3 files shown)


students/srmorehouse/Lesson09/donor_models.py
```diff
@@ -0,0 +1,81 @@
...
+"""
+Steve Morehouse
+Lesson 09
+"""
+
+class Donor (object):
+
+    def __init__(self, name, donations=None):
+        self.name = name
+        self.donations = donations if donations is not None else []
```
natasha-aleksandrova: you can shortcut this to -> `self.donations = donations or []`

students/srmorehouse/Lesson09/donor_models.py
```diff
@@ -0,0 +1,81 @@
...
+    
+    def list_donors (self):
+        return_list = ''
+        for donor in self.donors:
+            return_list += f'{donor}\n'
+        return return_list
+
+
+    def donor_exists(self, name):
+        return any(d for d in self.donors.values() if d.name.upper() == name.upper())
```
natasha-aleksandrova: since you're storing donor data as a dictionary, you should be able to easily access it via dictionary key:

```
return bool(self.donors.get(name))

```

if you're worried about string case then I would recommend storing it in all lower case and then adjust on display

PR #707, opened 2020-01-27
Added files for circle class assignment

N0vA4:


students/N0vA/lesson_08/circle.py
```diff
@@ -0,0 +1,156 @@
...
+
+        if isinstance(val, Circle):
+            # Add two circles
+            return Circle(self.radius + val.radius)
+        elif isinstance(val, int):
+            # Add to a radius
+            return Circle(self.radius + val)
+        
+
+    def __sub__(self, val): # Add two circles
```
natasha-aleksandrova: note that you only need 1 line space between methods in  a class

PR #708, opened 2020-01-28
Lesson 04

StrafingRun:
Lesson_04/  is ready for Review


students/grant_dowell/Lesson_04/mailroom_part2.py
```diff
@@ -0,0 +1,121 @@
...
+#            print(note)
+#            print('\n')
+#        elif cmd == '2' or cmd.lower() == 'create a report':
+#            report(donor_db)
+#        else:
+#            continue
+
+    menu = {"1":thanks, "2": report, "3":log_all_letters, "4": quitter}
+    cmd = None
+    while cmd != '4':
```
natasha-aleksandrova: this is not necessary, when user selects option "4" your `quitter()` function should call `sys.exit()` to exit the script.

so in this case your `while True` loop would just terminate

students/grant_dowell/Lesson_04/mailroom_part2.py
```diff
@@ -0,0 +1,121 @@
...
+            else:   # If name not found in db, add it
+                db[name] = []
+    else:
+        print(name)
+        donation = input("Please enter a donation amount: ")
+        if donation.lower() != 'quit':  #Check if user is attempting to quit
+            db[name].append(float(donation))
+            thank_you = generate_letter(name)
+        print("\n" + thank_you + "\n")
+    return db, thank_you
```
natasha-aleksandrova: I don't believe you need this return data?

students/grant_dowell/Lesson_04/mailroom_part2.py
```diff
@@ -0,0 +1,121 @@
...
+        if name.lower() == 'list':
+            for name in db.keys():
+                print(name)
+        elif name.lower() == 'quit':
+            break
+        else:
+            vld_name = True
+            for _ in db:    # Check for name in db
+                if name.lower == name.lower:
+                    break
```
natasha-aleksandrova: this is incorrect - because you're using a dictionary you should utilize the efficient key look up.

I also don't understand the purpose of this loop or why you would compare two identical names which would always be true and this would actually compare function objects. This also causes the program to crash when adding new donors.
```
if db.get(name.lower())
```

students/grant_dowell/Lesson_04/mailroom_part2.py
StrafingRun: Nope, deleted

students/grant_dowell/Lesson_04/mailroom_part2.py
StrafingRun: I was having some trouble with this giving a Traceback originally. I did a bit more digging though, and I think it was a result of running with the '-i' option. Fixed.

students/grant_dowell/Lesson_04/mailroom_part2.py
StrafingRun: Thanks for the feedback. I also altered the db structure to use all lowercase keys and just make the input lower case from the get-go.

natasha-aleksandrova: 👍

PR #711, opened 2020-01-30
Add remaining assignments from choltzman

asemlef:


students/choltzman/lesson09/donor_models.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def get_donor(self, name):
+        """Return a donor object by name. Return None if not found."""
+        if name in self.donors:
+            return self.donors[name]
+
+    @property
+    def names(self):
+        """Return a list of donor names."""
+        return [name for name in self.donors]
```
natasha-aleksandrova: this can just be -> ` return list(self.donors)`

PR #712, opened 2020-01-31
My OO version of the Mailroom exercise

Freemabe:
Lemme know if there is anything I need to fix! Thanks so much.


students/Ben_freeman/lesson09/Mailroom_OO/mailroom_OO.py
```diff
@@ -0,0 +1,91 @@
...
+    def mass_send_thanks(self):
+        """this is the method that writes thank you letters to all of our donors, and places them in one folder to
+        avoid headache """
+        try:
+            os.mkdir("Letters")
+        except FileExistsError:
+            pass
+        for item in self.donors:
+            with open(f"Letters/{item}.txt", "w+") as letter:
+                letter.write(str(Donor(self.donors[item],item)))
```
natasha-aleksandrova: this isn't utilizing the objects that DonorCollection is holding - instead you're creating a new object.

students/Ben_freeman/lesson09/Mailroom_OO/mailroom_OO.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def sorting_function(self):
+        """does what the name suggests, it sorts our dictionary of donors"""
+        temp_list = []
+        for keys, values in self.donors.items():
+            temp_list.append([sum(values), keys, len(values), sum(values) / len(values)])
```
natasha-aleksandrova: Donor information like donations avg, total and number of should be part of the Donor object

students/Ben_freeman/lesson09/Mailroom_OO/Cli.py
```diff
@@ -0,0 +1,86 @@
...
+    if answer == "YES":
+        while True:
+            file_name = input("What is the filename?\n")
+            if file_name.upper() == "QUIT":
+                break
+            data = mailroom_OO.FileHandling(file_name).open_file()
+            if data == "file not found":
+                print(data,", please try again")
+            else:
+                donor_list.update(eval(data))
```
natasha-aleksandrova: don't use `eval` 🙂 - this is bad practice

students/Ben_freeman/lesson09/Mailroom_OO/Cli.py
```diff
@@ -0,0 +1,86 @@
...
+    "A": send_thanks,
+    "B": create_report,
+    "C": mass_send_thanks,
+    "D": save_file_prompt,
+    "E": quit_now
+}
+
+# Script
+if __name__ == '__main__':
+    donor_list={}
```
natasha-aleksandrova: confusing name.. it's not a list right?

natasha-aleksandrova:
merging but this needs some more work.

Freemabe:
> @Freemabe good start here but not quite there.
> 
> You should create a **single** instance of DonorCollection class and work with that in your cli module. There will be no more "donor_list" data structure. DonorCollection replaces that. You will feed in some data to initialize it or you can just not initialize it.


Hey, first off thanks so much for all the feedback! So if im understanding correctly instead of calling 

`            else:`
                `if query in donor_list:`
                    `mailroom_OO.DonorCollection(donor_list_variable).update_donor(query, new_donate)`
               ` else:`
                    `mailroom_OO.DonorCollection(donor_list_variable).add_donor(query, new_donate)`

Like I do in lines 32 to 35 of Cli.py I should instead do something like `dummy_variable = mailroom_OO.DonorCollection(donor_list_variable)` and then do 

`dummy_variable.update_donor(query, new_donate)` and like wise with the other places it shows up?
> 
> DonorCollection class should work with Donor objects that are created when donor is created and be held in the collection class.
> 
> If you need more tips or more help, please reach out. These little details are the key to really understanding OO.

In general I don't think I understand this part. Should I be going for something like 
`load in donor data----> use that to create a Donor class object for each donor -----> collect all of them in DonorCollection class-----> Do stuff`

students/Ben_freeman/lesson09/Mailroom_OO/Cli.py
Freemabe: Good to know.

students/Ben_freeman/lesson09/Mailroom_OO/Cli.py
Freemabe: Nope you are correct, that is a confusing name sorry about that.

students/Ben_freeman/lesson09/Mailroom_OO/Cli.py
natasha-aleksandrova: If you'd like to read more about this:
https://nedbatchelder.com/blog/201206/eval_really_is_dangerous.html
http://lybniz2.sourceforge.net/safeeval.html

students/Ben_freeman/lesson09/Mailroom_OO/mailroom_OO.py
Freemabe: should my self.donors attribute be made up of Donor class objects instead of a list? That way line 49 can just read 
`letter.write(str(item))` or something similar?

Freemabe:
Also as a side note I submitted this assignment in the class website yesterday while submitting all of those other exercises. Is it possible for you to unsubmit this one for me? Sorry for any headache this causes.

natasha-aleksandrova:
yes on the dummy_variable example


take a look at this example and let me know if it helps to understand my suggestions on OO:

natasha-aleksandrova:
@Freemabe don't worry about resubmitting Lesson09 on the course website. I will be able to re-adjust your grade once you update the code.

students/Ben_freeman/lesson09/Mailroom_OO/mailroom_OO.py
natasha-aleksandrova: yes! (I sent you a link with an example showing just that pattern).

it would look like `letter.write(donor_obj.get_letter())`

PR #718, opened 2020-02-04
Some tweeks that hopefully are more OO

Freemabe:
Let me know if I'm moving in the right direction.

### DIFF
--- a/students/Ben_freeman/lesson09/Mailroom_OO/Cli.py
+++ b/students/Ben_freeman/lesson09/Mailroom_OO/Cli.py
@@ -1,9 +1,11 @@
 import sys
 import mailroom_OO
+import ast
 
 
-def create_report(donor_list_variable):
-    listy = mailroom_OO.DonorCollection(donor_list_variable).sorting_function()
+def create_report(donors_variable):
+    print(donors_variable.donors)
+    listy = donors_variable.sorting_function()
     title = ["Donor Name", "Total Given", "Num Gifts", "Average Gift"]
     print(f"{title[0]:<20} | {title[1]:>10} | {title[2]:<} | {title[3]:<}")
     print(f"-------------------------------------------------------------")
@@ -11,12 +13,12 @@ def create_report(donor_list_variable):
         print(f"{listy[i][1]:<21} ${listy[i][0]:>12.2f}{listy[i][2]:>11}  ${listy[i][3]:>12.2f}")
 
 
-def send_thanks(donor_list_variable, query="empty_1", new_donate="empty_2"):
+def send_thanks(donors_variable, query="empty_1", new_donate="empty_2"):
     while True:
         if query == "empty_1":
             query = input("Please enter a Full Name, enter 'list' to see a list of current donors, or quit to exit: ")
         if query == "list":
-            print(mailroom_OO.DonorCollection(donor_list_variable).list_donors())
+            print(donors_variable.list_donors())
             break
         elif query.lower() == "quit":
             break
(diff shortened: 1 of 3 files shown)


students/Ben_freeman/lesson09/Mailroom_OO/mailroom_OO.py
```diff
@@ -25,17 +24,21 @@ def donor_donations(self):
...
+        """adds a donation to an existing donors profile"""
+        self.donor.append(new_donate)
 
 class DonorCollection(object):
     """Class to handle all of the methods that have to do with multiple donors at once"""
 
     def __init__(self, donor_list):
-        self.donors = donor_list
+        self.donors = {a.name: a.donor for a in donor_list}
+        self.donors_object = donor_list
```
natasha-aleksandrova: you don't need both self.donors and self.donors_object.

As per my example, just have self.donors which is a dictionary with name as a key and Donor object as a value.

`self.donors = {donor.name: donor for donor in donors_list}`

students/Ben_freeman/lesson09/Mailroom_OO/mailroom_OO.py
```diff
@@ -25,17 +24,21 @@ def donor_donations(self):
...
+    def message(self):
         """our default thank you message"""
-        return f"\nThank you {self.name} for your donation of ${self.donor_sum:.2f}, your donations make the work "\
-            f"of the 'American society for taking donations' possible.\n\n""Sincearly,\n\n""A low paid intern\n "
+        return f"\nThank you {self.name} for your donations totaling ${self.donor_sum:.2f}, your donations make the " \
+            f" work of the 'American society for taking donations' possible.\n\n""Sincearly,\n\n""A low paid intern\n "
 
+    def update_donor(self, new_donate):
+        """adds a donation to an existing donors profile"""
+        self.donor.append(new_donate)
```
natasha-aleksandrova: naming is crucial 🙂 

I didn't notice it last time, but this time reading the code I was lost at what `donor` attribute is.

The attribute name actually should be `donations`. I should be able to understand attributes, functions, classes without having to fully read the code. Just the name alone should tell me a lot about what the attribute is or what the function is expected to do.

students/Ben_freeman/lesson09/Mailroom_OO/Cli.py
```diff
@@ -29,20 +31,30 @@ def send_thanks(donor_list_variable, query="empty_1", new_donate="empty_2"):
...
+                        return print(items.message())
 
 
-def save_file_prompt(chosen_dict):
+
+def data_handler(dictionary):
+    list_of_donors = []
+    for a in dictionary:
+        list_of_donors.append(mailroom_OO.Donor(dictionary[a],a))
+    return list_of_donors
```
natasha-aleksandrova: great place for a list comprehension 😉 

```
return [mailroom_OO.Donor(k, v) for k, v in dictionary.items()]
```

students/Ben_freeman/lesson09/Mailroom_OO/mailroom_OO.py
Freemabe: I'm confused how I can make me sorting_function on lines 54-59 use Donor.attribute things like you suggested last time, if I only use a dictionary that strips away (I think) the Objectness of my Donor class objects. Can I use the input from the __init__ def (specifically donor_list) in other methods in the same class without renaming them like I did before? Thank you.

Freemabe: Mmm, Okay so its a dictionary that identifies all of my donor objects. I think i misread what your sent in your example, so this makes sense now. Cheers

students/Ben_freeman/lesson09/Mailroom_OO/mailroom_OO.py
Freemabe: Will do.

students/Ben_freeman/lesson09/Mailroom_OO/mailroom_OO.py
natasha-aleksandrova: Right - your version is a dictionary with key and values as strings.

My version, keys are name (for easy lookup) and values are Donor objects for you to work with.

PR #719, opened 2020-02-04
Assignments: 3 & 4

tlhuman:
Assignment 3: Kata Fourteen: Tom Swift Under Milk Wood

### DIFF
--- a/students/tim_lurvey/lesson05/mailroom_pt3.py
+++ b/students/tim_lurvey/lesson05/mailroom_pt3.py
@@ -2,7 +2,10 @@
 __author__ = 'Tim Lurvey, ig408c'
 
 import sys
-from lurvey_classes import bunch
+# from lurvey_classes import bunch
+class bunch(object):
+    def __init__(self,**kwargs):
+        setattr(self, '__dict__', kwargs)
 
 my_data = {}
 my_data.update({'Tom Hanks': bunch(total=24536.20, number=3)})
@@ -17,7 +20,7 @@ def add_new_name(name: str = ""):
     return True
 
 
-def compose_email(name: str = "", donation: float = 0., data: any = None):
+def compose_email(name: str = "", donation: float = 0., data: bunch = None):
     """return the string of the formatted email"""
     if data.number < 1:
         s = ""
@@ -47,8 +50,11 @@ def add_donation(name: str = "", ammount: float = 0.):
 
 
 def print_name_list():
-    # I am familiar with one line list comprehension and I know I use it too much
-    # making my code difficult to read, including lambda functions
+    # I am familiar with one-line list comprehension but I know I use it too much
+    # making my code difficult to read, with including lambda functions.  The more
+    # I read and write code, the more I find the need to be verbose.  If speed is
+    # important I will bring the list comprehension into the C level with
+    # one-liners.
     # "\n".join(["The following names are available:",] + [name for name in my_data.keys()])
     print_list = "The following names are available:\n"
     for i, item in enumerate(my_data.items()):
@@ -105,14 +111,15 @@ def report():
     return True
 
 
-def quit():
+def fquit():
     """Exit the application by unseating the switch"""
     print("\nGoodbye! Exiting program...")
     return False
 
-def error(input: str = ""):
+
+def error(inpt: str = ""):
     """Report an error"""
-    print("\nError on input.  Invalid selection \"{}\"".format(input))
+    print("\nError on input.  Invalid selection \"{}\"".format(inpt))
     return True
 
 
@@ -133,9 +140,9 @@ def main(args):
             # keep looping until valid result
             continue
         # Define the logic
-        logic_dict = {'1':send_thank_you,
-                      '2':report,
-                      'q':quit}
+        logic_dict = {'1': send_thank_you,
+                      '2': report,
+                      'q': fquit}
         # Get the corresponding function and execute it
         CONTINUE = logic_dict.get(request, error)()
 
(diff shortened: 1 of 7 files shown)


students/tim_lurvey/lesson05/mailroom_pt3.py
```diff
@@ -0,0 +1,151 @@
...
+def fquit():
+    """Exit the application by unseating the switch"""
+    print("\nGoodbye! Exiting program...")
+    return False
+
+
+def error(inpt: str = ""):
+    """Report an error"""
+    print("\nError on input.  Invalid selection \"{}\"".format(inpt))
+    return True
```
natasha-aleksandrova: all of the `return` statements in your functions are not needed and should be removed.

students/tim_lurvey/lesson05/mailroom_pt3.py
```diff
@@ -0,0 +1,151 @@
...
+                                                                  avg=average, )
+    s += ("-" * 66) + "\n"
+    print(s)
+    return True
+
+
+def fquit():
+    """Exit the application by unseating the switch"""
+    print("\nGoodbye! Exiting program...")
+    return False
```
natasha-aleksandrova: you should call `sys.exit()` in this function to terminate your script and loop

students/tim_lurvey/lesson05/mailroom_pt3.py
```diff
@@ -0,0 +1,151 @@
...
+    return True
+
+
+def main(args):
+    """This is the controlling logic for the program.  The main loop will
+    repeat forever until the user specifies to quit."""
+    # this is the switch variable
+    CONTINUE = True
+    # while switch is true, loop
+    while CONTINUE:
```
natasha-aleksandrova: this should just be a while True loop and use all caps implies a constant (which is isn't)

students/tim_lurvey/lesson05/mailroom_pt3.py
```diff
@@ -0,0 +1,151 @@
...
+        request = input("Select:\n{}".format(msg)).strip()
+        # Error check
+        if len(request) > 1:
+            print("\nplease input only '1' or '2' or 'q'\n")
+            # keep looping until valid result
+            continue
+        # Define the logic
+        logic_dict = {'1': send_thank_you,
+                      '2': report,
+                      'q': fquit}
```
natasha-aleksandrova: ideally this should be defined outside of the loop.

also, missing option 3?

students/tim_lurvey/lesson05/mailroom_pt3.py
```diff
@@ -0,0 +1,151 @@
...
+        # Get the corresponding function and execute it
+        CONTINUE = logic_dict.get(request, error)()
```
natasha-aleksandrova: since this lesson is about exception handling - can you apply exception handling here?

students/tim_lurvey/lesson05/mailroom_pt3.py
```diff
@@ -0,0 +1,151 @@
...
+    while CONTINUE:
+        # Define main input
+        msg = """1 : Send a Thank You\n2 : Create a Report\nq : quit\n>>> """
+        # Query
+        request = input("Select:\n{}".format(msg)).strip()
+        # Error check
+        if len(request) > 1:
+            print("\nplease input only '1' or '2' or 'q'\n")
+            # keep looping until valid result
+            continue
```
natasha-aleksandrova: you don't need special error message logic -- if you apply exception handling below (hint: keyerror) then you can catch invalid option and print error message there.

students/tim_lurvey/lesson05/mailroom_pt3.py
```diff
@@ -0,0 +1,151 @@
...
+    for i, data in enumerate(my_data.items()):
+        total = "$ {:.2f}".format(data[1].total)
+        average = "$ {:.2f}".format(data[1].total / data[1].number)
+        s += "{name:<26}|{tot:>13}|{num:>11d}|{avg:>13}\n".format(name=data[0],
+                                                                  tot=total,
+                                                                  num=data[1].number,
+                                                                  avg=average, )
+    s += ("-" * 66) + "\n"
+    print(s)
+    return True
```
natasha-aleksandrova: missing sorting feature here

students/tim_lurvey/lesson05/mailroom_pt3.py
```diff
@@ -0,0 +1,151 @@
...
+                                       total=data.total,
+                                       is_are=is_are)
+    return email_str
+
+
+def add_donation(name: str = "", ammount: float = 0.):
+    data = my_data.get(name)
+    data.total += float(ammount)
+    data.number += 1
+    my_data.update({name: data})
```
natasha-aleksandrova: no need to rewrite the key value - your object update is enough.

students/tim_lurvey/lesson05/mailroom_pt3.py
```diff
@@ -0,0 +1,151 @@
...
+    return True
+
+
+def print_name_list():
+    # I am familiar with one-line list comprehension but I know I use it too much
+    # making my code difficult to read, with including lambda functions.  The more
+    # I read and write code, the more I find the need to be verbose.  If speed is
+    # important I will bring the list comprehension into the C level with
+    # one-liners.
+    # "\n".join(["The following names are available:",] + [name for name in my_data.keys()])
```
natasha-aleksandrova: often times one-liners (comprehensions) actually make code more readable because that's what most python devs are used to.

that line can actually be shorter -> 
`"\n".join(["The following names are available:",] + list(my_data))`

I find ^^ version a lot more readable than the 4 lines of code below - (if you really think about it.. reading 4 lines is harder than reading one concise line), but I do understand what you mean.. some people to tend to get really carried away with "one-liners".

PR #726, opened 2020-02-10
Lesson 07

StrafingRun:
Lesson 07 ready for review.


StrafingRun:
Lesson 08 is ready for review as well.

students/grant_dowell/Lesson_07/html_render.py
```diff
@@ -0,0 +1,124 @@
...
+"""
+A class-based system for rendering html.
+"""
+
+
+# This is the framework for the base class
+class Element(object):
+
+    tag = "html"
+    tag_attributes = dict()
```
natasha-aleksandrova: this should not be a class variable --- if you are setting it via `__init__` method then it's an instance attribute.

students/grant_dowell/Lesson_08/circle.py
```diff
@@ -0,0 +1,94 @@
...
+        
+    @property
+    def radius(self):
+        return self._radius
+    @radius.setter
+    def radius(self, new_radius):
+        if new_radius >= 0:
+            self._radius = new_radius
+        else:
+            raise ValueError
```
natasha-aleksandrova: this is fine and works, here's a more common pattern:
```
if new_radius < 0:
    raise ValueError

self._radius = new_radius
```

natasha-aleksandrova:
make sure to submit on the course website as well

PR #729, opened 2020-02-11
Grid Printer Exercise

JavaRod:
This is my submission for the Grid Printer exercise.


students/rod_musser/lesson01/front3.py
```diff
@@ -0,0 +1,3 @@
+def front3(str):
+    shortStr = str[0:3]
```
natasha-aleksandrova: make sure to stay consistent with python naming conventions, this should be snake_case

PR #732, opened 2020-02-13
Lesson 3 exercises

calamity-Jack:
Submitting the following lesson 3 exercise lab files for review: 
list_lab.py
slicing_lab.py
strformat_lab.py

### DIFF
--- a/students/jack_anderson/mailroom/mailroom.py
+++ b/students/jack_anderson/mailroom/mailroom.py
@@ -9,7 +9,7 @@ Mailroom assignment part 1
 
 # The list of donors and their amounts donated
 donors_list = [
-    ['Bubbles Trailer',[1500.23, 2523.12, 3012.89]],
+    ['Bubbles Trailer',[1500, 2523, 3012]],
     ['Julien Park',[2520.92, 8623.12]],
     ['Ricky Boys',[7042.76, 3845.56, 5123.25]],
     ['Jack Anderson',[1044, 2232, 4123.56]],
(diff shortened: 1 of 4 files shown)


students/jack_anderson/mailroom/mailroom.py
```diff
@@ -0,0 +1,188 @@
...
+def total_donated(x):
+    """
+    Action to add up the donation amount for each donor
+    :param x: Donors donation index number
+    :return: The sum of items in donor_index[1]
+    """
+    y = 0
+    for i in x:
+        y += float(i)
+    return(y)
```
natasha-aleksandrova: use sum() here

students/jack_anderson/mailroom/mailroom.py
```diff
@@ -0,0 +1,188 @@
...
+    :return: The index number of Name parameter
+    """
+    count = len(donor_names())
+    current = donor_names()
+    start = 0
+    while start < count:
+        if x == current[start]:
+            return(start)
+        start = start + 1
+    new_donor = [x,[]]
```
natasha-aleksandrova: this can be simplified with a for loop and is preferred:
```
for donor in donors_list:
    if donor[0] == x:
      return donor
  else:
    new_donor = [x, []]
    donors_list.append(new_donor)
    return new_donor
```

students/jack_anderson/mailroom/mailroom.py
```diff
@@ -0,0 +1,188 @@
...
+        elif action == 2:
+            create_report()
+
+        elif action == 3:
+            print("Quitting program ....")
+            exit()
+
+    else:
+        print("Please enter a numerical value")
+        start()
```
natasha-aleksandrova: this should be a while True loop -- no other functions should be calling your start() main function. Please review:
https://uwpce-pythoncert.github.io/PythonCertDevel/exercises/mailroom_tutorial.html#controlling-main-program-flow

PR #733, opened 2020-02-14
Assignment 6

tlhuman:
I wasn't able to fully follow the steps laid out within the website so I made my best attempt.  Everything appears to work and the result renders in the browser correctly.  I had to change the indent test so they would pass.

### DIFF
--- a/students/tim_lurvey/lesson07/html_render.py
+++ b/students/tim_lurvey/lesson07/html_render.py
@@ -1,0 +1,153 @@
+#!/usr/bin/env python3
+
+"""
+A class-based system for rendering html.
+"""
+
+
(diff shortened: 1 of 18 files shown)


students/tim_lurvey/lesson07/html_render.py
```diff
@@ -0,0 +1,204 @@
...
+A class-based system for rendering html.
+"""
+
+
+# This is the framework for the base class
+class Element(object):
+
+    def __init__(self, content: str = "") -> None:
+        self.indent = False
+        self.tags = ["<html>", "</html>"]
```
natasha-aleksandrova: the "html" should have been a class attribute (same with indent)

PR #735, opened 2020-02-16
Exceptions

ikempcode33:
Exceptions lab is finished!


ikempcode33:
Questions for mailroom part 3:
I was able to make some try and exceptions that worked well. However, in terms of comprehension, I was having difficulty finding a use for them. Underneath the thank_you() function I put a comprehension that did include a print, even though it was instructed not to, and actually turned out to work well, and not output any None's. This worked for both cases, but for the top, I commented it out and tried making a different comprehension, which works but the output is in a list and not as pretty as if I were just to make a 1 line print with for loop comprehension. Please advise on how you would handle these best, and where else in my code I could have put comprehensions. I was confused on how to make a comprehension out of the for loop in the function report(), for example, because there are calculations being made within the for loop. 

On a separate topic, related to exceptions how do we best determine what exceptions are possible/can rise from a given python command?

students/Isabella_Kemp/lesson05/mailroom_part3.py
```diff
@@ -0,0 +1,178 @@
...
+        "3": letters,
+        "4": exit_menu,
+    }
+
+    while True:
+        try:
+            num = get_user_input(main_prompt)
+            switch_func_dict.get(num)()
+        except TypeError:
+            print("Main menu input is invalid, enter 1 2 3 4 only please")
```
charlesrobison: Good

charlesrobison:
@ikempcode33 - note your trying to commit multiple .DS_Store files - these are Mac OS files that should not be committed.  This may be happening if you're committing all files with a star...?  I always name each file explicitly when committing to avoid this type of thing.  You'll want to remove those from this PR.

ikempcode33:
Thank you I did not catch that, I thought it was working fine, I will fix and resubmit ASAP.

Isabella

ikempcode33:
Fixed it.
For sending letters: the introduction of the try-except was not indented properly so the code was no longer included in the for loop going through the donors. Only the last name in the dictionary which was still formed in the file path ultimately got written, which was simply an indentation error with the try-except because I put them at the same level as the for, as they were before. 

In terms of new donations: the only error was having parentheses around the donation creating a tuple inside the donor data, which was not changeable so when I tried to change the donation a second time it blew up. Now, I use brackets.

For creating a new donor, it was also breaking at the try-except at lines 75-85 because I thought the finally only ran in the case of an exception, as opposed to always running in a try-except.

ikempcode33:
I was able to delete 1 .DS_store but it will not let me get rid of the others when I go into the files_changed tab in this PR. Delete is not an option to click on. 

 

 

 

----------------------------------------------------

Isabella Kemp 

Department of Bioengineering

Undergraduate Researcher

University of Washington Class of 2021

Ikemp33@uw.edu

students/Isabella_Kemp/lesson05/mailroom_part3 .py
```diff
@@ -0,0 +1,181 @@
...
+            print(file_path)
+            with open(file_path, 'w') as write_file:
+                letters = ('Dear {0},\n'
+                            'Thank you for your donations totaling '
+                            '$ {1:,}. We very much appreciate it.\n'
+                            'All the best,\n'
+                            '- Izzy (Charity President)').format(first_name, charity)                
+
+                write_file.write(letters)
+                write_file.close()
```
natasha-aleksandrova: you don't need close() with using `with` statement

students/Isabella_Kemp/lesson05/mailroom_part3 .py
```diff
@@ -0,0 +1,181 @@
...
+def get_donation():
+    money = get_user_input("Please enter a donation amount: $ ")
+    amount = float(money)
+    return amount
+
+
+# Send thank you to donor
+def email(name, amount):
+    print("Thank you {}, for your generous donation of ${:.2f} !".format(name, amount))
+    return
```
natasha-aleksandrova: no need for the redundant `return` statement here

students/Isabella_Kemp/lesson05/mailroom_part3 .py
```diff
@@ -0,0 +1,181 @@
...
+# Checks to see if provided name is in our donor list
+# returns donor record if found, else None
+def get_donor(donor_name):
+    record = None
+    try:
+        record = donors[donor_name]
+    except KeyError:
+        record = None
+    finally:
+        return record
```
natasha-aleksandrova: Using exception handling here is a bit an overkill as it makes for a very verbose code. You don't even need this function at all.

It is simply: `donors.get(donor_name)` which will return the donor data or None

students/Isabella_Kemp/lesson05/mailroom_part3 .py
```diff
@@ -0,0 +1,181 @@
...
+            # but it does not return None's and cleans up the lines, so I wanted to try it
+            [print("{0} {1}". format(key, donors[key])) for key in donors]
+        elif name == '4':  # allows them to quit and go back to menu
+            print("Finished processing thanks to donors...\n\n")
+            return
+        else:
+            # User entered a name, so process that
+            record = get_donor(name)
+            # print(record) # for debug comment out
+            if (record != None):
```
natasha-aleksandrova: when comparing to `None` values you should `is not None` (objects like None/True/False use `is`)

students/Isabella_Kemp/lesson05/mailroom_part3 .py
```diff
@@ -0,0 +1,181 @@
...
+                donors[name] = record
+                email(name, donation)
+
+            else:
+                # new donor is not in our list, add the donor to our list
+                # get donation amount, add donation to list, send thank you
+                print("!!! NEW DONOR !!!")
+                donation = get_donation()
+                donors.update({name: [donation]})
+                email(name, donation)
```
natasha-aleksandrova: there is a lot of duplicate code above - note that this can be combined and reduce a lot of lines.
Let me know if you'd like to see specific code for this.

students/Isabella_Kemp/lesson05/mailroom_part3 .py
```diff
@@ -0,0 +1,181 @@
...
+        if name == "list":
+            # comprehension
+            donor_names = [key for key in donors]
+            print(donor_names)
+            # [print(key) for key in donors] For a prettier output, no None's are outputted
+        elif name == "listall":
+            # entire record name with donations
+            # comprehension, I realize this isnt appropriate with a print statement
+            # but it does not return None's and cleans up the lines, so I wanted to try it
+            [print("{0} {1}". format(key, donors[key])) for key in donors]
```
natasha-aleksandrova: This is invalid use of comprehensions - you are storing None values in a list (memory allocation) that is wasted.

When you use comprehensions, the output should be stored in a variable.

students/Isabella_Kemp/lesson05/mailroom_part3 .py
```diff
@@ -0,0 +1,181 @@
...
+# User can enter a new donor, new donation, user and donation is added to list
+# We finish with a thank you
+def thank_you():
+    while True:
+        name = get_user_input("Enter a donor name to send a thank you letter, "
+                              "'list' or 'listall' for list of donors, '4' will exit: ")
+        if name == "list":
+            # comprehension
+            donor_names = [key for key in donors]
+            print(donor_names)
```
natasha-aleksandrova: good practice for using a comprehension - however, in this case it is simply -> `print(list(donors))`

students/Isabella_Kemp/lesson05/mailroom_part3 .py
```diff
@@ -0,0 +1,181 @@
...
+        record = None
+    finally:
+        return record
+
+
+# Get donation amount
+def get_donation():
+    money = get_user_input("Please enter a donation amount: $ ")
+    amount = float(money)
+    return amount
```
natasha-aleksandrova: what if they pass in a string? should you be catching an exception here?

students/Isabella_Kemp/lesson05/mailroom_part3 .py
```diff
@@ -0,0 +1,181 @@
...
+    # For loop going through donors and calculating sum, number of donations, and average donations
+
+    # not sure how to make a comprehension out of this for loop:
+    for donor in sort_donor:
+        total = sum(donor[1])
+        num_donations = len(donor[1])
+        average = total/num_donations
+        print('{:<20} ${:>11,.2f}{:>13}            ${:>11,.2f}'.format(
+            donor[0], total, num_donations, average))
+    return  # return to main menu
```
natasha-aleksandrova: no need for return here

natasha-aleksandrova:
> Questions for mailroom part 3:
> I was able to make some try and exceptions that worked well. However, in terms of comprehension, I was having difficulty finding a use for them. Underneath the thank_you() function I put a comprehension that did include a print, even though it was instructed not to, and actually turned out to work well, and not output any None's. This worked for both cases, but for the top, I commented it out and tried making a different comprehension, which works but the output is in a list and not as pretty as if I were just to make a 1 line print with for loop comprehension. Please advise on how you would handle these best, and where else in my code I could have put comprehensions. I was confused on how to make a comprehension out of the for loop in the function report(), for example, because there are calculations being made within the for loop.
> 
> On a separate topic, related to exceptions how do we best determine what exceptions are possible/can rise from a given python command?

the print() with comprehensions allocates an array of None values in memory, even though you can't see anything when you run your program:
```
>>> l = list(range(3))
>>> l
[0, 1, 2]
>>> [print(i) for i in l]
0
1
2
[None, None, None]
>>>
```

As far as your question on how do you know which exceptions to catch? it comes from understanding each library you use, how the methods you use work and what exceptions they may throw by reading their documentation or experimenting with the inputs - which is commonly done via unit tests.

natasha-aleksandrova:
@ikempcode33 you should be deleting the .DS_Store files via command line. 

It will look something like this:
```
git rm <path_to_DS_Store_file>
```
then run your usual git commit && git push commands.

Yes you should not be adding git files via `git .` and instead specifying each file path.

ikempcode33:
Thank you for the helpful comments! I edited them all, and eliminated the redundancy in lines 51-68 but let me know if there is a better way to do it. Also, I cannot find how to find the path on the mac that leads to the .DS file, is there a way to find the exact path so I can remove it in the terminal? otherwise, I can try to face call and get help.

natasha-aleksandrova:
@ikempcode33 sounds great! I will take another look.

natasha-aleksandrova:
Screen Shot 2020-02-18 at 10 14 04 AM


The highlighted portion is the path to the file, so in the command line, once you're at the root of this repository you can run:
```
git rm students/.DS_Store
```

natasha-aleksandrova:
@ikempcode33 I see that you removed the .DS_Store that was in your own directory.
You can have any files you want in your own directory.

There is still the file in the student directory. 

You will want to run:
```
git rm students/.DS_Store
```

ikempcode33:
I committed another .DS_store from the student directory, I believe I have removed it. Please tell me if you see it.  

 

 

 

----------------------------------------------------

Isabella Kemp 

Department of Bioengineering

Undergraduate Researcher

University of Washington Class of 2021

Ikemp33@uw.edu

natasha-aleksandrova:
you got it 👍

PR #739, opened 2020-02-18
lesson07: html_render

bromleysj:
I'm having some difficulty with this assignment; I feel I'm close to resolving the issue but could use some help.  I need to move on to lesson08 this week.

### DIFF
--- a/students/scott_bromley/lesson07/html_render.py
+++ b/students/scott_bromley/lesson07/html_render.py
@@ -20,8 +20,11 @@ class Element(object):
         else:
             self.attributes = kwargs
 
-    def append(self, new_content):
-        self.contents.append(new_content)
+    def append(self, content):
+        if hasattr(content, 'render'):
+            self.contents.append(content)
+        else:
+            self.contents.append(TextWrapper(str(content)))
 
     def render(self, out_file, cur_ind=""):
         out_file.write(cur_ind + self.indent)
@@ -138,4 +141,14 @@ class H(OneLineTag):
         self.tag = "h" + str(level)
         super().__init__(contents, **kwargs)
 
+
+class TextWrapper:
+    """
+    Wrapper that creates a class with a render method for simple text
+    """
+    def __init__(self, text):
+        self.text = text
+
+    def render(self, file_out):
+        file_out.write(self.txt)
 
--- a/students/scott_bromley/lesson07/html_render.py
+++ b/students/scott_bromley/lesson07/html_render.py
@@ -1,0 +1,141 @@
+#!/usr/bin/env python3
+
+"""
+A class-based system for rendering html.
+"""
+
+
+# This is the framework for the base class
+class Element(object):
+    tag = "html"
+    indent = "  "
+
+    def __init__(self, contents=None, **kwargs):
+        if contents is None:
+            self.contents = []
+        else:
+            self.contents = [contents]
+        if kwargs is None:
+            self.attributes = {}
+        else:
+            self.attributes = kwargs
+
+    def append(self, new_content):
+        self.contents.append(new_content)
+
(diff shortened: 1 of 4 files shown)


natasha-aleksandrova:
@bromleysj let me know what specific issues you're having and need help with please.

bromleysj:
Natasha,

My problem is that I'm getting this error when I run python
run_html_render.py:
  File "run_html_render.py", line 46, in <module>
    render_page(page, "test_html_output1.html")
  File "run_html_render.py", line 27, in render_page
    page.render(f)
  File
"/Users/scottbromley/PyCharmProjects/uw_python_210/SP_Online_PY210/students/scott_bromley/lesson07/html_render.py",
line 37, in render
    out_file.write(cur_ind + self.indent + content)
TypeError: can only concatenate str (not "TextWrapper") to str

Additionally, I'm getting this error whenever I run my test_html_render.py:

    def render(self, out_file, cur_ind=""):
        out_file.write(cur_ind + self.indent)
        out_file.write(self._open_tag())
        out_file.write("\n")
        for content in self.contents:
            try:
>               content.render(out_file, cur_ind + self.indent)
E               TypeError: render() takes 2 positional arguments but 3 were
given

html_render.py:35: TypeError

I had all 24 tests passing but the run_html_render.py would never get to
output the fifth file.

students/scott_bromley/lesson07/html_render.py
```diff
@@ -0,0 +1,154 @@
...
+# This is the framework for the base class
+class Element(object):
+    tag = "html"
+    indent = "  "
+
+    def __init__(self, contents=None, **kwargs):
+        if contents is None:
+            self.contents = []
+        else:
+            self.contents = [contents]
```
natasha-aleksandrova: this line allows a string to be stored in contents, instead you'll want to do:
```
if content:
    self.append(content)
```

students/scott_bromley/lesson07/html_render.py
```diff
@@ -0,0 +1,154 @@
...
+
+    def render(self, out_file, cur_ind=""):
+        out_file.write(cur_ind + self.indent)
+        out_file.write(self._open_tag())
+        out_file.write("\n")
+        for content in self.contents:
+            try:
+                content.render(out_file, cur_ind + self.indent)
+            except (TypeError, AttributeError):
+                out_file.write(cur_ind + self.indent + content)
```
natasha-aleksandrova: you don't need the exception handling, just call content.render() always.

natasha-aleksandrova: also, the render() call will call to TextWrapper which only accepts one param - this is why you see the second error.

natasha-aleksandrova:
going to close - go ahead and reopen when you have addressed the issues.

bromleysj:
Natasha,

I have a question about lesson08 exercise: in one of the videos
(Implementing Indexing and Slicing) Charles mentions importing numpy as the
package for working with a 2D array as it makes slicing and indexing less
clunky.

If I want to create the custom SparseArray class to be a SparseArray2D
class, is it recommended I use numpy?

Thanks,

Scott

natasha-aleksandrova:
@bromleysj there is no need to use numpy - in most cases it is not needed 🙂 (including the sparse array)

bromleysj:
So it appears that implementing this to accommodate a 2D array will be
somewhat difficult.  I like the extra challenge but I may need to move on
and get the assignment done for 8 and move on to 9.  I know 03/18 is my due
date.

This is my class __init__:

def __init__(self, sequence):
    self.items = list(sequence)
    self.sparse_array = dict(enumerate([[item for item in group if
item != 0] for group in self.items]))

PR #744, opened 2020-02-20
Merging branches and adding files.

cliff002:


cliff002:
Added lesson02 assignment. Creating a grid printer with python.

PR #745, opened 2020-02-21
Adding files for OO Mailroom assignment

N0vA4:
I wasn't completely sure how to test my display report function from the DonorCollection class.  I am assuming I would need to have it return something and be able to print that later in the CLI to be able to test and run it?


students/N0vA/lesson_09/donor_models.py
```diff
@@ -0,0 +1,74 @@
...
+        """Initialize DonorCollection object"""
+        
+        self._database = dict(database)
+
+    def add_donor(self, donor, donation): # Add donor information for report
+        
+        self._database[donor] = Donor(donor, [donation])
+        print("Database updated")
+
+        return self._database
```
natasha-aleksandrova: shouldn't need to return the attribute here given that it is a class attribute

students/N0vA/lesson_09/donor_models.py
```diff
@@ -0,0 +1,74 @@
...
+        sorted_data = sorted(self._database.values(), key=sort_key, reverse=True)
+        
+        report_rows = []
+        
+        # Print each row to format table
+        for per in sorted_data:
+            report_rows.append(member_row.format(per.name, ' ', 
+                int(sum(per.donations)), ' ', 
+                round(len(per.donations),2), ' ', 
+                round(sum(per.donations) / len(per.donations),2)))
```
natasha-aleksandrova: ideally these 3 attributes should be in Donor class

PR #746, opened 2020-02-21
comprehensions, better except hands, pep8

scotchwsplenda:


students/scotchwsplenda/Lesson_5/MailRoom_3.py
```diff
@@ -26,80 +25,87 @@ def main():
                         3: exit_program, 4: mass_mail}
             swit_dic[choice]()
             break
+        except KeyError:
```
natasha-aleksandrova: you don't want to have a `break` above -- you want your program to continue prompt user until they choose to exit.

the rest of main() function calls need to be removed.

students/scotchwsplenda/Lesson_5/MailRoom_3.py
```diff
@@ -26,80 +25,87 @@ def main():
...
+                    send_thanks(donor_inp, don_amount)
+                else:
+                    new_don = input(
+                        f"{donor_inp}'s an unknown donor, input donation"
+                        " amount: ")
+                    new_don = float(new_don)
+                    donators.update({donor_inp: [new_don]})
+                    send_thanks(donor_inp, new_don)
+            if respondy == 3:
+                mainy()
```
natasha-aleksandrova: as I mentioned before, there should be no separate calls to main() --- you just need to make sure your main loop is a while True loop.

PR #749, opened 2020-02-23
Lesson07 Assignment Partially Completed and Lesson08 Completed Assignment

lmburu-uw:
Lesson07 Assignment Partially Completed. Step 9 was a little difficult for me to solve.

### DIFF
--- a/students/larry_mburu/Lesson07/html_render.py
+++ b/students/larry_mburu/Lesson07/html_render.py
@@ -1,0 +1,18 @@
+#!/usr/bin/env python3
+
+"""
+A class-based system for rendering html.
+"""
+
+
+# This is the framework for the base class
+class Element(object):
+
+    def __init__(self, content=None):
+        pass
+
+    def append(self, new_content):
+        pass
+
+    def render(self, out_file):
+        out_file.write("just something as a place holder...")
--- a/students/larry_mburu/Lesson07/html_render.py
+++ b/students/larry_mburu/Lesson07/html_render.py
@@ -7,12 +7,124 @@ A class-based system for rendering html.
 
 # This is the framework for the base class
 class Element(object):
+    
+    indent = '   '
+    tag = 'html'
 
-    def __init__(self, content=None):
-        pass
+    def __init__(self, content=None, **attrs):
+        if content is not None:
+            self.content = [content]
+        else: 
+            self.content = []
+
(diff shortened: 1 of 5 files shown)


students/larry_mburu/Lesson07/html_render.py
```diff
@@ -0,0 +1,130 @@
...
+        self.compose = f"<{self.tag}>" # self.compose - to build out element
+        if len(self.attributes) == 0:  # zero attrs given. don't change tag structure!
+            return self.compose
+        else:
+            self.compose = f"<{self.tag}" # one or more attrs given, change tag structure.
+            for key in self.attributes: 
+                self.compose += f" {key}=\"{self.attributes[key]}\""
+            self.compose.rstrip()
+            self.compose += ">"
+            return self.compose
```
natasha-aleksandrova: you can condense above a bit --- it is safe to loop over empty dictionary:
```
compose = f"<{self.tag}"
for k, v in self.attributes:
    componse += f' {k}="{v}" '
compose.rstrip()
compose  += ">"
return compose
```

notice no `self` -- this is not an instance attribute, just a local variable.

students/larry_mburu/Lesson07/html_render.py
```diff
@@ -0,0 +1,130 @@
...
+
+    def render(self, outfile): 
+        if self.content:  #self closing tags shouldn't have content
+            raise TypeError
+        self.compose = "<"
+        self.compose += f"{self.tag}"
+        for key in self.attributes:
+            self.compose += f" {key}=\"{self.attributes[key]}\""
+        self.compose += " />"
+        outfile.write(self.compose)
```
natasha-aleksandrova: same here -- you don't need `self` with compose var

students/larry_mburu/Lesson08/circle.py
```diff
@@ -0,0 +1,84 @@
...
+    @classmethod
+    def from_diameter(cls, diameter):
+        return cls(diameter / 2)
+    def __str__(self):
+        return f"Circle with radius: {self.radius:.6f}"
+    def __repr__(self): 
+        return f"Circle({self.radius})"
+    def __add__(self, other): 
+        return Circle(self.radius + other.radius)
+    def __iadd__(self, other):
```
natasha-aleksandrova: each method should be separate by one line space

PR #754, opened 2020-02-27
mailroom.py Assignment, Part 1

cliff002:
mailroom.py prompts the user to choose 3 options. Based on the user's options the user can either view the donor's list, and/or send a thank you email to new donations that are saved to the data source.

### DIFF
--- a/students/clifford_butler/lesson03/mailroom.py
+++ b/students/clifford_butler/lesson03/mailroom.py
@@ -49,14 +49,8 @@ prompt_name = "\n".join(("Type the donors full name or,",
 prompt_amount = "\n".join(("Whats the donation amount?",
                 ">>> "))
 
-prompt_thank_you = "\n".join(("Thank you, {response} for the donation of {response_amount}",
+prompt_thank_you = "\n".join(("Thank you, {name} for the donation of {amount}",
                    ">>> "))
-
-prompt_email = f"\n".join(("Hi, {response}",
-                "",
-                "Thank you for the generous donation of {response_amount}",
-                "",
-                ">>> "))
 
 def get_index(donor_name):
     # Return the index number based on user input
@@ -73,11 +67,6 @@ def add_donor(donor_name):
         if item[0] not in donor_dict:
             donor_dict.append([donor_name])
             break
-        
-def send_email():
-    # Send email to donor thanking them.
-    prompt_email
-    
 
 def view_list():
     # Display the donor list.
@@ -98,7 +87,6 @@ def send_thank_you():
             response_amount = input(prompt_amount)
             x = get_index(response) 
             donor_dict[x].append(float(response_amount))
-            print (donor_dict)
             break
     
     else:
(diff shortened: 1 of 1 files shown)


students/clifford_butler/lesson03/mailroom.py
```diff
@@ -0,0 +1,162 @@
...
+        num_gifts = len(item[1:])
+        avg_gift = (str(round(sum_donations/num_gifts,2)))
+        spaces_two = ((15 - len(str(sum_donations)))*' ')
+        spaces_three = (42*" ")
+        
+        print(item[0], spaces, ('$'), sum_donations, spaces_two, num_gifts, spaces_three, ('$'), avg_gift)
+
+def get_index(donor_name):
+    # Return the index number based on user input
+    for item in (donor_dict):
```
natasha-aleksandrova:
```suggestion
    for item in donor_dict:
```

no need for parens

students/clifford_butler/lesson03/mailroom.py
```diff
@@ -0,0 +1,162 @@
...
+        if item[0] == donor_name:
+            index = donor_dict.index(item)
```
natasha-aleksandrova: if you're looping over your list you already should have index available, instead of using index() which comes at cost

take a look `enumerate()`

students/clifford_butler/lesson03/mailroom.py
```diff
@@ -0,0 +1,162 @@
...
+            
+            return index
+        
+    return None
+
+def add_donor(donor_name):
+    # Add donor name to the data structure.
+    for item in donor_dict:
+        if item[0] not in donor_dict:
```
natasha-aleksandrova: I don't believe this is valid logic as it will never be true since you're comparing a name to a nested list of lists.

I also don't think you need this function at all. The only line that you need is:
`donor_dict.append([donor_name])`

which you can just put inline with the base code.

students/clifford_butler/lesson03/mailroom.py
```diff
@@ -0,0 +1,162 @@
...
+            donor_dict[x].append(float(response_amount))
+            break
+    
+    else:
+        # Add the name to the list
+        add_donor(response)
+        # Request the user for a donation amount.
+        response_amount = input(prompt_amount)
+        x = get_index(response)
+        donor_dict[x].append(float(response_amount))
```
natasha-aleksandrova: there is some duplicate code in the for loop and else block, could it be condensed ?

students/clifford_butler/lesson03/mailroom.py
```diff
@@ -0,0 +1,162 @@
...
+1 and 3 donations each. You can store that data 
+structure in the global namespace.
+The script should prompt the user (you) to choose 
+from a menu of 3 actions: “Send a Thank You”, 
+“Create a Report” or “quit”.
+'''
+
+import sys
+
+donor_dict = [
```
natasha-aleksandrova: wouldn't this be `donor_list` ?

natasha-aleksandrova:
@cliff002 feel free to resubmit via a new PR when you're ready.

PR #757, opened 2020-02-28
Exercise 2.3 FizzBuzz

mmanc125uw:
Exercise 2.3 FizzBuzz


students/mmancini/lesson02/Lesson2GridPrintPart03.py
```diff
@@ -0,0 +1,31 @@
+                    
+def PrintGrid(iNumBlocks, iRowCols):
```
natasha-aleksandrova:
```suggestion
def print_grid(num_blocks, row_cols):
```

python uses snake_case naming style

PR #760, opened 2020-03-02
Lesson 4 Exercises

JavaRod:
Includes code for ungraded exercises and the trigrams assignment.  Code for part 2 of the mailroom assignment not included in this pull request.

### DIFF
--- a/students/rod_musser/lesson04/mailroom.py
+++ b/students/rod_musser/lesson04/mailroom.py
@@ -1,0 +1,90 @@
+#!/usr/bin/env python3
+import sys
+
+donors = {
+    'Carmelo Anthony': [1, 50],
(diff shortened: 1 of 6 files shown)


students/rod_musser/lesson04/mailroom.py
```diff
@@ -0,0 +1,90 @@
...
+        response = input(welcome_prompt)
+        menu_selection = main_menu_dict.get(response)
+        if menu_selection is None:
+            print("Not a valid option!")
+        else:
+            menu_selection()
+
+
+def list_donors():
+    newLine = '\n'
```
natasha-aleksandrova:
```suggestion
    new_line = '\n'
```

should be snake_case

PR #762, opened 2020-03-02
circle complete

tlhuman:


students/tim_lurvey/lesson08/circle.py
```diff
@@ -0,0 +1,134 @@
+#!/usr/bin/env python
+__author__ = 'Timothy Lurvey'
+
+from math import pi, floor
+
+num = (int, float, complex)
+
+
+class Circle(object):
+    _r = 0.
```
natasha-aleksandrova: this should not be a class variable -- class variable are typically static, not dynamic

students/tim_lurvey/lesson09/mailroom_oo/donor_classes.py
```diff
@@ -0,0 +1,146 @@
+#!/usr/bin/env python
+__author__ = 'Timothy Lurvey'
+
+from running_total import RunningTotal
+
+
+class Donor(RunningTotal):
```
natasha-aleksandrova: there isn't a good use case here to have Donor inherit from another class

students/tim_lurvey/lesson09/mailroom_oo/donor_classes.py
```diff
@@ -0,0 +1,146 @@
...
+                n, t, c = self._expand_input_values(d_obj)
+                add_obj = Donor(name=n, total=t, count=c)
+            self._data.append(add_obj)
+
+    def add_new_donor(self, obj) -> None:
+        """add a Donor object or a sequence of data to be converted (name, total, count)
+        Donor('x', 1. ,1 )or ('x', 102.5, 2)
+        In the data input sequence, total and count are optional"""
+        try:
+            assert isinstance(obj, Donor) or isinstance(obj, (list, tuple, set))
```
natasha-aleksandrova: assert statements should not be used in "production" code

tlhuman: What is the most pythonic way to type check objects and raise an error?  I have previously used a "If not isinstance(...): raise TypeError(...)" but the If ... else, to me, is more for flow control, not for type checking.

students/tim_lurvey/lesson09/mailroom_oo/donor_classes.py
tlhuman: Outside of class, I repurposed the runningTotal object in my logging decorator that tracks how many times each method is called and how much time they have accumulated.  It's much better than the dictionary, even if it is overkill.

students/tim_lurvey/lesson09/mailroom_oo/donor_classes.py
natasha-aleksandrova: yes if/raise is the proper way to handle error checking in production code

PR #764, opened 2020-03-03
PR for assignments and exercises from lessons 4 to 7

calamity-Jack:
- Updated lesson03 mailroom program as suggested
- Lesson 04 assignment kata_fourteen.py
- Lesson 05 except_exercise.py and except_test.py
- Lesson 06 mailroom_pt4.py, test_mailroom.py, test_walnut_party.py, walnut_party.py
- Lesson 07 html_render.py, test_html_render.py


students/jack_anderson/lesson06/mailroom_pt4.py
```diff
@@ -0,0 +1,298 @@
...
+    donor = name.title()
+    donors = list_names()
+    for i in donors:
+        if name.lower() == i.lower():
+            donations = donors_dict[donor]
+            total = donated + donations
+            donors_dict[donor] = total
+
+        else:
+            donors_dict[donor] = donated
```
natasha-aleksandrova: this is incorrect use of your dictionary data structure

here is how your logic should look like:
```
if donors_dict.get(donor):
    # update donor data
else:
   # add donor data
```

students/jack_anderson/lesson06/mailroom_pt4.py
```diff
@@ -0,0 +1,298 @@
...
+    report_header()
+    action = [donor_details(name, donations) for name, donations in (sorted(donors_dict.items(), key =
+             lambda x:(sum(x[1]), x[0]), reverse=True))]
+    return action
+
+
+def send_email_to_all():
+    # Action to create Thank you email for ALL persons
+    location = prompt_confirmation()
+    action = [create_file(name, thanks_template(name, donation[0]), location) for name, donation in donors_dict.items()]
```
natasha-aleksandrova: create_file has no return value - why is this  a list comprehension?

students/jack_anderson/lesson07/html_render.py
```diff
@@ -0,0 +1,161 @@
...
+
+    indent = "    "
+
+    tag = "html"
+
+
+    def __init__(self, content=None, **kwargs):
+        self.attribs = dict()
+        for k, v in kwargs.items():
+            self.attribs[k] = v
```
natasha-aleksandrova: you can just assign kwargs to `self.attribs` directly

students/jack_anderson/lesson07/html_render.py
```diff
@@ -0,0 +1,161 @@
...
+
+class A(OneLineTag):
+
+    tag = 'a'
+
+    def __init__(self, link, content):
+        x = link
+        self.attribs = dict()
+        self.attribs['href']= link
+        self.contents = [content]
```
natasha-aleksandrova: you should be passing in just the href via kwargs and calling the super init method here.

something like this:
```
def __init__(self, link, content=None, **kwargs):
    kwargs['href'] = link
    super().__init__(content, **kwargs)
```

students/jack_anderson/lesson07/html_render.py
```diff
@@ -0,0 +1,161 @@
...
+    tag = "ul"
+
+    def __init__(self, **kwargs):
+        self.attribs = dict()
+        self.contents = []
+        for key, value in kwargs.items():
+            self.attribs[key] = value
+
+    def append(self, content):
+        self.contents.append(content)
```
natasha-aleksandrova: there should be no methods defined here, it acts the the same as parent class

students/jack_anderson/lesson07/html_render.py
```diff
@@ -0,0 +1,161 @@
...
+
+
+class Li(Element):
+    tag = "li"
+
+
+class H(OneLineTag):
+
```
natasha-aleksandrova: missing `tag` class attribute here

students/jack_anderson/lesson07/html_render.py
```diff
@@ -0,0 +1,161 @@
...
+    def __init__(self, size, content):
+        try:
+            header_size = int(size)
+        except:
+            raise ValueError
+
+        self.attribs = dict()
+        self.contents = [content]
+        self.tag = f'h{header_size}'
```
natasha-aleksandrova: you should only update `self.tag` with the header size --- then call super constructor

students/jack_anderson/lesson08/assignment/circle_class.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+import math
+
+
+
+
+class Circle(object):
+
+    radius = ()
```
natasha-aleksandrova: radius is not a tuple nor should it be a class attribute

PR #778, opened 2020-03-13
Godwin Assignments

Headieh:
I have a modified:   ../.DS_Store in my git status. I don't want to commit/push that. Is there a way to add it to .gitignore?

### DIFF
--- a/students/headieh_godwin/lesson02/fizz_buzz.py
+++ b/students/headieh_godwin/lesson02/fizz_buzz.py
@@ -1,0 +1,14 @@
+NE = 101
+NS = 1
+
+for num in range(NS,NE): 
+    if num % 3 != 0 and num % 5 != 0:
+        print(num)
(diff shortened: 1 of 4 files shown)


natasha-aleksandrova:
you can add to .giitignore that is in your student directory

students/headieh_godwin/lesson02/fizz_buzz.py
```diff
@@ -0,0 +1,14 @@
...
+NS = 1
+
+for num in range(NS,NE): 
+    if num % 3 != 0 and num % 5 != 0:
+        print(num)
+    if num % 3 == 0 and num % 5 == 0: 
+        print("FizzBuzz") 
+    if num % 3 == 0 and num % 5 != 0:
+        print("Fizz")
+    if num % 3 != 0 and num % 5 == 0:
```
natasha-aleksandrova: in this situation you'll want to use if/elif

PR #780, opened 2020-03-14
PR for assignments 7 and 8 for lesson08 and lesson09

calamity-Jack:
Submitting PR for assignments 7 and 8. Made suggested changes for Circle_class. Completed the mailroom obj orientated assignment.


students/jack_anderson/lesson09/mailroom_oo/donor_models.py
```diff
@@ -0,0 +1,227 @@
...
+class Donor():
+    """
+    The Donor() class handles all functionality required for a single donor
+    """
+
+
+    def __init__(self, name, donation=None):
+        # Set name variable and ensure donation is list type
+        self.name = name.title()
+        if donation == None:
```
natasha-aleksandrova: use `is` with `None` (objects and True/False)

students/jack_anderson/lesson09/mailroom_oo/donor_models.py
```diff
@@ -0,0 +1,227 @@
...
+
+class DonorCollection():
+
+    def __init__(self):
+        # Load donors info from donors.pckl file, if no file exists, create default file.
+        # Set self sorted dict variable
+        try:
+            with open('donors.pckl', 'rb') as f:
+                self.donors_dict = pickle.load(f)
+                f.close()
```
natasha-aleksandrova: you dont need to use close() when you use `with` statement

students/jack_anderson/lesson09/mailroom_oo/donor_models.py
```diff
@@ -0,0 +1,227 @@
...
+        except (FileNotFoundError, AttributeError):
+            self.create_sample_list()
+            self.__init__()
+        self.sorted_dict = sorted(self.donors_dict.items(), key=lambda x: (sum(x[1]), x[0]), reverse=True)
```
natasha-aleksandrova: your DonorCollection class should hold a reference to multiple Donor object (which would be the correct usage of two classes)

right now you're sometimes creating Donor objects in few places

PR #782, opened 2020-03-16
Exercise 3.1 Slicing Lab submission

mokiko81:
This is my completed slicing lab submission. My two questions are:

 Is there a better way to do the third_shuffle(seq)?
 Is there a better way to do exchange_first_last(seq) to handle strings or tuples with a single character/item?

### DIFF
--- a/students/adam_strong/lesson03/slicing.py
+++ b/students/adam_strong/lesson03/slicing.py
@@ -1,0 +1,54 @@
+#Slicing Lab: Exercise 3.1
+
+a_string = "this is a string"
+a_tuple = (2, 54, 13, 12, 5, 32)
+
+print('Function 1: exchange_first_last(seq)')
+print('Function 2: every_other_removed(seq')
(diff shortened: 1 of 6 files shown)


mokiko81:
I just added my list_lab.py to my pull request. For some reason, and I have no idea how this happened, my pull request also has several assignments that were not from my local commits.

students/adam_strong/lesson03/slicing.py
```diff
@@ -0,0 +1,54 @@
...
+    return seq[4:len(seq)-4]
+
+def reverse(seq):
+    return seq[::-1]
+
+def third_shuffle(seq):
+    length = len(seq)
+    first_th = round(length/3)
+    last_th = round(2*length/3)
+    return seq[last_th:] + seq[:first_th] + seq[first_th:last_th]
```
charlesrobison: This is a fine solution - most I see are some variation of this.  Nice use of slicing in this exercise overall!

PR #783, opened 2020-03-16
my head is spinning

scotchwsplenda:
this was really hard. I completely relied on the tutorial. I'm not sure how far I would've gotten without that. Class inheritance seems like a meaty subject. I hope I get better at it. I also watched a lot of outside youtube videos. I feel like maybe an easier interim assignment to get familiar with OOP might've helped me personally since I don't know HTML .


natasha-aleksandrova:
@scotchwsplenda I agree that this can be a challenging assignment for someone not familiar with HTML  - however this is a really great example of inheritance! Glad you were able to use the tutorial and make progress.

PR #799, opened 2020-03-24
lesson09 assignment: cli_main.py, donor_models.py, test_mailroom_oo.py

bromleysj:


students/scott_bromley/lesson09/mailroom_oo/donor_models.py
```diff
@@ -0,0 +1,272 @@
...
+        :param donations: donations
+        """
+        self._donor_name = donor_name
+        self.donor_name = donor_name if donor_name is not None else ""
+        self._donations = []
+        if not donations:
+            self._donations = []
+        elif hasattr(donations, "__iter__"):
+            try:
+                [self.add_donation(donation) for donation in donations]
```
natasha-aleksandrova: this is not a valid use case for a comprehension, the result should be stored in a variable.

this should be just a regular for loop

PR #807, opened 2020-03-28
circle.py

ikempcode33:
circle and circle test works and all of the tests pass.


students/Isabella_Kemp/lesson08/circle.py
```diff
@@ -0,0 +1,81 @@
+# Isabella Kemp
+# Circle Class Exercise
+
+import math
+
```
charlesrobison: 2 blank lines after import statements

PR #816, opened 2020-03-31
lesson06 assignment5 mailroom4

mmanc125uw:


students/mmancini/lesson06/test_mailroom4.py
```diff
@@ -0,0 +1,82 @@
...
+
+def test_suite():
+
+    # run test suite tests
+    print(f"Mailroom Test Suite Started")
+    test_create_report()
+    test_send_thankyou_email()
+    test_all_donors_listed()
+    test_write_letters()
+    print(f"Mailroom Test Suite Completed")
```
natasha-aleksandrova, line 69: you should use `pytest` utility to run and produce results for your tests instead of manually calling test functions.

PR #820, opened 2020-04-04
Lesson 2 Assignment Corrections

krysgu:
Corrected based on comments from Natasha.  Thank you.


krysgu:
Added Lesson 3 Slicing lab.

students/kristoffer_jonson/lesson3/list_lab.py
```diff
@@ -0,0 +1,92 @@
...
+    print(fruit_list)
+
+    #Add another fruit to the beginning of the list using insert() and display the list.
+    additional_fruit_to_add = 'Strawberries'
+    fruit_list.insert(0,additional_fruit_to_add)
+    print(fruit_list)
+
+    fruit_with_p = []
+    for fruit in  fruit_list:
+        if 'p' in fruit.lower():
```
natasha-aleksandrova, line 30: per instruction you need to check if fruit starts with "P" -- you're testing if the word contains "P"

you can either use startswith() function or check 0th element.

PR #827, opened 2020-04-07
trigrams

ikempcode33:
turning in the trigrams assignment I missed, was a little hard to follow the instructions for this one but  I tried


students/Isabella_Kemp/lesson04/trigrams.py
```diff
@@ -0,0 +1,92 @@
+# Isabella Kemp
+# Trigrams assignment
+
+import random
+import re
```
charlesrobison, line 5: re imported but not used

PR #834, opened 2020-04-09
Adding lesson05 work - need some help on mailroom3.py

mokiko81:
Hello,

 I have submitted the lesson05 .pys for review. I feel pretty good about the exception exercise but I need some help with the mailroom3.py assignment:

- I feel like I have added enough exceptions to catch various import errors and to reprompt after an error message (I added two Try/Except blocks with specific error catches in - lines 42-51 and lines 92-95)

- What I don't feel good about are adding any comprehensions, there are a few points where I believe I may be able to add a comprehension. Here is my intuition for lines where I think I may be able to make a list comprehension, I'm just looking to see if these would be good place to try them:

Line 68-69
Line 60-61
Somewhere in the send_letter function

I am getting stuck because many of these utilize a print function which I am cagey about working with alongside the comprehensions. Any help is appreciated on where I should be looking in my code to make these comprehensions.

Thank you!

### DIFF
--- a/students/adam_strong/lesson05/mailroom3.py
+++ b/students/adam_strong/lesson05/mailroom3.py
@@ -1,0 +1,99 @@
+#!/usr/bin/env python3
+# Mailroom2.py - Lesson04 Assignment
+import sys
+import pathlib
+
+# Changed the list into a dict
+donor_db = {"Scrooge McDuck": [8000.00, 70000.00],
+            "Montgomery Burns": [49.53],
+            "Richie Rich": [1000000.00, 500000.00],
+            "Chet Worthington": [200.00, 44387.63, 10200.00],
+            "Silas Skinflint": [0.25, 1.00, 0.43]}
+
+main_prompt = "\n".join(("", "Welcome to the donors list",
+          "Please choose from below options:",
+          "1 - Send a thank you",
+          "2 - Create a report",
+          "3 - Send a letter to all the donors",
+          "4 - Quit",
+          "Type a number to select >>> "))
+
+ty_prompt = "\n".join(("", "Please type the full name of the donor OR",
+        "type 'list' to see a list of donors",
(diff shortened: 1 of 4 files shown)


students/adam_strong/lesson05/mailroom3.py
```diff
@@ -0,0 +1,99 @@
...
+        report()
+    try:
+        amt = input("Please enter the donation amount >>>")
+        if donor_db.get(tyname) == None:
+            donor_db[tyname] = [float(amt)]
+        else:
+            donor_db.get(tyname).append(float(amt))
+        input(ty_message.format(tyname, float(amt)))
+        print('')
+    except ValueError:
```
natasha-aleksandrova, line 50: a good rule of thumb is to try to dial in exception handling to the code that would throw that exception.

so for example you could do:
`amt = float(mat)` and put the exception handling just around that line

students/adam_strong/lesson05/mailroom3.py
```diff
@@ -0,0 +1,99 @@
...
+    print('\nShutting down the program\n')
+    sys.exit()
+
+def thank_you():
+    tyname = input(ty_prompt)
+    if tyname == 'list':
+        report()
+    try:
+        amt = input("Please enter the donation amount >>>")
+        if donor_db.get(tyname) == None:
```
natasha-aleksandrova, line 44: you'll want to use `is` here, `is None` 

`is` is used with object comparisons, None and True/False

natasha-aleksandrova:
@mokiko81 

your exception handling looks great, as far as comprehensions .. since mailroom code is unique to each student, it is possible that you will not find any use of comprehensions.

I agree, there is no good place for a comprehension in your code and using it with print() is actually an antipattern because it will generate None values as part of its return.

PR #837, opened 2020-04-11
Completed Assignment 6

JavaRod:


students/rod_musser/lesson07/html_render.py
```diff
@@ -0,0 +1,130 @@
...
+                    content.render(out_file, curr_ident + self.indent)
+                else:
+                    if not isinstance(self, OneLineTag):
+                        out_file.write(curr_ident + self.indent)
+                    out_file.write(str(content))
+                    if not isinstance(self, OneLineTag) and content == self.content_list[-1]:
+                        out_file.write("\n")
+            if not isinstance(self, OneLineTag):
+                out_file.write(curr_ident)
+            out_file.write("</" + self.tag_name + ">" + "\n")
```
natasha-aleksandrova, line 43: this is your base class - it should have no knowledge of the child classes 

Think about it this way, imagine you have a base car model then you can take that base model and do various changes or upgrades like color, sunroof etc. This is the same pattern, your base class is bare with common functionality then you can take it (think like extend it) and customize as needed.

natasha-aleksandrova:
please resubmit with a new PR when you've made corrections.

PR #839, opened 2020-04-12
Done with Lesson 3

tianxie812:


students/Tianx/Lesson3/mailroom.py
```diff
@@ -0,0 +1,151 @@
...
+    print("Exiting the menu now")
+    return "Exit Menu"
+
+
+main_dispatch = {"1": send_thank_you,
+                 '2': create_report,
+                 '3': quit,
+                 }
+
+if __name__ == '__main__':
```
charlesrobison, line 150: Code works and looks great!  Just missing the sorting function for the report.  Donors should be sorted by total amount given, most to least.

students/Tianx/Lesson4/mailroom2.py
```diff
@@ -0,0 +1,175 @@
...
+    return "Exit Menu"
+
+
+main_dispatch = {"1": send_thank_you,
+                 '2': create_report,
+                 '3': send_all,
+                 '4': quit,
+                 }
+
+if __name__ == '__main__':
```
charlesrobison, line 174: Nice job on this too!  Just needs the sorting function as well.

PR #845, opened 2020-04-15
lesson 5 updated

Headieh:
updated code based on feedback - hoping for it to get regraded

### DIFF
--- a/students/headieh_godwin/lesson05/mailroom.py
+++ b/students/headieh_godwin/lesson05/mailroom.py
@@ -2,64 +2,41 @@
 import inquirer
 import os
 import time
-
-donor1 = {
-    "name": "Karen",
-    "donations": (20,20,100)
-}
-donor2 = {
-    "name": "Susan",
-    "donations": (20)
-}
-donor3 = {
-    "name": "Larry",
-    "donations": (40,50)
-}
-donor4 = {
-    "name": "Curly",
-    "donations": (20.99,20,100)
-}
-donor5 = {
-    "name": "Mo",
-    "donations": (2)
-}
+import sys
+import re
 
 donors = {
-  "donor1": donor1,
-  "donor2": donor2,
-  "donor3": donor3,
-  "donor4": donor4,
-  "donor5": donor5
+  "Karen": (20,20,100),
+  "Susan": (20,),
+  "Larry": (40,50),
+  "Curly": (20.99,20,100),
+  "Mo": (2,)
 }
 
 def menu():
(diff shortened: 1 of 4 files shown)


students/headieh_godwin/lesson05/mailroom.py
```diff
@@ -2,64 +2,41 @@
...
-  "donor1": donor1,
-  "donor2": donor2,
-  "donor3": donor3,
-  "donor4": donor4,
-  "donor5": donor5
+  "Karen": (20,20,100),
+  "Susan": (20,),
+  "Larry": (40,50),
+  "Curly": (20.99,20,100),
+  "Mo": (2,)
```
natasha-aleksandrova, line 13: using a tuple is incorrect here , tuples are immutable and are used for specific use cases

your data here is actually mutable and should be stored in a list.

students/headieh_godwin/lesson05/mailroom.py
```diff
@@ -117,59 +94,54 @@ def valid_money():
...
-                    "donations": float(response2)}
-        donors.update({'donor' + str(len(donors)): new_donor} )
-        email(new_donor['name'], new_donor['donations'], z=0)
-    elif response1 in names:
+        donors.update( {response1 : (response2,)} )
+        email(response1, response2, freq=0)
+    #elif response1 in names:
+    else:
         response2 = valid_money()
         filters1 = names.index(response1)
```
natasha-aleksandrova, line 128: looks like filters1 is not used

students/headieh_godwin/lesson05/mailroom.py
```diff
@@ -117,59 +94,54 @@ def valid_money():
...
-    #names2 = []
-    #for key in donors:
-    #    names2.append(donors[key]['name'])
-
-    #list comprehension
-    names = []
-    names = [donors[key]['name'] for key in donors]
-
+    response1 = valid_name()
+    names = [key for (key,value) in donors.items()]
```
natasha-aleksandrova, line 118: to get a list of names you can just do `list(donors)` and use it together with print > `print(list(donors))`

natasha-aleksandrova:
don't forget to submit on the course website

Headieh:
@natasha-aleksandrova This was already submitted and graded (73%) - I was hoping to get it regraded - there's no way that I see to resubmit it with a different pull request link.

natasha-aleksandrova:
@Headieh your grade has been adjusted!

PR #847, opened 2020-04-17
mailroom

ikempcode33:
tried to make OO mailroom. the thank you mail test doesn't work couldn't figure out why.

### DIFF
--- a/students/Isabella_Kemp/lesson09/cli_main.py
+++ b/students/Isabella_Kemp/lesson09/cli_main.py
@@ -1,0 +1,85 @@
+from donor_models import *
+import sys
+import time
+
+# includes all of the user intereaction functions and main program flow 
+
+donors = {"John": [150080.00, 41.28],
+          "Irene": [1600.00, 24.47],
+          "Rob": [19000.00, 200.47],
+          "Kathy": [819.00, 34.5],
+          "Laureen": [830.00, 47.00, 982.13],
+          "Miles": [24.50, 87.00, 193.00]}
+data = DonorCollection(**donors)
+# get donation amount
+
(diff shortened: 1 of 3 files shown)


students/Isabella_Kemp/lesson09/test_mailroom_OO.py
```diff
@@ -0,0 +1,36 @@
...
+    d = Donor('Jessica')
+    d.add_donation(300.0, 100)
+    assert d.total_donation == 400.0
+    assert d.num_donations == 2
+    assert d.avg_donation == 200
+
+def test_thank_you_mail():
+    d1 = Donor("Lilly", [200])
+    expected = ('Thank you Lilly, for your generous donation of $200.00 !')
+    assert d1.thank_you_mail() == expected
```
natasha-aleksandrova, line 29: your function requires an input param:
```
thank_you_mail(self, amount):
```

you should supply amount or change the signature of the function.

students/Isabella_Kemp/lesson09/cli_main.py
```diff
@@ -0,0 +1,85 @@
...
+        elif name == "listall":
+            # entire record name with donations
+            for key in donors:
+                print("{0} {1}". format(key, donors[key]))
+        elif name == '4':  # quit and return to main menu
+            print("Finished processing thanks to donors...\n\n")
+            return
+        donation = get_donation()
+        if data._data.get(name):
+            data._data.get(name).add_donation(donation)
```
natasha-aleksandrova, line 47: this should be handled via DonorCollection class, this code shouldn't know internal details of the class

PR #850, opened 2020-04-18
Lesson 3 Correction and Lesson 4 Mailroom

krysgu:
Is how I sorted the dictionary for the report the preferred method?

### DIFF
--- a/students/kristoffer_jonson/lesson4/mailroom.py
+++ b/students/kristoffer_jonson/lesson4/mailroom.py
@@ -1,0 +1,132 @@
+#!/usr/bin/env python3
+
+import sys
+from datetime import date
+today = date.today()
+
+donor_db = {"William": [653772.32, 12.17],
+            "Jeff": [877.33],
+            "Paul": [663.23, 43.87, 1.32],
+            "Mark": [1663.23, 4300.87, 10432.0],
+            "Elon": [234.25, 2764.87, 9783.0],
+            }
+response = ''
+
+def main(donor_db):
+    """
+    Prompts user for navigation of donor database
+    :param donor_db: Database of donations and donors
+    """
+    response = ''
+
+    switch_func_dict = {
+        '1':thank_you,
+        '2':create_report,
+        '3':all_thank_you,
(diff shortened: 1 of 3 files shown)


students/kristoffer_jonson/lesson4/mailroom.py
```diff
@@ -0,0 +1,132 @@
...
+import sys
+from datetime import date
+today = date.today()
+
+donor_db = {"William": [653772.32, 12.17],
+            "Jeff": [877.33],
+            "Paul": [663.23, 43.87, 1.32],
+            "Mark": [1663.23, 4300.87, 10432.0],
+            "Elon": [234.25, 2764.87, 9783.0],
+            }
```
natasha-aleksandrova, line 12: 👍

students/kristoffer_jonson/lesson4/mailroom.py
```diff
@@ -0,0 +1,132 @@
...
+    switch_func_dict = {
+        '1':thank_you,
+        '2':create_report,
+        '3':all_thank_you,
+        '4':exit_program
+    }
+
+    while True:
+        response = initial_prompt()
+        switch_func_dict.get(response,"nothing")(donor_db)
```
natasha-aleksandrova, line 31: your donor_db structure is global so you don't need to pass it into each function

students/kristoffer_jonson/lesson4/mailroom.py
```diff
@@ -0,0 +1,132 @@
...
+
+    if response.lower() == 'list':
+        donor_list()
+    elif response in donors:
+        donation = enter_donation(donors, response)
+        create_card(response,donation)
+    else:
+        donors[response] = []
+        donation = enter_donation(donors, response)
+        create_card(response,donation)
```
natasha-aleksandrova, line 64: you can make this a bit less repetitive:
```
if response.lower() == "list":
    donor_list()
else:
    
    if response not in donors:
        donors[response] = []
    donation = enter_donation(donors, response)
    create_card(response,donation)
```

PR #852, opened 2020-04-20
refactored mailroom, built test_mailroom4 - 84% coverage, some questions

mokiko81:
This was definitely challenging and I learned quit a bit from the lesson06 material I was more unfamiliar with in my prior experience. So far:

 - I refactored my mailroom code quite a bit
 - Have tests that pass covering 84% of the code (leaving some higher level functions)

I have two questions:

1) In order to complete the higher level function tests I could:
   - I could refactor further and create more functions - thank_you() is an example
   - I could mock multiple things in a test function
   - Some other way about it

2) With pytest and the way I structured the test_mailroom4.py file. If I rearranged two of my tests I would get a failure. For example: function test_ty_logic_new() if swapped in order with test_report() would cause test_report() to fail because there would be another key in the dictionary. This is because it is not resetting the instance between function calls I believe. What would be a good solution to this problem. Would it be through builtins?

Thank you!

### DIFF
--- a/students/adam_strong/lesson06/mailroom4.py
+++ b/students/adam_strong/lesson06/mailroom4.py
@@ -1,0 +1,139 @@
+#!/usr/bin/env python3
+# Mailroom4.py - Lesson06 Assignment - Unittesting with Pytest
+
+import sys
+import pathlib
+
+donor_db = {"Scrooge McDuck": [8000.00, 70000.00],
+            "Montgomery Burns": [49.53],
+            "Richie Rich": [1000000.00, 500000.00],
+            "Chet Worthington": [200.00, 44387.63, 10200.00],
+            "Silas Skinflint": [0.25, 1.00, 0.43]}
+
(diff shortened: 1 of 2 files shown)


natasha-aleksandrova:
@mokiko81 these are all great questions!


Keep in mind that as mailroom evolves each lesson, it becomes a better implementation. The best version is really going to be in Lesson 9 using classes and it will be the easiest to test and control data modifications.

With that being said, your test setup isn't ideal and we don't expect 100 percent coverage because there are more advanced testing topics that would need to be covered.

1. yes mocking would be very helpful to get more coverage!
2. ideally you want to have control over the data you're manipulating in your tests, so yes relying on the order of tests isn't good. You can get fancy with pytest fixture (again, more advanced topics) - to do something similar you capture donors data state, run your test and then reset it back:
```
original = mailroom.donors 
update_donor(). # modifies data
assert ...
mailroom.donors = original
```

PR #872, opened 2020-05-03
Complete Circle and Sphere classes

JavaRod:
Created two separate py files for each classes' test.  (i.e. One that tests Circle and one that tests Sphere.)


students/rod_musser/lesson08/circle.py
```diff
@@ -0,0 +1,110 @@
...
+from math import pi
+from functools import total_ordering
+
+"""
+A class representing a circle
+"""
+
+
+@total_ordering
+class Circle(object):
```
natasha-aleksandrova, line 11: note that in python3 you don't need to extend from object, you can simply do `class Circle:`

PR #873, opened 2020-05-04
finished lesson03

mimalvarez:
Finished Lesson03

### DIFF
--- a/students/mimalvarez_pintor/lesson03/list_lab.py
+++ b/students/mimalvarez_pintor/lesson03/list_lab.py
@@ -1,0 +1,64 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+"""
+Created on Sun Apr 26 20:35:32 2020
+
+@author: miriam
+"""
+
+# Series 1
+L = ['Apples', 'Pears', 'Oranges', 'Peaches']
+print(L)
+F = input("Please type in a fruit: ")
+L.append(F)
+print(L)
+
+for num in L:
+    num = int(input("\nPlease enter in a fruit number (1-5): "))
+    print(num, L[num-1])
+    break
+else:
+    print('')
+L = ["Grapes"] + L
+print(L)
+L.insert(0, "Bananas")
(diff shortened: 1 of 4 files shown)


students/mimalvarez_pintor/lesson03/list_lab.py
```diff
@@ -0,0 +1,64 @@
...
+
+@author: miriam
+"""
+
+# Series 1
+L = ['Apples', 'Pears', 'Oranges', 'Peaches']
+print(L)
+F = input("Please type in a fruit: ")
+L.append(F)
+print(L)
```
natasha-aleksandrova, line 14: make sure you to follow pep8 style guidelines for python:
* use lower case snake_case for var names
* use more user friendly variable names


so in this case it would be `fruit_list` and `fruit` for example

students/mimalvarez_pintor/lesson03/mailroom.py
```diff
@@ -0,0 +1,91 @@
...
+        elif response == "2":
+            create_report()
+        elif response == "3":
+            quit_action()
+        else:
+            print("Not a valid option!")
+
+
+if __name__ == "__main__":
+    main()
```
natasha-aleksandrova, line 91: great work ^^^

students/mimalvarez_pintor/lesson03/slicing_lab.py
```diff
@@ -0,0 +1,48 @@
...
+    return(seq[-1:]+seq[1:-1]+seq[:1])
+#print(exchange_first_last(a_string))
+#print(exchange_first_last(a_tuple))
+
+def every_other_rmv(seq):
+    return(seq[0::2])
+#print(every_other_rmv(a_string))
+#print(every_other_rmv(a_tuple))
+
+def FirstLastFourRmv_EveryOther(seq):
```
natasha-aleksandrova, line 19: make sure to use snake_case function naming -> `def first_last_four_remove_every_other`

PR #874, opened 2020-05-04
Adding lesson08 - circles for review. Cleaned up lesson07 with a linter

mokiko81:
Submitting lesson08 assignment and testing module for review. Also fixed up some PEP-8 issues with html_render from lesson07.

### DIFF
--- a/students/adam_strong/lesson08/circle.py
+++ b/students/adam_strong/lesson08/circle.py
@@ -1,0 +1,82 @@
+#!/usr/bin/env python3
+
+"""
+The ultimate circle.. and sphere
+Circle(object) or Sphere(Circle), __init__(self,radius)
+"""
+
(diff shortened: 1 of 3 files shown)


students/adam_strong/lesson08/circle.py
```diff
@@ -0,0 +1,82 @@
...
+
+    @property
+    def radius(self):
+        return self._radius
+    @radius.setter
+    def radius(self, val):
+        if val < 0:
+            raise ValueError('You cannot have a negative radius')
+        else:
+            self._radius = val
```
natasha-aleksandrova, line 28: there is no need for both `self.radius` and `self._radius` since you're using `radius` as a property.

you should also make sure to define `self._radius` in the init method.

PR #879, opened 2020-05-09
Lesson 7

tianxie812:


students/Tianx/Lesson7/run_html_render.py
```diff
@@ -0,0 +1,231 @@
...
+
+head = hr.Head()
+head.append( hr.Meta(charset="UTF-8") )
+head.append(hr.Title("PythonClass = Revision 1087:"))
+
+page.append(head)
+
+body = hr.Body()
+
+body.append( hr.H(2, "PythonClass - Example") )
```
charlesrobison, line 207: Not seeing an H class in your html_render file.

students/Tianx/Lesson7/html_render.py
```diff
@@ -0,0 +1,126 @@
...
+
+class Ul(Element):
+    tag = "ul"
+
+
+class Li(Element):
+    tag = "li"
+
+
+class Header(OneLineTag):
```
charlesrobison, line 118: Is this supposed to be H instead of Header?  Unable to execute the run file without changing this.

PR #880, opened 2020-05-10
Revise Mailroom

mayue3688:


students/Yue_Ma/Lesson03/mailroom.py
```diff
@@ -54,32 +50,33 @@ def send_a_thank_you_note():
...
                 print('Added the name to our database.')
                 print(email(user_name))
-                back_to_main()
-            break
         elif user_send_a_thank_you_note == '2':
-            back_to_main()
             break
         else:
             print("Not a valid option!!!")
+    return main()
```
natasha-aleksandrova, line 63: you don't want to call back to the main here as well, just `return` is sufficient

students/Yue_Ma/Lesson03/mailroom.py
```diff
@@ -89,28 +86,17 @@ def create_a_report():
...
     print(line_title)
     print(line_x)
-    for i in range(0, len(donor_list)):
+    for i in range(0, len(donation_matrix)):
         main_lines = f'| {donation_matrix[i][0]:<20}| {donation_matrix[i][1]:<15.2f} | {donation_matrix[i][2]:<10} ' \
                      f'| {donation_matrix[i][1] / donation_matrix[i][2]:<15.2f}|'
         print(main_lines)
     print(line_1)
-    back_to_main()
+    return main()
```
natasha-aleksandrova, line 94: same here - don't call back to main(), you're already in a while True loop.

students/Yue_Ma/Lesson03/mailroom.py
```diff
@@ -54,32 +50,33 @@ def send_a_thank_you_note():
...
+    print(donor_list_sort)
+
     # calculate total donation
     donation_matrix = []
-    for j in range(0, len(donor_list)):
-        donation_matrix.append([list(donor_list.keys())[j], sum(list(donor_list.values())[j]),
-                                len(list(donor_list.values())[j])])
+    for j in range(0, len(donor_list_sort)):
+        donation_matrix.append([list(donor_list_sort.keys())[j], sum(list(donor_list_sort.values())[j]),
+                                len(list(donor_list_sort.values())[j])])
```
natasha-aleksandrova, line 79: this function looks a bit complicated, not sure why you're generating a dictionary?

a somewhat simpler version may look like this:
```
sorted_list = sorted(donor_list, key=total_donated, reverse=True)
for item in sorted_list:
    print("...")
```
```

PR #883, opened 2020-05-12
Updated lesson08 assignment and mailroom_oo package for review

mokiko81:
Updated the circles.py file to use self._radius for functions inside the class (and instantiated it as well)

Uploaded mailroom_oo package for review:
   - I think I have the right balance of business logic and command interface logic between the 'cli_main.py' and the 'donor_models.py' files
   - 100% of the donor_models.py file is covered in tests
   - Added additional functionality and tests in order to handle things like: Uncapitalized names and negative donation amounts
   - made PEP8 formatting for the files (except for line length - all below 95 characters per line)
   - The main thing I'm unsure about is if I have Donor and DonorCollections classes working with each other in the best way

Thanks!

### DIFF
--- a/students/adam_strong/lesson09/mailroom_oo/donor_models.py
+++ b/students/adam_strong/lesson09/mailroom_oo/donor_models.py
@@ -1,0 +1,95 @@
+#!/usr/bin/env python3
+
+""" Donor classes to be run by cli_main.py
+"""
+
+# Starting data structure of donors:
+
+import copy
+import pathlib
+
+donor_db = {"Scrooge McDuck": [8000.00, 70000.00],
+            "Montgomery Burns": [49.53],
+            "Richie Rich": [1000000.00, 500000.00],
+            "Chet Worthington": [200.00, 44387.63, 10200.00],
+            "Silas Skinflint": [0.25, 1.00, 0.43]}
+
+report_head = '{:20}| {:>15}|{:>15}| {:>15}'.format(
+    'Donor Name', 'Total Given', 'Num Gifts', 'Average Gift')
+
+letter = "\n".join(("", "Dear {},", "",
+                        "    Thank you for your very kind donations totaling ${:.2f}.", "",
+                        "    It will be put to very good use.", "",
+                        "               Sincerely,",
+                        "                  - The team"))
+
+break_line = '-'*72
+
+ty_message = "\n".join(("", "Dear {}",
+                            "Thank you for your generous donation of {:.2f}", ""))
+
+repform = '{:20} ${:>15.2f}  {:>15}  ${:15.2f}'
+
+
+class DonorCollections(object):
+    '''DonorCollections class used to store and manipultate the main dict'''
+
+    donors = copy.deepcopy(donor_db)
+
+    def __init__(self, donors=donors):
+        self.donors = donors
+
+    def add_donor(self, tyname, amt):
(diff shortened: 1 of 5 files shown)


students/adam_strong/lesson09/mailroom_oo/donor_models.py
```diff
@@ -0,0 +1,95 @@
...
+
+    def add_donor(self, tyname, amt):
+        '''Adds a new donor or updates an existing donor, returns thank you letter'''
+        tyname = tyname.title()
+        if self.donors.get(tyname) is None:
+            self.donors[tyname] = [amt]
+        else:
+            self.donors.get(tyname).append(amt)
+        ty_output = ty_message.format(tyname, amt)
+        return ty_output
```
natasha-aleksandrova, line 50: so the idea is to use Donor class as part of your DonorCollection class, which would actually reference Donor instances.

For example, your structure could be a dictionary with key for donor name and value Donor instance.

students/adam_strong/lesson09/mailroom_oo/donor_models.py
```diff
@@ -0,0 +1,95 @@
...
+            don = Donor(key, value)
+            line = don.format_donor()
+            aggregate.append(line)
+        return aggregate
+
+    def send_letter(self):
+        '''Executes the loop to the Donor class to send letters to each donor'''
+        for key, value in sorted(self.donors.items(), key=lambda i: sum(i[1]), reverse=True):
+            don = Donor(key, value)
+            letter = don.create_letter()
```
natasha-aleksandrova, line 65: all of these functions would change to reference existing donor instances instead of creating new ones.

students/adam_strong/lesson09/mailroom_oo/donor_models.py
```diff
@@ -0,0 +1,95 @@
...
+
+class Donor(object):
+    '''Donor class used to access data for a single donor'''
+
+    def __init__(self, name, value):
+        self.name = name
+        self.value = value
+        self.d_times = len(value)
+        self.total = sum(self.value)
+        self.avg = self.total/self.d_times
```
natasha-aleksandrova, line 76: the donor calculations should be properties or functions since they may change for a donor as donations are updated

mokiko81:
Hello Natasha,

  Thank you for looking through my code, I think I know exactly what you are talking about. I basically took a top down approach - instantiating the DonorCollections from the source dictionary and then generating Donor objects as needed. What I really need to do instead is a bottom up approach - generate Donor objects, use them to instantiate the DonorCollections class and work with them that way.
  Is that more or less correct? I wanted to check before I start refactoring.

natasha-aleksandrova:
@mokiko81 I have an example "skeleton" code to show case the structure:

Let me know if know if that helps and if not we can discuss further!

natasha-aleksandrova:
also, @mokiko81 I don't see a submission for Lesson 08 from you. Can you please submit so that you can receive a grade?

mokiko81:
Thank you Natasha for the link. I really appreciate it, I was working in IPython on refactoring code yesterday and in the DonorCollections class and had figured out the flow. However, the way that you have the __init__ for DonorCollections is much much cleaner and less awkward than the implementation I was going to try and create for that data structure!
  I'll let you know if I need further help, thanks again. Also I just submitted the circle.py assignment.

PR #884, opened 2020-05-13
Updated

mayue3688:


students/Yue_Ma/Lesson03/mailroom.py
```diff
@@ -60,38 +60,40 @@ def send_a_thank_you_note():
...
-    # Print the report
     print(line_1)
     print(line_title)
     print(line_x)
-    for i in range(0, len(donation_matrix)):
-        main_lines = f'| {donation_matrix[i][0]:<20}| {donation_matrix[i][1]:<15.2f} | {donation_matrix[i][2]:<10} ' \
-                     f'| {donation_matrix[i][1] / donation_matrix[i][2]:<15.2f}|'
+    for item in donation_matrix_sorted.keys():
+        main_lines = f'| {item:<20}| {donation_matrix_sorted[item]:<15.2f} | {len(donor_list[item]):<10} ' \
+                     f'| {float(donation_matrix_sorted[item]) / len(donor_list[item]):<15.2f}|'
```
natasha-aleksandrova, line 93: here is the simplest approach demonstrated using some sample code:
```
>>> d = {"bob": (100, 200), "jen": (1, 2)}
>>> def _key(data):
...     return sum(data[1])
...
>>> sorted(d.items(), key=_key, reverse=True)
[('bob', (100, 200)), ('jen', (1, 2))]
>>> sorted(d.items(), key=_key, reverse=False)
[('jen', (1, 2)), ('bob', (100, 200))]
```

above code gives you sorted list in a few clear lines of code, then
```
for donor in sorted_donors:
    print("...")
```

in comparison your code right now has two additional loops that are not needed and can make the reader of the code get "lost" when understanding the logic.

PR #885, opened 2020-05-14
Updated Version

mayue3688:
Thank you so much for the note. I revised the part that you indicate. Please let me know if there are other problems!

### DIFF
--- a/students/Yue_Ma/Lesson03/mailroom.py
+++ b/students/Yue_Ma/Lesson03/mailroom.py
@@ -63,18 +63,17 @@ def send_a_thank_you_note():
     return
 
 
+def total(data):
+    return sum(data[1])
+
+
 def create_a_report():
     title = {'Donor Name': ['Total Given', 'Num Gifts', 'Average Gift']}
     value_title = list(title.values())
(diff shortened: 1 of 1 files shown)


students/Yue_Ma/Lesson03/mailroom.py
```diff
@@ -63,18 +63,17 @@ def send_a_thank_you_note():
...
-    donation_matrix = {}
-    for item in donor_list.keys():
-        donation_matrix[item] = sum(list(donor_list[item]))
-
     # sort
-    donation_matrix_raw = sorted(donation_matrix.items(), reverse=True, key=lambda t: t[1])
+    donation_matrix_raw = sorted(donor_list.items(), reverse=True, key=total)
     donation_matrix_sorted = {}
     for i in range(len(donation_matrix_raw)):
         donation_matrix_sorted[donation_matrix_raw[i][0]] = donation_matrix_raw[i][1]
```
natasha-aleksandrova, line 79: if you reference the example I provided last time, you'll see that the data returned from sorted() function call is what you would use for the report on line 90. This extra for loop is not necessary.

PR #886, opened 2020-05-14
Completed list_lab.py

cdelapena:


students/chris_delapena/lesson03/list_lab.py
```diff
@@ -62,7 +63,7 @@
 print("***SERIES 3***")
 #Display list
 print(fruit)
-fruit_3 = fruit
+fruit_3 = fruit_1
 #Go through list asking if user likes that fruit
 for j in fruit_3:
```
charlesrobison, line 68: Note when response is 'no' the next fruit in the list is skipped - can you see why?

students/chris_delapena/lesson03/list_lab.py
```diff
@@ -29,7 +28,9 @@
 print("Adding pineapple to the list")
 add_fruit_3 = 'Pineapple'
 fruit.insert(0, add_fruit_3)
-print(fruit)
+#Saving configuration of list in fruit_1
+fruit_1 = fruit
+print(fruit_1)
 
 #Display all fruits that begin with "P" using for loop
 print("displaying only fruits beginning with a 'P'")
```
charlesrobison, line 36: Using lower() or similar here will catch case sensitive user input

PR #892, opened 2020-05-17
Refactored mailroom_oo assignment- using classes correctly

mokiko81:
This is my refactored mailroom_oo assignment for lesson09. I feel like it is in a really good place now:

- 100% coverage of donor_models.py
- Using Donor classes within the DonorCollections class
- Changed the sum, total donations and average donations into properties in the Donor class for dynamics

Thank you for your help!

### DIFF
--- a/students/adam_strong/lesson09/mailroom_oo/donor_models.py
+++ b/students/adam_strong/lesson09/mailroom_oo/donor_models.py
@@ -5,14 +5,7 @@
 
 # Starting data structure of donors:
 
-import copy
 import pathlib
-
-donor_db = {"Scrooge McDuck": [8000.00, 70000.00],
-            "Montgomery Burns": [49.53],
-            "Richie Rich": [1000000.00, 500000.00],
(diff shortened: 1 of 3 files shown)


students/adam_strong/lesson09/mailroom_oo/donor_models.py
```diff
@@ -93,3 +64,34 @@ def write_letter(self):
...
+        ty_output = ty_message.format(tyname, amt)
+        return ty_output
+
+    def make_report(self):
+        '''Creates and returns the report on the data'''
+        lines = []
+        aggregate = ['\n', report_head, break_line]
+        for donor_ob in self.donors.values():
+            lines.append([donor_ob.format_donor(), donor_ob.total])
+        [aggregate.append(item[0]) for item in sorted(lines, key=lambda i: i[1], reverse=True)]
```
natasha-aleksandrova, line 91: a more correct way would be:
```
aggregate += [item[0] for item in sorted(...)]
```
in your code, it generates None values in a list that are not used (append() returns None)

PR #894, opened 2020-05-18
Exercise in Lesson 2

derizor:
Completed exercise in Lesson 2.
Please take a look at the series.py, the "__name__== "__main__"" didn't work for me.  Also, not sure if it is by design but in the template the entry "assert fibonacci(0)" has a value.  If I understood the challenge correctly, it should not have any output.


students/Deniss_Semcovs/Lesson02/print_grid2.py
```diff
@@ -0,0 +1,8 @@
+ def print_grid(m,n):
+	addl1 = (" -"*(n)+" +")
```
charlesrobison, line 2: Looks like you're using tabs here, make sure you're using spaces or have your text editor convert them for you.  4 spaces per indent.

students/Deniss_Semcovs/Lesson02/print_grid2.py
```diff
@@ -0,0 +1,8 @@
...
+	addl2 = ("  "*(n)+" |")
+	for i in range(m):
+		print("+"+(addl1*m))
+		for i in range(n):
+			print("|"+(addl2*m))
+	print("+"+(addl1*m))
```
charlesrobison, line 8: Works well

students/Deniss_Semcovs/Lesson02/print_grid_n.py
```diff
@@ -0,0 +1,10 @@
+ def print_grid(n):
+	line1 = ("+"+" -"*(n)+" +"+" -"*(n)+" +")
```
charlesrobison, line 2: Indentation issues here

students/Deniss_Semcovs/Lesson02/printing_a_grid.py
```diff
@@ -0,0 +1,32 @@
...
+|         |         |
+|         |         |
+|         |         |
+|         |         |
++ - - - - + - - - - +
+""") 
+
+# or
+
+print("+"+" -"*4+" +"+" -"*4+" +"),("|"+"  "*4+" |"+"  "*4+" |"),("|"+"  "*4+" |"+"  "*4+" |")
```
charlesrobison, line 32: Hmm...don't think this is working as expected.  You may want to take another look.

students/Deniss_Semcovs/Lesson02/series.py
```diff
@@ -0,0 +1,81 @@
...
+			sn = nth
+			nn += 1
+
+			
+if __name__== "__main__":
+	"""Testing code"""
+	# test fibonacci numbers
+	#assert fibonacci(0) == 
+	assert fibonacci(1) == 0
+	assert fibonacci(2) == 1
```
charlesrobison, line 54: Calling the function is returning the entire series, so this doesn't quite work.

derizor:
Hi Charles,
I am not sure what is wrong with series.py. When I run code it works as
expected, but when I add __name__ == "__main__" it returns
"AssertionError"  Also, the template has values that doesn't make sense to
me, for example:
 "assert fibonacci(1) == 1" if value of fibonacci is 1 it has to return 0
by default.  Also, values (0) as in "assert fibonacci(0) == 0" or "assert
lucas(0) == 2" should not have any return since (0) doesn't call out any
returns.  What am i missing?

Thank you,

Deniss

natasha-aleksandrova:
@derizor we would need to see your code to help you with assertion errors - can you send us a message via course email?

PR #899, opened 2020-05-23
Lesson 4 Assignment scripts

zaandrew:
Lesson 4 Assignment scripts

### DIFF
--- a/students/andrey_zkn/lesson04/trigrams.py
+++ b/students/andrey_zkn/lesson04/trigrams.py
@@ -1,0 +1,69 @@
+#!/usr/bin/env python3
+
+import sys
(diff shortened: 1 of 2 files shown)


students/andrey_zkn/lesson04/trigrams.py
```diff
@@ -0,0 +1,69 @@
...
+import sys
+import string
+import re
+import random
+
+
+def read_in_data(filename): 
+    with open(filename, 'r') as f:
+        output = f.read()
+    f.close()
```
natasha-aleksandrova, line 12: note that you don't need to manually call close() when using `with` statement.

PR #901, opened 2020-05-25
Lesson 9 Final Project

tianxie812:


students/Tianx/Lesson9/mailroom_oo/donor_models.py
```diff
@@ -0,0 +1,184 @@
...
+    def add_new_donation(self, donor_name, donation):
+        """
+        Handles new donation, add donation to exiting the donor if present.
+        Creates new donor and add donation if name is not found.
+        :param donor_name, donation
+        """
+        for donor in self.donor_list:
+            if donor_name == donor.name:
+                donor.add_donation(donation)
+                return
```
natasha-aleksandrova, line 138: I think you can use your get_donor() method here instead

PR #909, opened 2020-05-29
made some changes

derizor:
Please take a look at the files series.py and series_01.py.   The code works when I ran it, but __name__ == "__main__" trows errors.  I noticed test "assert fibonacci(0) == 0" which i don't understand, since fibonacci(0) or lucas(0) shouldn't have any output.


natasha-aleksandrova:
please review the instructions for the exercise:
https://uwpce-pythoncert.github.io/PythonCertDevel/exercises/fib_and_lucas.html

> The function should return the nth value in the fibonacci series (starting with zero index).

which means your function should have an output and asserts 

if the assert was throwing an error then perhaps your code logic may be incorrect?

please reach out if you need any help with this assignment.

PR #910, opened 2020-05-30
Assignment 8: Object Oriented Mailroom

stellie82:
- create classes (Donor, DonorCollection) to hold donor information
- update main source code to use classes
- create functions under classes for donor updates
- add testing for functions and classes

### DIFF
--- a/students/stellie/lesson09/donor_models.py
+++ b/students/stellie/lesson09/donor_models.py
@@ -1,0 +1,66 @@
+#!/usr/bin/env python3
+
+# Stella Kim
+# Assignment 8: Object Oriented Mailroom
+
+"""Donor Models"""
+
+
+class Donor(object):
+    def __init__(self, donor_name, donations=None):
+        self.name = donor_name
+        if donations is None:
+            self.donations = []
+        else:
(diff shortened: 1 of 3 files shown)


students/stellie/lesson09/donor_models.py
```diff
@@ -0,0 +1,66 @@
...
+    def create_report(self):
+        def sum_total(donor_record):
+            return(sum(donor_record[1]))  # return donor donations sum in DB
+        donor_stat = []
+        # Sort database by sum amounts in descending order
+        sorted_db = sorted(self.items, key=sum_total, reverse=True)
+        for item in sorted_db:
+            total = Donor(item[0], item[1]).total_donations()
+            count = Donor(item[0], item[1]).donation_count()
+            average = Donor(item[0], item[1]).avg_donations()
```
natasha-aleksandrova, line 63: this is an incorrect approach - see notes below:
* self.items is not needed in your class
* you should be sorting self.data and define a sort_key method that sorts on an attribute
* once you have your sort objects, you would loop over them and work with donor objects instead of recreating them

PR #911, opened 2020-05-31
added lesson 3 Assignment 1, mailroom part1

TacSPx:
Please see my completed lesson 3 Assignment 1, mailroom part1 - mailroom.py

### DIFF
--- a/students/TacSPx/lesson03/mailroom.py
+++ b/students/TacSPx/lesson03/mailroom.py
@@ -1,0 +1,156 @@
+# ---------------------------------------------------------------------------- #
+# Title: mailroom part 1
+# Description: A program that holds a list of donors and amounts they donated.
+#              Prompt the user to choose 3 menu actions; thank you, report, quit.
+#
+# <05/30/2020>, Created Script
+# ---------------------------------------------------------------------------- #
+
+import sys  # imports go at the top of the file
+
+# Data ---------------------------------------------------------------------- #
+# Task 1: Create data
+# Create a data structure that holds a list of your donors and a history of the amounts they have donated.
+# Have least five donors, with between 1 and 3 donations each, store in the global namespace.
+
+donor_info = [("Jim Zorn", [3772.32, 1201.17]),
+            ("Jermaine Kearse", [877.33]),
+            ("Marcus Trufant", [1563.23, 1043.87, 1.32]),
+            ("K.J. Wright", [21663.23, 300.87, 100432.0]),
+            ("Curt Warner", [663.23, 300.87, 10432.0]),
+            ]
+
+# Processing  --------------------------------------------------------------- #
+# Task 2:  Menu choices
+# Prompt the user to choose from a menu of 3 actions: “Send a Thank You”, “Create a Report” or “quit”.
+###############################################
+#   "Menu of Choices"
+###############################################
+
+def main():
+    '''
+    Menu of Choices:  Thank You, Report, Create a Report
+    :return:
+    '''
+
+    while (True):
+        print("""
+        \n**** XYZ Nonprofit Charity ****\n
+        Menu of Options
+        1) Send a "Thank You"
+        2) Create a Report
+        3) Exit Program
+        """)
+        strChoice = str(input("Which option would you like to perform? [1 to 3] - "))
+        print()  # added a new line for spacing looks
+        if (strChoice.strip() == '1'):
+            thank_you()
+        elif (strChoice.strip() == '2'):
+            report()
+        elif (strChoice.strip() == '3'):
+            exit_program()
+        else:
+            print("\nPlease choose only 1, 2, or 3!")
+
+###############################################
+#   #1 Send a "Thank You"
+###############################################
+def thank_you():
+    '''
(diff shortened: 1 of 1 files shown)


students/TacSPx/lesson03/mailroom.py
```diff
@@ -0,0 +1,156 @@
...
+
+    while (True):
+        print("""
+        \n**** XYZ Nonprofit Charity ****\n
+        Menu of Options
+        1) Send a "Thank You"
+        2) Create a Report
+        3) Exit Program
+        """)
+        strChoice = str(input("Which option would you like to perform? [1 to 3] - "))
```
charlesrobison, line 44: You'll want variables (and almost everything else except classes) to be snake_case

students/TacSPx/lesson03/mailroom.py
```diff
@@ -0,0 +1,156 @@
...
+                    print(f"'{donor_name}' is in registry added new donation amount!")
+                    break
+                # if name not in list add new name to donor info list
+                else:
+                    donor_info.append((donor_name, [donor_amount]))
+                    print(f"New donor name '{donor_name}' added to registry!")
+                    # print(donor_info) - for testing
+                    break
+            input("\nPress Enter for Thank You letter:\n")
+            print(f'''Dear {donor_name},
```
charlesrobison, line 94: This isn't summing new donations to existing donations.  Can you see why?

students/TacSPx/lesson03/mailroom.py
```diff
@@ -0,0 +1,156 @@
...
+        print(header)
+
+        #define format row
+        row = "| {Name:<18s} | ${Total:<10.2f} | {NumbGifts:^11d} | ${AverageGifts:<10.2f} |".format
+
+        for i in donor_info:
+            print(row(Name=i[0], Total=(sum(i[1])), NumbGifts=len(i[1]), AverageGifts=sum(i[1], 0) / len(i[1])))
+        print("________________________________________________________________")
+        input("Press Enter to Continue")
+        break
```
charlesrobison, line 135: Missing the sort function - donors should be listed by total donated, in descending order.

TacSPx:
Hi Charles or Natasha - When I post my work on the course site it asks to
include a link to GitHub.  I used my fork link and now I am thinking that
may not be the correct link, could you please clarify this for me before I
post my other assignments?

Here is the link I submitted:
https://github.com/TacSPx/SP_Online_PY210/blob/master/students/TacSPx/lesson02/grid.py

Thanks,
Steve Pietzke

natasha-aleksandrova:
@TacSPx thank you for checking, that's correct it should be a pull request link not the file link.

The reason is, we can only leave comments and feedback on pull request page.

TacSPx:
Hi Charles - I have made the updates you recommended, but I am struggling
with the summing of new donations.  It appears the donors that are
pre-load will sum just fine, but if you add a new name they append but keep
duplicating as new entries.  I have spent at least 6 hours testing and
searching the web trying to solve this today and I am stumped,
unfortunately I really need to move on.  These types of hiccups I make are
costing me a lot of time to research and I only get the weekends to work on
this as I work full time 50 hours a week.  I understand if you can't tell
me why, I keep thinking it has something to do with the data structure
set-up, but I think we fix this in part 2 when we convert to a dictionary.

Also, as I make these updates do I need to create a new pull request or can
I just upload the updated file to the course grading site?

Thanks,
Steve

TacSPx:
Yes - thank you please help me troubleshoot this it was driving me crazy.
I look forward to finding the issue and fixing it, the pull request version
would be fine to test.  I am able to fix the bug in mailroom part 2 when I
change the data structure to a dictionary if that helps with any
troubleshooting.

students/TacSPx/lesson03/mailroom.py
natasha-aleksandrova: The issue here is the for loop above on line 82. Let's step through it.

Let's say your list has ['bob', 'joe', 'kat'] and we would like to add 'natasha'

in the first loop iteration you have two checks (if and else), 

first one `if i[0] == donor_name:` which will be false because 'natasha' isn't in your list and so it will go into `else` and add "natasha" and `break` out of the loop

the second time you run a program with intent to update donations to natasha, it will run through the same logic again.. however ``if i[0] == donor_name:`` will be FALSE _again_ well because 'bob' != 'natasha' so it will go into `else` block and ADD natasha _again_ and `break` out...

you see the issue there?

so how do we fix it? what you really want is an else statement in a different place:
```
for i in donor_info:
    if i[0] == donor_name:
        # append
        break
else:
    # add
```
in above loop, you will continue checking EACH donor name against the one from the user input, _if_ it matches you will break, if it will not match any of the list entries, it will go into the `else` block of the `FOR` loop but only after exhausting all list entries.

natasha-aleksandrova:
@TacSPx see my inline comment above about the the bug - let me know if you need further help plz.

TacSPx:
Wow, I see it now I kept trying to make it a bigger issue and rewrite my
code another way.  Thanks!

natasha-aleksandrova:
you're welcome 🙂

PR #914, opened 2020-06-01
Added Lesson 6 files

markmcduffie:


students/mark_mcduffie/Lesson6/mailroom_test.py
```diff
@@ -0,0 +1,37 @@
...
+    assert os.path.exists(letter1)
+    assert os.path.exists(letter2)
+
+if __name__== "__main__":
+    donors = mailroom.donors()
+    test_donor_list()
+    test_donor_names()
+    test_email()
+    test_final_letters()
+    print("Pass")
```
natasha-aleksandrova, line 37: make sure you're using pytest (or similar) to run your tests - it appears that you're manually calling and running your tests as part of main.

PR #915, opened 2020-06-02
Lesson 02&03 - series.py exercise 

derizor:
Fixed the code to generate correct result.  Nevertheless, unable to pass __name__ == "__main__"


derizor:
In the slicing exercise (line 10), I don't understand what is "the middle third".

derizor:
I see a bug in the list_lab.py line 47-50, but not sure how to fix it.  When the input is "no" the next item in the list is skipped.

PR #919, opened 2020-06-03
Added Exception handling to Mailroom3.py code

zaandrew:
Added 2 exception handling to mailroom3.py code. 

Thank you for your hints.


zaandrew:
Thank you for letting me now about indentation issue in these codes. I will modify it in Lesson 6 assignment.

PR #921, opened 2020-06-07
Lesson02 series.py assignment

luftsmeerflier:
All tests pass.


luftsmeerflier:
Thank you. I have a question about the recursion function in lesson 2:

def fact(n):
    """compute the factorial of the input value, n"""
    if n == 0:
        return 1
    else:
        return n * fact(n-1)


Who do I talk to about this?

PR #935, opened 2020-06-17
Modified both files to pass unit test.

zaandrew:
Modified both files of the Assignment 6 to pass unit test with PyTest package. It looks like it works now. Thank you so much for your help.


natasha-aleksandrova:
grade adjusted

PR #936, opened 2020-06-17
added resubmit 3 assignment 1, mailroom, part1

TacSPx:
Please see my resubmitted lesson 3 assignment 1, mailroom, part1:
mailroom_modified.py

### DIFF
--- a/students/TacSPx/lesson03/mailroom_modified.py
+++ b/students/TacSPx/lesson03/mailroom_modified.py
@@ -1,0 +1,166 @@
+# ---------------------------------------------------------------------------- #
+# Title: mailroom part 1
+# Description: A program that holds a list of donors and amounts they donated.
+#              Prompt the user to choose 3 menu actions; thank you, report, quit.
+#
+# <05/30/2020>, Created Script, mailroom part 1
+# <06/12/2020>, Clean-up to snake case, adding sorting in reporting descending totals, fix indentation error
+# ---------------------------------------------------------------------------- #
+#imports
+import sys
+from collections import defaultdict
+from operator import itemgetter
+# Data ---------------------------------------------------------------------- #
+# Task 1: Create data
+# Create a data structure that holds a list of your donors and a history of the amounts they have donated.
+# Have least five donors, with between 1 and 3 donations each, store in the global namespace.
+
+
+donor_info = [("Jim Zorn", [3772.32, 1201.17]),
+            ("Jermaine Kearse", [877.33]),
+            ("Marcus Trufant", [1563.23, 1043.87, 1.32]),
+            ("K.J. Wright", [21663.23, 300.87, 100432.0]),
+            ("Curt Warner", [663.23, 300.87, 10432.0]),
+            ]
+
+# Processing  --------------------------------------------------------------- #
+# Task 2:  Menu choices
+# Prompt the user to choose from a menu of 3 actions: “Send a Thank You”, “Create a Report” or “quit”.
+###############################################
+#   "Menu of Choices"
+###############################################
+
+def main():
+    '''
+    Menu of Choices:  Thank You, Report, Create a Report
+    :return:
+    '''
+
+    while (True):
+        print("""
+        \n**** XYZ Nonprofit Charity ****\n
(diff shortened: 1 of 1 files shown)


students/TacSPx/lesson03/mailroom_modified.py
```diff
@@ -0,0 +1,166 @@
...
+#   #1 Send a "Thank You"
+###############################################
+def thank_you():
+    '''
+    Thank You:  Will append donations to existing donors, or adds new donors with donations.
+    User can see list of existing donors by typing 'list'.  Once donation is entered script will create
+    custom thank you letter/ email.
+    :return:
+    '''
+    while (True):
```
natasha-aleksandrova, line 68: no need for parens, this is typically written as `while True`

students/TacSPx/lesson03/mailroom_modified.py
```diff
@@ -0,0 +1,166 @@
...
+        print("""
+        \n**** XYZ Nonprofit Charity ****\n
+        Menu of Options
+        1) Send a "Thank You"
+        2) Create a Report
+        3) Exit Program
+        """)
+        str_choice = str(input("Which option would you like to perform? [1 to 3] - "))
+        print()  # added a new line for spacing looks
+        if (str_choice.strip() == '1'):
```
natasha-aleksandrova, line 49: instead of having strip() everywhere, just strip once:
`str_choice = str(input("Which option would you like to perform? [1 to 3] - ")).strip()`

students/TacSPx/lesson03/mailroom_modified.py
```diff
@@ -0,0 +1,166 @@
...
+                if i[0] == donor_name:
+                    i[1].append(donor_amount)
+                    print(f"'{donor_name}' is in registry added new donation amount!")
+                    break
+                # if name not in list add new name to donor info list
+            else:
+                donor_info.append((donor_name, [donor_amount]))
+                print(f"New donor name '{donor_name}' added to registry!")
+                #print(donor_info) #- for testing
+                break
```
natasha-aleksandrova, line 95: you don't need `break` statement here and it actually introduces a bug here


your thank you letter is never shown for new donors

students/TacSPx/lesson03/mailroom_modified.py
```diff
@@ -0,0 +1,166 @@
...
+            print("\nDONOR REGISTRY:")
+            donor_sort = (sorted(donor_info))
+            for donor in donor_sort:
+                donor_lst = (row(Name=donor[0]))
+                print(donor_lst)
+            input("Press Enter to Continue")
+            print()
+            continue
+
+        if (donor_name.lower() != 'list'):
```
natasha-aleksandrova, line 82: this should be an else statement for if statement on line 70 -- you should never have to test the same variable twice, if there are only two choices (either it's a "list" or not)

students/TacSPx/lesson03/mailroom_modified.py
```diff
@@ -0,0 +1,166 @@
...
+    Will exit the program
+    :return:
+    '''
+    while (True):
+        exit_choice = input("Would you like to Exit?\nEnter 'y' or 'n': ")
+        if (exit_choice.lower() == 'y'):
+            print("\nExiting!")
+            sys.exit()
+        elif (exit_choice.lower() == 'n'):
+            main()
```
natasha-aleksandrova, line 161: should not have main() calls inside functions, this should be a simple `return` statement here

students/TacSPx/lesson03/mailroom_modified.py
```diff
@@ -0,0 +1,166 @@
...
+    '''
+    Will exit the program
+    :return:
+    '''
+    while (True):
+        exit_choice = input("Would you like to Exit?\nEnter 'y' or 'n': ")
+        if (exit_choice.lower() == 'y'):
+            print("\nExiting!")
+            sys.exit()
+        elif (exit_choice.lower() == 'n'):
```
natasha-aleksandrova, line 160: these parens are also redundant (and would usually be flagged as such by linters)

PR #938, opened 2020-06-18
Submitting mailroom and list lab

cdelapena:


students/chris_delapena/lesson03/mailroom.py
```diff
@@ -0,0 +1,122 @@
...
+                " Regards, the Club Owners")
+                print('-'*60)
+                return
+        else:
+            print('-'*60)
+            print(f"Adding new donor {donor_name}")
+            print('-'*60)
+            new_donor = (donor_name, [])
+            donor_db.append(new_donor)
+            return
```
natasha-aleksandrova, line 89: you are missing a feature here - you should ask for donation in both case (new or existing donor)

PR #945, opened 2020-06-21
Grid print part 3 updated.

duanez2021:
Lesson 3 List series exercises and slice lab.

### DIFF
--- a/students/duanez2021/lesson03/List_Lab_Series_01.py
+++ b/students/duanez2021/lesson03/List_Lab_Series_01.py
@@ -1,0 +1,76 @@
+#!/usr/bin/env python3
+import sys
+##############################################################
+# 20200615    djm list lab
+#
+# Duane McCollum Python self-paced winter 2020
+#
+# Exercise 3.1: List Lab (graded)
+# https://startlearning.uw.edu/courses/course-v1:UW+PYTHON210+2019_Winter/courseware/
+#
+#############################################################
+
+# Series 1
+
+# Create a list that contains “Apples”, “Pears”, “Oranges” and “Peaches”.
+# Display the list (plain old print() is fine…).
+# Ask the user for another fruit and add it to the end of the list.
+# Display the list.
+# Ask the user for a number and display the number back to the user and the fruit
+# corresponding to that number (on a 1-is-first basis). Remember that Python uses zero-based indexing, so you will need to correct.
+# Add another fruit to the beginning of the list using “+” and display the list.
+# Add another fruit to the beginning of the list using insert() and display the list.
+# Display all the fruits that begin with “P”, using a for loop
+#
+fruit = list(('Apples', 'Pears', 'Oranges' ,'Peaches'))
+
+def add_a_fruit(response, seq):
+    seq.append(response)
+    return seq
+
+def get_a_fruit(i, seq):
+    if int(i) <= 0: i = '1'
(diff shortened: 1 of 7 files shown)


duanez2021:
Lesson 3 exercises for slice lab and list exercises.

students/duanez2021/lesson03/List_Lab_Series_01.py
```diff
@@ -0,0 +1,76 @@
...
+    str_return = i + ' ' + seq[int(i)-1]
+    return str_return
+
+def add_a_fruit_to_top(item, seq):
+    seq = [item] + seq
+    return seq
+
+def add_a_fruit_by_index(i, item, seq):
+    if int(i) < 0: i = '0'
+    else: i = int(i) - 1
```
natasha-aleksandrova, line 42: each statement should be on separate line

students/duanez2021/lesson03/Slice_Lab_01.py
```diff
@@ -0,0 +1,129 @@
...
+        x[len(x)-1] = first_item
+        x = ''.join(x)
+    else:
+        #it's a not string....
+        x = seq[:]
+        last_item = seq[len(seq) - 1]
+        first_item = seq[0]
+        x[0] = last_item
+        x[len(x) - 1] = first_item
+    return (x)
```
natasha-aleksandrova, line 37: this problem should be solved with slicing only, there should be no use of index access or having to check the type

if you only use slicing, it will work for each data type (hint: it's a one liner)

students/duanez2021/lesson03/Slice_Lab_01.py
```diff
@@ -0,0 +1,129 @@
...
+print(s)
+
+# it's not a string
+print(fruit)
+print(swap_first_and_last(fruit))
+print(fruit)
+
+def swap_first_and_last2(seq):
+    x = seq[:]
+    return x[-1:] + x[1:-1] + x[0:1:1]
```
natasha-aleksandrova, line 50: no need to copy the sequence, slicing & `+ ` will create a new copy

students/duanez2021/lesson03/Slice_Lab_01.py
```diff
@@ -0,0 +1,129 @@
...
+
+fruit2=list(('Açaí', 'Ackee', 'Apple', 'Apricot', 'Avocado', 'Banana', 'Bilberry', 'Blackberry', 'Blackcurrant',
+             'Black sapote', 'Blueberry', 'Boysenberry', 'Breadfruit', "Buddha's hand", 'Cactus pear',
+             'Crab apple', 'Currant', 'Cherry', 'Cherimoya', 'Chico fruit', 'Cloudberry', 'Coconut', 'Cranberry',
+             'Damson', 'Date', 'Dragonfruit', 'Durian', 'Elderberry', 'Feijoa', 'Fig', 'Goji berry', 'Gooseberry',
+             'Grape', 'Raisin', 'Grapefruit', 'Guava'))
+
+def remove_first_and_last_four(seq):
+    x = seq[:]
+    return(x[4:len(x)-4])
```
natasha-aleksandrova, line 86: `return` is not a function call but a statement - you don't use parens with statements 

you also don't need to check the length, this can simply be `seq[4:-4]`

this function is supposed to also remove every other

same comment here, no need to copy the sequence

PR #947, opened 2020-06-23
added resubmit lesson 4 assignment 2, mailroom - part2

TacSPx:
Please see my resubmitted lesson 4 assignment 2, mailroom - part2:
mailroompart2.py

### DIFF
--- a/students/TacSPx/lesson04/mailroompart2.py
+++ b/students/TacSPx/lesson04/mailroompart2.py
@@ -1,13 +1,14 @@
 # ---------------------------------------------------------------------------- #
-# Title: mailroom part 1
+# Title: mailroom part 2
 # Description: A program that holds a list of donors and amounts they donated.
 #              Prompt the user to choose 3 menu actions; thank you, report, quit.
 #
 # <05/30/2020>, Created Script, mailroom part 1
 # <06/12/2020>, Clean-up to snake case, adding sorting in reporting descending totals
-# <06/13/2020>, Clean-up to snake case, adding sorting in reporting descending totals, fixed indentation error
+# <06/13/2020>, Fixed indentation error
 # <06/18/2020>, Modified data struct to dicts, updated dict switch for users selections, &
 #               thank you letter to write to file
+# <06/22/2020>, Code clean-up, split off (report) sorting into a new function
 # ---------------------------------------------------------------------------- #
 # imports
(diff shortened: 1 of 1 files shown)


students/TacSPx/lesson04/mailroompart2.py
```diff
@@ -69,38 +64,35 @@ def thank_you():
...
             # if name is in dict append dollar amount to donors bucket
-            for i in donor_info:
-                if i == donor_name:
-                    donor_info[donor_name].append(donor_amount)
-                    print(f"'{donor_name}' is in registry added new donation amount!")
-                    input("\nPress enter to continue to 'Thank You' letter:")
-                    # print(donor_info)  # - for testing
-                    break
-                # if name not in list add new name to donor info list
+            if donor_name in donor_info:
```
natasha-aleksandrova, line 78: since you're checking for a key in dictionary the correct way to check is via get()

so `donor_info.get(donor_name)`

students/TacSPx/lesson04/mailroompart2.py
```diff
@@ -118,46 +110,37 @@ def letter(donor_name, donor_amount):
...
-        # sort dictionary by totals
-        donor_sort = sorted(donor_merge.items(), key=itemgetter(1), reverse=True)
-        # report printing
-        for i in donor_sort:
-            print(row(Name=i[0], Total=i[1][0], NumbGifts=i[1][1], AverageGifts=i[1][2]))
+        # revised and split off sorting process into new report_sorting function
+        call_sorting = report_sorting(donor_info)
+        for i in range(0, len(call_sorting)):
+            print(f"| {call_sorting[i][1]:<18s} | ${call_sorting[i][0]:<10.2f} | {call_sorting[i][2]:^11d}"
+                  f" | ${call_sorting[i][3]:<10.2f} |")
```
natasha-aleksandrova, line 140: the sorting is much simpler than that, see below:
```
def sort_key(item):
    return sum(item[1])


sorted_donors = sorted(donor_info.items(), key=sort_key, reverse=True)
```
items() creates a list of tuples (key + value), sort_key will get an item from a list and you will return the sum of the second item which is the donations.

then simply loop over (you don't need to use range here) the sorted list:
```
for donor, donations in sorted_donor:
    print(...)
```

PR #952, opened 2020-06-24
lesson09

mimalvarez:
lesson09


students/mimalvarez_pintor/lesson09/donor_models.py
```diff
@@ -0,0 +1,83 @@
...
+        for entry in self.donors.values():
+            total = sum(entry.donation_amt)
+            num = len(entry.donation_amt)
+            average = total/num
+            print('{:<20} ${:>13,.2f}{:>13}  ${:>12,.2f}'.format(entry.name,total,num,average))
+        print('')
+  
+    def donor_exists(self, name):
+        #Search for donor name in donor collection.    
+        return any(d for d in self.donors.values() if d.name.upper() == name.upper())
```
natasha-aleksandrova, line 83: since you're using a dictionary and storing by name, you would only ever match to one name - so you can just check for donor via `self.donors.get(name)`

mimalvarez: oohhhhhh! ok! Thank you!

PR #955, opened 2020-06-25
lists_lab.py

luftsmeerflier:
All four methods work.

### DIFF
--- a/students/luftsmeerflier/lesson03/slicing_lab.py
+++ b/students/luftsmeerflier/lesson03/slicing_lab.py
@@ -65,34 +65,3 @@ assert elements_reversed((1,2,3)) == (3,2,1)
 
 print("All tests pass")
 
-
-# with every other item removed
-
-# def main():
-#     print(exchange_first_last("hello world")
-
-# if __name__ == "__main__":
-#     main()
-
-
-
-# switch first and last
-# first = _input[0]
-# last = _input[-1]
-# _list = [char for char in _input]
-# _list[-1] = first
-# _list[0] = last
-
-# exclude first and last four
-# output = [_input[i] for i in range(4, length - 4)]
-# print(''.join(output))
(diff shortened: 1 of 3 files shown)


students/luftsmeerflier/lesson03/slicing_lab.py
```diff
@@ -0,0 +1,67 @@
...
+assert first_and_last_four_removed("helloworld") == "ow"
+assert first_and_last_four_removed((1,2,3,4,5,6,4,3,2,1)) == (5,6) 
+
+# Pass
+assert elements_reversed("hello") == "olleh"
+assert elements_reversed((1,2,3)) == (3,2,1)
+
+
+print("All tests pass")
+
```
natasha-aleksandrova, line 67: for this assignment:
* use only slicing syntax, and do not check any data types - slicing works the same for tuple, string and list
* `with the first 4 and the last 4 items removed, and the nevery other item in the remaning sequence` - you're missing every other feature
* missing thirds function
* your indentation is incorrect, should be exactly 4 spaces, no tabs

luftsmeerflier:
Hello,

>use only slicing syntax, and do not check any data types - slicing works
the same for tuple, string and list
Duly noted. Though for the first method of the drill, I'm not sure how to
"mutate" the python without turning it into a list first, and then
returning it according to its specified type.

>your indentation is incorrect, should be exactly 4 spaces, no tabs
I've checked, and it looks like my tabs are four spaces wide. I am using a
custom PEP-8 linter in my Sublime text since my Linux distro won't allow
non-free software, such as the standard Sublime text PEP-8 plugin.

luftsmeerflier:
I meant, I don't know how to "mutate" a tuple without first turning it into
a list and back into its respective type.

I've just resubmitted a PR.

Thanks,

John

PR #962, opened 2020-06-28
Mailroom Part 1

westobrien:
Lesson 03 assignment.


students/westobrien/Lesson03/mailroom.py
```diff
@@ -0,0 +1,94 @@
...
+    len_name = 0
+    for donor in donor_db[:]:
+        sum_donation = 0
+        num_gifts = 0
+        avg_gift = 0
+        if len(donor[0]) > len_name:
+            len_name = len(donor[0])
+        for donation in donor[1]:
+            sum_donation += donation
+            num_gifts += 1
```
natasha-aleksandrova, line 56: use sum() and len() functions instead of manually deriving values

PR #963, opened 2020-06-29
All of lesson 02 files

imx-lk:
These are all my files for Lesson 02. 
The Grid example has two files.

### DIFF
--- a/students/imx-lk/lesson02/series.py
+++ b/students/imx-lk/lesson02/series.py
@@ -1,0 +1,80 @@
+#!/usr/bin/env python3
+
+"""
+a template for the series assignment
+"""
+
+
+def fibonacci(n):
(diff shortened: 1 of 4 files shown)


students/imx-lk/lesson02/series.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+def fibonacci(n):
+    '''Takes in a number n, returns the nth value of the fibonacci series'''
+    return (sum_series(n))
+
+
+def lucas(n):
+   '''Takes in a number n , and returns the nth value and the optional arguments of 2 and 1 of the Lucas series'''
+   return (sum_series(n,2,1))
```
natasha-aleksandrova, line 15: no need for parens with return statement -> `return sum_series(n, 2, 1)`

PR #968, opened 2020-07-02
Lesson 9 files

markmcduffie:


students/mark_mcduffie/Lesson9/donor_models.py
```diff
@@ -0,0 +1,77 @@
...
+        for names in self.donors:
+            list.append(names)
+        return '\n'.join(list)
+
+    #Writes new files to save donor thank you messages
+    def send_all(self):
+        for donor in self.donors.values():
+            filename = donor.name + '.txt'
+            with open(filename, 'w') as f:
+                f.write(Donor.print_thanks(donor))
```
natasha-aleksandrova, line 56: this is an incorrect usage of the class, `donor` variable is already an object so you would just call the method on the object like: `donor.print_thanks()`

students/mark_mcduffie/Lesson9/donor_models.py
```diff
@@ -0,0 +1,77 @@
...
+    #Calculates report data and stores into sorted list
+    def make_report(self):
+        report = []
+        total = 0
+        for value in self.donors.values():
+            total = sum(value.donation)
+            count = len(value.donation)
+            average = total / count
+            data = (value.name, total, count, average)
+            report.append(data)
```
natasha-aleksandrova, line 67: total, average, and number of donations can be part of the Donor class.

students/mark_mcduffie/Lesson9/donor_models.py
```diff
@@ -0,0 +1,77 @@
...
+        report.sort(key=sort_key, reverse=True)
+        return report
+
+    #Checks if donor is already in database or not
+    def donor_exists(self, name):
+        # Search for donor name in donor collection.
+        return any(d for d in self.donors.values() if d.name.lower() == name.lower())
+
+def sort_key(donations):
+    return donations[1]
```
natasha-aleksandrova, line 77: note that you can sort a list of objects, in which case the sort_key would accept a single donor object.

PR #972, opened 2020-07-04
lesson09

Headieh:
I made a few changes to this. It has been graded already and I've moved on to the next course but I am hoping since I haven't used the full 4 months yet because of summer break you can take one more quick look at it and give me any feedback to make it better, if not that's okay.


students/headieh_godwin/lesson09/donor_models.py
```diff
@@ -70,47 +68,45 @@ def list_donors(self):
...
-    def list_donors(self, donors = []):
+    def list_donors(self, donors):
+        """list of donors"""
         self.donor_obs_list.extend(donors)
 
+    def find_donor(self, new_donor):
+        """find donor in list"""
+        return next((donor for donor in self.donor_obs_list
+                     if(donor._name.lower() == new_donor.lower())
+                     ), None)
```
natasha-aleksandrova, line 79: you could have used a dictionary to store your donor object for easier look ups - where a key is donor name and value is donor object

students/headieh_godwin/lesson09/donor_models.py
```diff
@@ -119,35 +115,115 @@ def write_thanks_all(self, name, total,numb):
...
+        filename = os.path.join(parent, timestr+'report' + '/donors_report.html')
+        f = io.StringIO()
+        page.render(f)
+        html_content = f.getvalue()
+        with open(filename, 'w') as outfile:
+            outfile.write(html_content)
+        if os.path.exists('/Applications/Safari.app'):
+            os.system(f'open /Applications/Safari.app {filename}')
+        else:
+            webbrowser.open(filename) #this did not work for me on safari
```
natasha-aleksandrova, line 229: highly recommend checking out https://pypi.python.org/pypi/tabulate and playing with it


it's a great library with different format options, I always suggest to use ready libraries vs trying to write code from scratch 🙂

PR #978, opened 2020-07-09
added lesson 6, assignment 5, mailroom, part4 python files

TacSPx:
Please see my completed lesson 6, assignment 5, mailroom, part4 python files:
mailroompart4.py
test_mailroom.py

### DIFF
--- a/students/TacSPx/lesson06/mailroompart4.py
+++ b/students/TacSPx/lesson06/mailroompart4.py
@@ -1,0 +1,236 @@
+# ---------------------------------------------------------------------------- #
+# Title: mailroom part 4
+# Description: A program that holds a list of donors and amounts they donated.
+#              Prompt the user to choose 3 menu actions; thank you, report, quit.
+#
+# <05/30/2020>, Created Script, mailroom part 1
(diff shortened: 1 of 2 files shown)


TacSPx:
Hi - Just wanted to check-in with a few questions since I am getting close
to the end of the first course of the Python Cert.  I noticed lesson 7 was
graded before lesson 6, just wanted to verify that I loaded everything
correctly in the grading system.  I am watching my percentage closely to
make sure I will at least hit the 80% to move on to the second course.  In
the progress dashboard I see 64% so far (with 6 not included yet) for a
point total of 118 out of 157.  I assume the points must be weighted since
118/157 = 75% (and not the 64% in the dashboard)?  At this time I have been
hard at work on lesson 8 and I am happy to say it's going a lot better than
7, with that being said could you please confirm that I have until August 8
to get my last two remaining assignments in?

Thanks,
Steve

natasha-aleksandrova:
@TacSPx thank you for reaching - we've been catching since our 2 week break and have been slower on grading, apologies for that.

I have updated your grade.

PR #979, opened 2020-07-10
Pull request for Lessons01-03

s-mar206:
I have added Lessons 01 through Lessons 03 and also added a readme file

### DIFF
--- a/students/Steven/lesson03/mailroom.py
+++ b/students/Steven/lesson03/mailroom.py
@@ -1,0 +1,74 @@
+#! bin/user/env python3
+import sys
+"""Program Part1:
+It should have a data structure that holds a list of your donors and a history of the amounts they have donated.
+This structure should be populated at first with at least five donors, with between 1 and 3 donations each.
+You can store that data structure in the global namespace.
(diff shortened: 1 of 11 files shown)


students/Steven/lesson03/mailroom.py
```diff
@@ -0,0 +1,74 @@
...
+    donor_list.sort(reverse=True, key=sort_list)
+    fields = ["Donor Name", "Total Given", "Num Gifts", "Average Gift"]
+    report_list = []
+    print(f"{fields[0]:<20} | {fields[1]:<11} | {fields[2]:<3} | {fields[3]}")
+    print("{}".format('-' * 61))
+    for item in seq:
+        report_list.append((item[0], item[1], item[2], item[1]/item[2]))
+    for donor in report_list:
+        print(f"{donor[0]:<20} $ {donor[1]:>11.2f} {donor[2]:>11} $ {donor[3]:>11.2f}")
+    main()
```
natasha-aleksandrova, line 66: you want your main entry code to use while True loop, you can check out the mailroom guidelines

then remove all of the manual function calls to main(), there should be no such function calls within your functions.

PR #987, opened 2020-07-17
Correcting according to comments

Pnaghavi:
Just correcting the code according to your kind comments. Please let me know if you see anything else that could make my code look more correct or professional anywhere. Thank you.
Also, I finished lesson 6! 

Best Regards
Pirouz Naghavi


students/Pirouz_N/lesson04/mailroom.py
```diff
@@ -56,11 +55,22 @@ def main_menu():
         try:
             option_dict[option]()
         except KeyError:
-            print('Please only type 1 or 2 or 3 or 4 and hit enter. Any other option will not be acceptable.')
+            print(f"Please choose from following options: {str([int(item) for item in list(option_dict)])[1:-1]}")
```
natasha-aleksandrova, line 58: what you have is fine - if you're curious on a more pythonic version:
```
print(f"Please choose from following options: {' ,'.join(map(str, option_dict))}")
```

`join()` will join a sequence on an str char (in this case comma); it does require to be a string so I am using map() to match each seq item to a string.

students/Pirouz_N/lesson06/test_mailroom.py
```diff
@@ -0,0 +1,247 @@
...
+    mailroom.donor_db = {
+        'William Gates, III': [653772.32, 12.17],
+        'Jeff Bezos': [877.33],
+        'Paul Allen': [663.234, 43.87, 1.32],
+        'Mark Zuckerberg': [1663.23, 4300.87, 10432.0],
+        'Jayz': [999.9, 100.0, 100.0],
+        'Beyonce': [1000.0, 100.0],
+        'Melinda Gates': [10000000.0, 1000000.0, 1000020.0],
+    }
+    mailroom.continue_loop = True
```
natasha-aleksandrova, line 24: I don't believe this variable is in your module anymore

students/Pirouz_N/lesson04/mailroom.py
Pnaghavi: Cool! Thank you. I will update it!

students/Pirouz_N/lesson06/test_mailroom.py
Pnaghavi: Oh true! Thank you I missed that.

PR #995, opened 2020-07-21
Added fiz_buzz.py file

Altaf410:
I understand there are more 'pythonic' ways to go about answering this problem. Which ones would you recommend looking into?

### DIFF
--- a/students/altaf_lakhi/lesson_02/grid_printer.py
+++ b/students/altaf_lakhi/lesson_02/grid_printer.py
@@ -1,0 +1,59 @@
+def grid_printer():
+    plus = ('+ ' + ('- ' * 4)) * 2
+    tally = ('|' + (' ' * 9)) * 3
+    print(plus, end='+')
+    print()
+    print(tally)
+    print(tally)
+    print(tally)
+    print(tally)
+    print(plus, end='+')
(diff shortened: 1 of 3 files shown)


students/altaf_lakhi/lesson_02/series.py
```diff
@@ -0,0 +1,80 @@
...
+    if n == 0:
+        return num[0]
+    elif n == 1:
+        return num[1]
+    else:
+        for i in range(n-1):
+            fib = num[0] + num[1]
+            num.append(fib)
+            num.pop(0)
+        return num[1]
```
natasha-aleksandrova, line 29: you could also use a recursion here

students/altaf_lakhi/lesson_02/grid_printer.py
```diff
@@ -0,0 +1,59 @@
...
+    print(tally)
+    print(tally)
+    print(tally)
+    print(plus, end='+')
+    print()
+    print(tally)
+    print(tally)
+    print(tally)
+    print(tally)
+    print(plus, end='+')
```
natasha-aleksandrova, line 16: you could you use for loops here - notice the repetition 🙂

PR #996, opened 2020-07-21
added mail room 3 and lesson 5 exercises

oieeri26:


students/erik_oien/lesson05/mailroom3.py
```diff
@@ -0,0 +1,165 @@
...
+}
+
+# main menu for program 
+
+def menu():
+    while True:
+        welcome_input = input(opening_promt)
+        try:
+            main_menu_switch[welcome_input]()
+        except(KeyError):
```
natasha-aleksandrova, line 21: nice! but you don't parens here -> `except KeyError`

students/erik_oien/lesson05/mailroom3.py
```diff
@@ -0,0 +1,165 @@
...
+        past_donor = False
+        if donor_name.lower() in ["menu", "quit"]:
+            sub_menu_switch[donor_name.lower()]()
+        elif donor_name.lower() == 'list':
+            past_donor = False
+            for donor in donor_dict.items():
+                print(donor[1]["first_name"] + " " + donor[1]["last_name"])
+        else:
+            try:
+                for donor in donor_dict.items():
```
natasha-aleksandrova, line 37: see my comment about your data structure - if you actually change it to utilize it then you can just do `donor_dict.get(donor_name)`

students/erik_oien/lesson05/mailroom3.py
```diff
@@ -0,0 +1,165 @@
...
+                
+# create report menu
+
+def create_report():
+    while True:
+        donor_report = []
+        [donor_report.append(donor_info(donor)) for donor in donor_dict.items()]
+        donor_report_info = sorted(donor_report, key=sort_key, reverse=True)
+        table_header()
+        [row_formatter(donor) for donor in donor_report_info]
```
natasha-aleksandrova, line 62: even though this lesson is about trying to use comprehensions - you actually want to avoid this pattern.

the issue here is that this return a list of None values, since `row_formatter` has no return value and you're not storing the final result anyway.

using a regular for loop here is a better practice.

students/erik_oien/lesson05/mailroom3.py
```diff
@@ -0,0 +1,165 @@
...
+    else:
+        return(donor_name, donor_sum, donor_length, donor_average)
+
+# thank you letter
+def thank_you_letter(text, dir_path):
+    letter_text = "Dear {},\n\n\tYou have made {} donations to the kittens and metal charity totaling ${}.\n\nSincerely,\n\n\tKitten and Metal Charity \m/".format(*text)
+    file_name = dir_path + "/" + text[0] + ".txt"
+    letter = open(file_name, "w")
+    letter.write(letter_text)
+    letter.close()
```
natasha-aleksandrova, line 133: best practices is to use `with` statement which handles closing the file for you.

PR #1007, opened 2020-07-28
Updated Mailroom part 1 and added Mailroom Part 2 for assignment 04

s-mar206:
Fixed some code in Mailroom Part 1 where I had main function calls manually and I have also uploaded Mailroom Part 2 from assignment 04.

### DIFF
--- a/students/Steven/lesson04/mailroom02.py
+++ b/students/Steven/lesson04/mailroom02.py
@@ -1,0 +1,97 @@
+#! bin/user/env python3
+import sys
+from operator import itemgetter
+"""
+Program Part2:
+Update your mailroom program to:
+
+Use dicts where appropriate.
+See if you can use a dict to switch between the user’s selections.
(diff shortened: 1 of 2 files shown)


students/Steven/lesson04/mailroom02.py
```diff
@@ -0,0 +1,97 @@
...
+
+# Ask for a donor name and donation amount, send an email to that donor once completed
+def thank_you():
+    name = input("Enter a first and last name: ").title()
+    if name.lower() == "list":
+        print(donor_list)
+        thank_you()
+    gift = float(input("Enter a donation amount: "))
+    for donor in donor_list:
+        if name in donor:  # if name is in the dictinoary
```
natasha-aleksandrova, line 56: since you're now dealing with a dictionary you don't want to loop over it, instead you should access dictionary key to check if donor exists

```
if donors.get(donor)
```

PR #1014, opened 2020-07-31
Adding lesson 9 files

cdelapena:


students/chris_delapena/lesson09/donor_models.py
```diff
@@ -0,0 +1,36 @@
...
+    def __init_(self):
+        self.donor_db = []
+
+    def add_donor(self, donor, donation):
+        self.donor_db[donor] = Donor(donor,[donation])
+        return self.donor_db
+
+    @property
+    def donor_list(self):
+        return list(self.donor_db)
```
natasha-aleksandrova, line 36: couple other functions that should live in this class:
* data for report generation
* searching donors - remember that the user of the class should not know that you're actually storing donors data in a dictionary

PR #1015, opened 2020-07-31
Added rest of Lesson 3 exercises

Lina-271:
Please review rest of Lesson 3 - list_lab.py, strformat_lab.py and mailroom.py

### DIFF
--- a/students/Lina/Lesson3/mailroom.py
+++ b/students/Lina/Lesson3/mailroom.py
@@ -1,0 +1,146 @@
+#! python
+
+#------------------------------------------------
+# Lesson 3 - Assignment 1: Mailroom, Part 1
+#------------------------------------------------
+
+import sys
+from operator import itemgetter
+from datetime import date
+
+#This list contains the donor names and a history of their donation amounts.
+donor_db = [("William Gates, III", [653772.32, 12.17]),
+            ("Jeff Bezos", [877.33]),
+            ("Paul Allen", [663.23, 43.87, 1.32]),
+            ("Mark Zuckerberg", [1663.23, 4300.87, 10432.0]),
+            ("Oprah Winfrey", [3312.00, 5500.00]),
+            ("Joe Smith", [250.75, 120.00, 55.50, 100.50])
+            ]
+
+def show_donor_list():
+    #This function will show all donors in the list.
+    for row in donor_db:
+        print(row[0])
+
+def verify_donor(name):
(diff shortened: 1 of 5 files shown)


students/Lina/Lesson3/mailroom.py
```diff
@@ -0,0 +1,146 @@
...
+def verify_donor(name):
+    #This function will check if the donor is in the list. If not there,
+    #it will add to the list.
+    new_donor = True
+    for row in donor_db:
+        if row[0] == name:
+            new_donor = False
+            break
+
+    if new_donor:
```
natasha-aleksandrova, line 34: instead of using a variable you can use an else clause with for loop:
```
for donor in donor_db:
    if donor[0] == name:
        # update
        break
else:
    # goes here if not found
```

students/Lina/Lesson3/mailroom.py
```diff
@@ -0,0 +1,146 @@
...
+        donor_db.append((name, []))
+
+def add_donation(name, amount):
+    #This function will add a donation amount of a donor.
+    for i in range(len(donor_db)):
+        if donor_db[i][0] == name:
+            donor_db[i][1].append(amount)
+            break
```
natasha-aleksandrova, line 42: I would combine verify and this add function together - no need to loop twice over the same data.

also, use the other looping style - where you're directly looping over the list `for donor in donor_db`

Lina-271:
Hi Natasha,

This is regarding my pull request# 1015.  It appeared that my Lesson 4 assignments were included in this same pull request because I was not able to create a new one, not sure if it was because it hasn’t been approved at the time.  I had already submitted my Lesson 4 assignments with pull request #1015, now I got the notification that the pull request has been approved and closed. So I just wonder if my Lesson 4 assignements will still get reviewed.  Sorry for the confusion, please advice.


Thank you,
Lina

Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows 10

natasha-aleksandrova:
@Lina-271 thank you for reaching out - since there can only be one pull request open it does happen sometimes that additional lessons are added while your PR is open and it's ok.

Your lesson04 was reviewed as well and there were no issues.

Lina-271:
Good to know, thank you!

Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows 10

PR #1029, opened 2020-08-06
submitting lesson 7 assignment

kimc05:
submission for lesson 7 assignment 6 for Christine Kim

### DIFF
--- a/students/kimc05/Lesson08/circle.py
+++ b/students/kimc05/Lesson08/circle.py
@@ -1,0 +1,59 @@
+"""
+Christine Kim
+Lesson 8
+Assignment Circles
+"""
+import math
+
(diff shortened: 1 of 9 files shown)


students/kimc05/Lesson08/circle.py
```diff
@@ -0,0 +1,59 @@
...
+    @diameter.setter
+    def diameter(self, diameter):
+        self.radius = diameter / 2
+
+    @property
+    def area(self):
+        return math.pi * (self.radius ** 2)
+
+    @classmethod
+    def from_diameter(c, diameter):
```
natasha-aleksandrova, line 57: the convention is to use `cls` for the first param variable name (short for class)

PR #1030, opened 2020-08-07
Assignments for Lesson 5 and 6

Lina-271:


students/Lina/Lesson5/mailroom3.py
```diff
@@ -0,0 +1,176 @@
...
+        else:
+            continue
+
+    while True:
+        amount = input(prompt2)
+        if amount == "q":
+            return
+        try:
+            amount = float(amount)
+        except ValueError:
```
charlesrobison, line 115: Would like to see more exceptions added

PR #1039, opened 2020-08-12
Finished HTML Render Project

westobrien:


natasha-aleksandrova:
@westobrien hello, the code submitted doesn't contain any code - maybe you forgot to commit and push the code?

westobrien:
Hi Natasha,

My mistake. I committed the code and pushed. Are you seeing the new request
with the code? It didn't let me submit another pull request, but it looks
like it added onto the one I already have?

Thank you!

students/westobrien/Lesson07/html_render.py
```diff
@@ -0,0 +1,144 @@
...
+    indent = "    "
+
+    def __init__(self, content=None, **kwargs):
+        if content is None:
+            self.contents = []
+        else:
+            self.contents = [content]
+        self.attributes = []
+        for item in kwargs.items():
+            self.attributes.append(item)
```
natasha-aleksandrova, line 22: you can do the same thing with `list(d.items())` although there is no need to convert to a list, you can just use it as a dictionary.

PR #1040, opened 2020-08-12
Updated mailroom_part3

derizor:
Please take a look at function "def send_ty_note():" lines 84-101.  After changing data source from list to dict the function does not search through all names, but sees only first one on the list.
Thank you
Deniss

### DIFF
--- a/students/Deniss_Semcovs/Lesson05/mailroom_part03.py
+++ b/students/Deniss_Semcovs/Lesson05/mailroom_part03.py
@@ -1,0 +1,153 @@
+#!/usr/bin/env python3
+
+import tempfile
+import os.path
+
+# KeyError exception added
+def mailroom02(reply, action):
+    while True:
+        try:
+            resp = input(reply)
+            if action[resp]() is None or False:
+                pass
+            elif action[resp]() == "exit menu":
+                break
+        except KeyError:
+            keyerror()
+        except KeyboardInterrupt:
+            keyboardinterrupt()
+
+# Complerssed "list"
+def print_list():
+    new_data = dict_donor()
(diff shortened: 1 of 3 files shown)


students/Deniss_Semcovs/Lesson05/mailroom_part03.py
```diff
@@ -0,0 +1,153 @@
...
+    print(">>>> Hint: 'q'")
+
+data = [
+    ["Tom Hanks", 1000, 500],
+    ["Leonardo DiCaprio", 300, 1200, 500],
+    ["Natalie Portman", 5000],
+    ["Harrison Ford", 500, 2000, 100],
+    ["Scarlett Johansson", 1500, 1000],
+    ["Jennifer Aniston", 2000, 2000, 2000],
+]
```
natasha-aleksandrova, line 130: this should be a dictionary data structure.

students/Deniss_Semcovs/Lesson05/mailroom_part03.py
```diff
@@ -0,0 +1,153 @@
...
+                print("Hello {}, thank you for your generous donation of ${} to support our cause.".format(i['name'], i['don']))
+            # I had to remove "else:" because function stoped checking values 
+            # after first dictionary, please let know what did i do wrong here?
+            """
+                break
+            else:
+                print("The donor's name is not on the list! Please add a new donor through 'Add new donation' option.")
+                print(i["name"])
+                break
+                """
```
natasha-aleksandrova, line 99: this is because you have a `break` here - to find donor not in your list you want your `else` to be on the same indent level as `for`

PR #1044, opened 2020-08-13
Lesson 9 assignment

Lina-271:


students/Lina/Lesson9/donor_models.py
```diff
@@ -0,0 +1,164 @@
...
+            if details:
+                self.details = details
+            else:
+                self.details = {"gift_amounts" : [0]}
+        else:
+            self.details = donor_details
+
+    @classmethod
+    def lookup_donor(cls, name):
+        return DonorCollection().lookup_donor(name)
```
natasha-aleksandrova, line 28: this method does not belong here - remember this class is all about a single donor, looking up a donor deals with looking through a collection of donors and belongs in DonorCollection class

students/Lina/Lesson9/donor_models.py
```diff
@@ -0,0 +1,164 @@
...
+# Donor class contains information of a single donor
+class Donor:
+    def __init__(self, donor_name, donor_details=None):
+        self.full_name = donor_name
+        if donor_details == None:
+            details = self.lookup_donor(donor_name)
+            if details:
+                self.details = details
+            else:
+                self.details = {"gift_amounts" : [0]}
```
natasha-aleksandrova, line 22: why store this in a dictionary? it seems that it is always a single value.

we use classes to help us structure data, so whenever possible you want to use a class attribute to stored data.

students/Lina/Lesson9/donor_models.py
```diff
@@ -0,0 +1,164 @@
...
+                          "{sender_alignment}{sender_name}",
+                          "{sender_alignment}{sender_title}",
+                          "\n\n\n"))
+        return letter_layout, sender_template
+
+# DonorCollection class contains information about all donors
+class DonorCollection:
+    #This dictionary database contains the donor names and a history of their donation amounts.
+    def __init__(self):
+        self.donor_db = self.get_donors_history()
```
natasha-aleksandrova, line 94: this class should hold reference to your donor objects, so a good data structure could be:
```
self.donor_db = {"<donor_name_str>": <donor_object_instance>}
```

students/Lina/Lesson9/donor_models.py
```diff
@@ -0,0 +1,164 @@
...
+        self.donor_db[name]["gift_amounts"].append(amount)
+
+    def donors_summary(self):
+        #Returns a list with summary of each donor in a tuple, sorted by total donated from
+        #highest to lowest.
+        # [(name, total donated, donation count, average amount)]
+        summary = []
+        for name, detail in self.donor_db.items():
+            d = Donor(name, detail)
+            summary.append((name, d.total_gift_amount, d.total_gift_count, d.avg_gift_amount))
```
natasha-aleksandrova, line 117: you should use the donor information as part of your collection, not create an instance on the fly - this defeats the purpose of having a Donor class.

students/Lina/Lesson9/donor_models.py
```diff
@@ -0,0 +1,164 @@
...
+            line = row(name=col[0], total_donated=col[1], gift_count=col[2], avg_amount=col[3]),
+            lines += tuple(line)
+        return "\n".join(lines)
+
+    def letters_to_all_donors(self):
+        #Compose a letter to all donors with their total contribution included.
+        #Returns all letters in a list of tuples [(letter, fullname), (letter, fullname), ...]
+        letters = []
+        for name, detail in self.donor_db.items():
+            d = Donor(name, detail)
```
natasha-aleksandrova, line 139: same here - your data should come from `self.donor_db`

Lina-271:
Hi Natasha,

Thanks for all the great comments. I have one question, when should you use class attributes and when to use instance attributes?

Thanks,
Lina

Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows 10

natasha-aleksandrova:
@Lina-271 that's a great question - let's go back to our Lesson 7 html renderer assignment which shows a good example of that.

```
class Element:
    indent = 4 # class attribute
    def __init__(self, content):
        self.content = content # instance attribute
```


what above class is saying:
* the indent is owned by the class, meaning that each instance you create `Element('stuff')` will have indent=4, indent is not unique to the instance. Yes you can change the class attribute if you wanted an indent for your program to be 8.. but that is not something you intend to change with each instance.
* the "content" is most likely unique to each instance, each time you create an `Element` instance you will provide the content that is appropriate: `Element('stuff2')` `Element('stuff3')`


here is another example, where class attribute is more like a constant here:
```
class Nissan:
    maker = 'Japan'
    def __init__(self, model, features):
        self.model = model
        self.features = features

altima = Nissan('Altima', **kwargs)
maxima = Nissan('Maxima', **kwargs)
```

Lina-271:
I think I understand the examples that you  provided.  So now my question is, why should the donor name and donor details in the donor class should be stored in class attributes since they are unique to the instance?  Thanks!

Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows 10

natasha-aleksandrova:
@Lina-271 I don't believe my comments suggested that donor name or details should be class attributes - my comments talked about what class things belong in.

Let me know give you a more concrete example on what is expected here.

Lina-271:
That was my misunderstanding. Would you please give me an example of what you meant in your last comment below?  Thank you!


+# Donor class contains information of a single donor

+class Donor:

+    def __init__(self, donor_name, donor_details=None):

+        self.full_name = donor_name

+        if donor_details == None:

+            details = self.lookup_donor(donor_name)

+            if details:

+                self.details = details

+            else:

+                self.details = {"gift_amounts" : [0]}

why store this in a dictionary? it seems that it is always a single value.

we use classes to help us structure data, so whenever possible you want to use a class attribute to stored data.



Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows 10

Lina-271:
This helps, thank you!

Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows 10

natasha-aleksandrova:
glad to hear!

PR #1045, opened 2020-08-14
adding first Lesson01 files and README.md

nicksama88:
Went through and did all of the "Warmups" on the CodingBat site.  Didn't realize my progress would be lost, so only have the solution to the final warmup problem saved.  Will add the rest when I get a chance, and will try and do the rest of the problems beyond the "Warmups" as well.


nicksama88:
Thanks!  Excited to be here.

PR #1046, opened 2020-08-14
Added mailroom part 3 script and modified except_exercise script for …

s-mar206:
…lesson 05

I am not sure if I needed to add any list comprehensions for my mailroom code, it didn't seem there was any appropriate places with my current structure. Also, only saw a couple obvious locations to add a try-except code block but not sure if I should be adding more.

### DIFF
--- a/students/Steven/lesson05/mailroom03.py
+++ b/students/Steven/lesson05/mailroom03.py
@@ -1,0 +1,101 @@
+#! bin/user/env python3
+import sys
+from operator import itemgetter
+"""
+Program Part2:
+Update your mailroom program to:
+
+Use dicts where appropriate.
+See if you can use a dict to switch between the user’s selections.
+See if you can use a dict to switch between the users selections. see Using a Dictionary to switch for what this means.
+Convert your main donor data structure to be a dict.
+Try to use a dict and the .format() method to produce the letter as one big template,
+rather than building up a big string that produces the letter in parts.
+"""
+
+# Donor name, donation amount, donation history
+donor_list = {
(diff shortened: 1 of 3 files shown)


students/Steven/lesson05/mailroom03.py
```diff
@@ -0,0 +1,101 @@
...
+            choice = int(input("What would you like to do? "))
+            # Using a dictionary for as a switch case
+            switch_func_dict = {
+                1: thank_you,
+                2: report,
+                3: all_letter,
+                4: quit
+            }
+            switch_func_dict.get(choice)()
+        except ValueError as error_message:
```
natasha-aleksandrova, line 39: 👍 - a nice thing to do is to add a user friendly message here `print("Invalid menu option")`


what happens when they enter 5 ? (it will crash your program) this should be caught as well

students/Steven/lesson05/mailroom03.py
```diff
@@ -0,0 +1,101 @@
...
+        if name.lower() == "list":
+            print(donor_list)
+            thank_you()
+        gift = float(input("Enter a donation amount: "))
+        if name in donor_list.keys():  # check if name is in the dict
+            donor_list[name][0] += gift  # Total balance of donations
+            donor_list[name][1] += 1  # Keeping track of donation history
+        else:
+            donor_list.update({name: [gift, 1]})
+    except ValueError as error_message:  # Catch if the user doesn't input a number
```
natasha-aleksandrova, line 64: you should have this around line 58 so it's clear what piece of code you're catching

PR #1052, opened 2020-08-17
Jenkins: Lesson 5 Complete

jdj0891:
Completed Lesson 5 exercises and assignments.

I was unable to identify an appropriate location to use a Comprehensions in my mailroom part 3 assignment. Do you have any suggestion where they would have been value added?


students/jason_jenkins/lesson05/mailroom_part3.py
```diff
@@ -0,0 +1,186 @@
...
+    print('   2 - Create a Report.')
+    print('   3 - Send letters to all donors.')
+    print('   4 - Quit.')
+
+    response = input("Input numbered option you wish to do: ").strip()
+
+    if response in menu_dict:
+        menu_dict[response]()
+    else:
+        print(f"{response} is not a valid input.")
```
natasha-aleksandrova, line 164: this is a good place to implement exception handling (catch KeyError)

PR #1059, opened 2020-08-20
added mailroom part 4 and unit testing scripts for lesson6

s-mar206:
Mailroom Part 4 script and Unit testing for mailroom part 4 script.

### DIFF
--- a/students/Steven/lesson06/mailroom04_test.py
+++ b/students/Steven/lesson06/mailroom04_test.py
@@ -1,0 +1,39 @@
+#! bin/user/env python3
+
+import mailroom04 as mailroom
+import os.path
+'''
+Unit testing for mailroom code
+'''
+
+def test_donor_list():
+    donor_list = mailroom.donor_list
+    assert donor_list['John Doe'] == [2000, 2]
+    assert donor_list['Jane Doe'] == [3000, 3]
+
+def test_update_donor_list():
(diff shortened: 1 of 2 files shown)


students/Steven/lesson06/mailroom04_test.py
```diff
@@ -0,0 +1,39 @@
...
+import mailroom04 as mailroom
+import os.path
+'''
+Unit testing for mailroom code
+'''
+
+def test_donor_list():
+    donor_list = mailroom.donor_list
+    assert donor_list['John Doe'] == [2000, 2]
+    assert donor_list['Jane Doe'] == [3000, 3]
```
natasha-aleksandrova, line 12: this typically wouldn't constitute as a test - a test should test your code and would require execute a function from your code.

students/Steven/lesson06/mailroom04_test.py
```diff
@@ -0,0 +1,39 @@
...
+    assert os.path.isfile('thank_you_benjamin franklin.txt')
+
+if __name__== "__main__":
+
+    test_donor_list()
+    test_update_donor_list()
+    test_thank_you_letter()
+    test_get_report()
+    test_all_letters()
+    print("Mailroom unit tests have all passed")
```
natasha-aleksandrova, line 39: make sure you're using `pytest` to run your tests instead of manually calling the functions.

PR #1062, opened 2020-08-21
Lesson03 Finished

jacobian91:
Finished Lesson 3. I got mixed up on number on my EX3.x numbering in my commit messages. Ignore the Lesson2 files in here, use prior pull request for lesson2 #1058.


students/jerickson/Lesson3/mailroom.py
```diff
@@ -0,0 +1,154 @@
...
+        if not sorted_data:
+            sorted_data.insert(0, donor_record)
+        else:
+            for index_sorted, sorted_record in enumerate(sorted_data[:]):
+                if donor_record[1] >= sorted_record[1]:
+                    sorted_data.insert(index_sorted, donor_record)
+                    break
+            else:
+                sorted_data.append(donor_record)
+    donation_data = sorted_data
```
natasha-aleksandrova, line 26: you should use sorted() function here 🙂

students/jerickson/Lesson3/mailroom.py
```diff
@@ -0,0 +1,154 @@
...
+            "list",
+            "King ArThur",
+            "400.2",
+            "create a report",
+            "quit",
+        ]
+        mocked_resp_gen = response_generator(mocked_responses)
+
+    quit_flag = False
+    while not quit_flag:
```
natasha-aleksandrova, line 141: you can simplify this a bit by using a `while True` with a `break` statement; avoids using a flag variable.

students/jerickson/Lesson3/slicing_lab.py
```diff
@@ -0,0 +1,55 @@
...
+
+def every_other(seq):
+    """Returns after removing every other element"""
+    every_other = seq[::2]
+    return every_other
+
+
+def missing_head_tail_every_other(seq):
+    """Returns after; removing the head-4 and tail-4 elements, then cutting every other element"""
+    missing_head_tail = seq[4:-3]
```
natasha-aleksandrova, line 15: this should be -4 (instead of -3)

students/jerickson/Lesson3/mailroom.py
```diff
@@ -0,0 +1,154 @@
...
+        command = input(
+            "\nChoose: Say “Send a Thank You”, “Create a Report” or “Quit” ->: "
+        )
+        command = command.lower()
+        if command == "send a thank you":
+            thank_you()
+        elif command == "create a report":
```
natasha-aleksandrova, line 148: you should have numeric options for the user to choose from - typing out several words is a lot of work and error prone

students/jerickson/Lesson3/mailroom.py
jacobian91: sorted now used in Lesson04 :)

students/jerickson/Lesson3/mailroom.py
jacobian91: Incorporated into Lesson04 mailroom, thank you.

students/jerickson/Lesson3/mailroom.py
jacobian91: Incorporated into Lesson04 mailroom, thank you.

PR #1075, opened 2020-08-28
Lesson05 except_exercise & Mailroom 3

nicksama88:
except_exercise.py along with the updated Mailroom program with some exception handling.  Not sure if there are more places where exception handling would be good to have, wasn't able to force any exceptions to be thrown.  Also, not sure where comprehension could have made the code cleaner... if you have any suggestions I'd love to hear them!


students/Nick_Haury/Lesson05/mailroom3.py
```diff
@@ -0,0 +1,172 @@
...
+        "4 - Quit",
+        ">> "
+        ))
+
+    while True:
+        choice = input(prompt)
+        if choice in choice_dict:
+            choice_dict[choice]()
+        else:
+            print(str(choice) + " is an invalid selection, "
```
natasha-aleksandrova, line 63: this is a good place to utilize exception handling (hint: you can catch keyerror)

students/Nick_Haury/Lesson05/mailroom3.py
```diff
@@ -0,0 +1,172 @@
...
+                      " amount\n")
+            else:
+                if donation_amount > 0:
+                    add_donation(ty_input, donation_amount)
+                    print(
+                        "\n", create_email(ty_input, donation_amount), sep=""
+                        )
+                    break
+                else:
+                    print("Please enter a positive number\n")
```
nicksama88, line 102: I was thinking this is also a good place for exception handling, but I wasn't sure how to create an error for a non-positive number.  After your suggestion on the other spot for exception handling, I was able to figure out how to make it work here too up in the earlier "try" statement.

students/Nick_Haury/Lesson05/mailroom3.py
nicksama88: I see... This makes sense now.  It's like doing the if/else statement in reverse.  I was able to rework the code; thanks for the suggestion!

PR #1076, opened 2020-08-28
Lesson05 Finished

jacobian91:


students/jerickson/Lesson5/mailroom.py
```diff
@@ -0,0 +1,294 @@
...
+            print(f"Unrecognized number: {donor_amount}. Try again.")
+
+    new_donation(donor_name, donor_amount)
+    email = compose_new_donation_email(donor_name, donor_amount)
+    print(email)
+
+
+def quit_menu():
+    """Return the string "quit" to exit a menu-level"""
+    return "quit"
```
natasha-aleksandrova, line 202: if you use sys.exit() here then you can utilize the switch dict and not need special checks below.

jacobian91: I haven't implemented it yet, but my plan is to use this as a 'quit' a sub-menu as well using the menu-selection recursively. If I don't implement that in the oo_mailroom, I'll implement the sys.exit(). Thank you.

PR #1077, opened 2020-08-28
Lesson06 Finished

jacobian91:
Lesson05 Files committed as part of #1076


students/jerickson/Lesson6/test_mailroom.py
```diff
@@ -0,0 +1,542 @@
...
+        # Execute
+        mailroom.report_cli()
+
+        # Assert
+        for _call in mocked_print.call_args_list:  # Report printed in order
+            _call_arg = _call.args[0]  # Argument print was called with
+            assert _call_arg == report_list[_call_arg - 1]
+
+
+class Test_Report:
```
natasha-aleksandrova, line 55: there is no reason to use class with a single test function (classes are used with `unittest` module)

with `pytest` you can just use functions directly.

jacobian91: I use them like that for consistency so that looking through the code is a bit more clear. Is this bad practice?
Thank you for the feedback!!

natasha-aleksandrova: yes it is not standard way to write pytest tests. they are usually written as individual functions.

PR #1081, opened 2020-08-30
Lesson06 Mailroom Part 4

nicksama88:
Mailroom updated to include unit testing.  Wasn't sure if the testing should go in a separate file or not, so it is at the bottom of the code.  Instructions didn't seem to indicate either way?


students/Nick_Haury/Lesson06/mailroom4.py
```diff
@@ -0,0 +1,317 @@
...
+            else:
+                add_donation(ty_input, donation_amount)
+                print(
+                    "\n", create_email_text(ty_input, donation_amount), sep=""
+                    )
+                break
+
+
+def get_donor_list(donors=donors):
+    return list(donors.keys())
```
nicksama88, line 109: The refactoring suggestion in the prompt seemed to indicate this functionality should be broken out for testing.  It's very straightforward, though, so not sure if it should really have been broken out?

students/Nick_Haury/Lesson06/mailroom4.py
```diff
@@ -0,0 +1,317 @@
...
+    return(f"Dear {donor_name},\n\n"
+           f"It is with incredible gratitude that we accept your wonderfully "
+           f"generous donation of ${donation_amount:,.2f}.  Your "
+           "contribution will truly make a difference in the path forward "
+           "towards funding our common goal."
+           "\n\nEver Greatefully Yours,\n\n"
+           "X" + ("_" * 20) + "\n")
+
+
+def add_donation(donor_name, donation_amount, donors=donors):
```
nicksama88, line 127: Since I have my donors dictionary as a global, it seemed like the only way I could use custom donor dictionaries in my tests was to include a keyword argument for the dictionary.  I'm not sure if this is good practice or not though... I think I remember one of the readings discussing creating dummies somehow; may need to investigate that further if this is bad practice.

students/Nick_Haury/Lesson06/mailroom4.py
```diff
@@ -0,0 +1,317 @@
...
+    return text_list
+
+
+def write_letters(donors=donors):
+    for key in donors:
+        with open(f'{key}.txt', 'w') as f:
+            f.write(create_email_text(key, donors[key][-1]))
+    print()
+
+# Run unit tests
```
nicksama88, line 173: Wasn't sure if the tests should be broken out into a separate file or not.  Prompt didn't seem to indicate either way, so kept them within the main file.

students/Nick_Haury/Lesson06/mailroom4.py
natasha-aleksandrova: yes, in your test file you can override your global data structure

```
import mailroom
mailrom.donors = {}
```

students/Nick_Haury/Lesson06/mailroom4.py
natasha-aleksandrova: yes you do want to keep your tests in a separate file

PR #1086, opened 2020-08-31
Added circle class assignment for lesson 08.

s-mar206:
circle_class.py and test_circle_class.py scripts added

### DIFF
--- a/students/Steven/lesson08/circle_class.py
+++ b/students/Steven/lesson08/circle_class.py
@@ -1,0 +1,92 @@
+#!/usr/bin/env python3
+"""
+The goal is to create a class that represents a simple circle.
+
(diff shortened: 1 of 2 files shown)


students/Steven/lesson08/circle_class.py
```diff
@@ -0,0 +1,92 @@
...
+    # Circle object with a given radius
+    def __init__(self, radius):
+        self.radius = radius
+
+    @property
+    def get_diameter(self):
+        return self.radius * 2
+
+    @get_diameter.setter
+    def get_diameter(self, value):
```
natasha-aleksandrova, line 29: since its a property you want to use nouns for the name (for functions you'd use verbs like get_*)

natasha-aleksandrova (approved):
@s-mar206 great work here!

PR #1091, opened 2020-09-05
Lesson07 files, Lesson08 files, and Mailroom4 refactoring tests

nicksama88:
Lesson07 files completed.  Refactoring Mailroom4 tests into their own test file.


students/Nick_Haury/Lesson07/html_render.py
```diff
@@ -0,0 +1,159 @@
...
+class H(OneLineTag):
+
+    tag = "h"
+    header_level = ""
+
+    def __init__(self, header_level="", content=None, **kwargs):
+        try:
+            int(header_level)
+            self.header_level = header_level
+        except ValueError:
```
nicksama88, line 145: Is this an appropriate way to test that an int was passed?

students/Nick_Haury/Lesson08/test_circle.py
```diff
@@ -0,0 +1,113 @@
...
+    assert repr(c) == "Circle(5.000)"
+    assert str(c) == "Circle with radius: 5.000"
+
+def test_add():
+    c1 = Circle(1)
+    c2 = Circle(2)
+    c3 = c1 + c2
+    print(c3)
+    # assert c3.radius == 3.0
+    assert repr(c3) == 'Circle(3.000)'
```
nicksama88, line 62: is checking the repr() a good way to go here, or should I be checking using something like... isinstance maybe?

students/Nick_Haury/Lesson07/html_render.py
```diff
@@ -0,0 +1,159 @@
...
+            if header_level != "":
+                print("header_level must be blank or an integer."
+                      " Defaulting to blank.")
+                self.header_level = ""
+        super().__init__(content, **kwargs)
```
natasha-aleksandrova, line 150: you can simplify this a bit:
```
try:
    self.tag = f"h{int(header)}"
except ValueError:
    raise Exception("must provide a number for heading")
```

best thing to do here is to raise an exception to the user of your class.

students/Nick_Haury/Lesson08/circle.py
```diff
@@ -0,0 +1,82 @@
...
+    def __str__(self):
+        return f"Sphere with radius: {self.radius:.3f}"
+
+    @property
+    def volume(self):
+        return (4 / 3) * math.pi * self.radius**3
+    
+    @property
+    def area(self):
+        return 4 * math.pi * self.radius**2
```
natasha-aleksandrova, line 82: looks good 👍

students/Nick_Haury/Lesson08/test_circle.py
natasha-aleksandrova: ideally your test should be asserting on what you're testing, and unless you're testing the `__repr__` method you shouldn't be using `repr()` 

since you're testing the add() method you should just verify the radius attribute equal your expected new radius attribute

natasha-aleksandrova (approved):
@nicksama88 nice work here 👍

PR #1094, opened 2020-09-05
adding lesson04

belarsongit:


charlesrobison (approved):
Nice updates to mailroom and you're almost there.  Currently adding donations to existing donors is creating duplicate donors instead of adding to their total.  Take a look and let me know if you have questions.  This is the basis for further iterations of mailroom, so important to get this working correctly.

Ready to merge.

PR #1097, opened 2020-09-07
Fixed mailroom_part03 for assignemt 4

derizor:


natasha-aleksandrova (commented):
@derizor please review the guidelines and the below code stub for correct implementation of OO design for the mailroom:
https://repl.it/@naaleksandrova/Lesson9Example#main.py

PR #1099, opened 2020-09-08
Lesson01_Activities

mellamokrysti:
Activities from Lesson 01

### DIFF
--- a/students/Kristy_Martini/lesson01/break_me.py
+++ b/students/Kristy_Martini/lesson01/break_me.py
@@ -1,0 +1,41 @@
+def createNameError():
+    name = "Kristy"
+    try:
+        print("My name is " + nome)
+    except NameError:
+        print("You have a NameError")
+
+def createTypeError():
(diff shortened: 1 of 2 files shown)


students/Kristy_Martini/lesson01/break_me.py
```diff
@@ -0,0 +1,41 @@
+def createNameError():
```
charlesrobison, line 1: Functions (and almost everything except classes) should be snake_case in Python: create_name_error():

students/Kristy_Martini/lesson01/sleep_in.py
```diff
@@ -0,0 +1,7 @@
+def sleep_in(weekday, vacation):
+  if vacation:
```
charlesrobison, line 2: Make sure all indentations are 4 spaces

charlesrobison (approved):
Nice start!

Ready to merge

PR #1103, opened 2020-09-10
Lesson06 sorting added and lesson09 assignment

ssinghshweta07:
Final assignment added here as well recommendation of lesson06 for adding sorting on sum of donation amount had been added.

### DIFF
--- a/students/Shweta/Lesson09/donor_model.py
+++ b/students/Shweta/Lesson09/donor_model.py
@@ -1,0 +1,84 @@
+#!/usr/bin/env python
+
+'''thankyou message for donation'''
+mty_msg="\n".join(("Thankyou {}",
+                   "for your generous donation of {:.2f}",))
+
+
+ltr="\n".join(("", "Dear {},",
+              "Thank you for your very kind donation of ${:.2f}.",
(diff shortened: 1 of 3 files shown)


students/Shweta/Lesson09/donor_model.py
```diff
@@ -0,0 +1,84 @@
...
+        return (self.total_sum/self.tn_times)
+
+
+    def ind_letter(self):
+        indv_letter=ltr.format(self.name,self.total_sum)
+        return indv_letter
+    
+    def formatting(self):
+        self.frmt='{:20} ${:>20.2f}   {:18}   ${:20.2f}'.format(self.name,self.total_sum,self.tn_times,self.avg)
+        return self.frmt
```
natasha-aleksandrova, line 42: instance attributes should be defined in `__init__` - however I don't think you need this to be an instance attrb.

this just be `frmt = ...` or 
`return '{:20} ${:>20.2f}   {:18}   ${:20.2f}'.format(self.name,self.total_sum,self.tn_times,self.avg)`

natasha-aleksandrova (approved):
@ssinghshweta07 great work here!

PR #1106, opened 2020-09-10
Mailroom Part 2 for final check

belarsongit:


students/becky_larson/lesson04/Mailroom_Part2.py
```diff
@@ -0,0 +1,198 @@
...
+    print(f'{col1:25} | {col2:13} | {col3:11} | {col4:13}')
+    print('-'*70)
+
+    donors = dict(sorted(donors.items(), key=sort_key, reverse=True))
+    for donor, donations in donors.items():
+        count = len(donations)
+        total = sum(donations)
+        avg = total/count
+        print(f'{donor:25}  ${total:13.2f}   {count:11}  ${avg:12.2f}')
+    return
```
charlesrobison, line 123: This doesn't appear to be generating a 'thank you' to the screen

students/becky_larson/lesson04/Mailroom_Part2.py
```diff
@@ -0,0 +1,198 @@
...
+
+    temp_path = tempfile.gettempdir()
+    folder_name = user_folder + '_' + today.strftime("%b_%d_%Y")
+
+    folder_path = os.path.join(temp_path, folder_name)
+
+    if not os.path.exists(folder_path):
+        os.makedirs(folder_path)
+
+    return folder_path
```
charlesrobison, line 146: Love this function, but not sure if it's working correctly.  Not finding the letters generated

students/becky_larson/lesson05/Mailroom_Part3.py
```diff
@@ -0,0 +1,232 @@
...
+    while True:
+        prompt = "Please enter a valid amount to donate >"
+        donation = input(prompt)
+        if donation == "q":
+            return
+
+        try:
+            donation = float(donation)
+        except ValueError:
+            continue
```
charlesrobison, line 67: Ideally you'd prompt the user with a statement of some sort

charlesrobison (approved):
Looks like a couple of issues with both mailroom files - please take a look and let me know if you have questions.  Would like to see more exceptions as well in part 3.  Make sure you submit your code to the course site for grading as well.

Ready to merge

belarsongit:
Thanks for the reply!

I do have some questions:

*Section 1 Comment:  This doesn't appear to be generating a 'thank you' to
the screen*
Looks like this is the create_report function.  If you could point me to
this in the assignment so I know how to incorporate it correctly, I will
update accordingly.

*Section 2 Comment:  Love this function, but not sure if it's working
correctly. Not finding the letters generated*
Yes, I spent extra time to figure out and learn the tempfile stuff.  It is
working on my computer, but maybe because it's windows?  I thought about
that as I didn't code for mac.  Are you on a mac?  If on windows, it should
create a folder in the temp folder.  Mine creates a folder in:
"C:\Users\belarson\AppData\Local\Temp\"

*Section 3 Comment:  Ideally you'd prompt the user with a statement of some
sort*
Do you mean more text than I have ("Please enter a valid amount to donate >")?
I am an old programmer and am usually cryptic.  Let me know if I
understand correctly and I will provide a better prompt.

I will submit Part 2 for grade and improve the exceptions in Part 3 before
submitting.

Thanks!

natasha-aleksandrova:
@belarsongit 

1. Is part of the original requirement: https://uwpce-pythoncert.github.io/PythonCertDevel/exercises/mailroom-part1.html#send-a-thank-you; the last bullet point
```Finally, use string formatting to compose an email thanking the donor for their generous donation. Print the email to the terminal and return to the original prompt.```
2. Given that this will be in a different location on different platforms, a nice feature would be to tell the user where to expect those files. You can just print this variable `folder_path`
3. Yes, since this is an interactive program you'll want to let the user know that what they entered is invalid. Using a `continue` in exception block is less common (there are valid cases for that of course).
```
except ValueError:
    print("You have entered Invalid value for donation amount. Please try again.")
```

 


@charlesrobison please add if I missed anything.

PR #1108, opened 2020-09-11
Add mailroom 00 scripts for lesson 09

s-mar206:


students/Steven/lesson09/donor_models.py
```diff
@@ -0,0 +1,97 @@
...
+    def __init__(self, donor_name, new_donation=None):
+        self.name = donor_name
+        if new_donation is None:
+            self.donation = []
+        else:
+            self.donation = list(new_donation)
+
+    def new_donation(self, new_amount):
+        update_donation = self.donation.append(new_amount)
+        return update_donation
```
natasha-aleksandrova, line 14: you don't need to return anything - you have access to the `self.donation` attribute.

besides, `append()` returns `None` so the return here has no purpose

students/Steven/lesson09/donor_models.py
```diff
@@ -0,0 +1,97 @@
...
+    def donation_count(self):
+        return len(self.donation)
+
+    @property
+    def donation_avg(self):
+        return round(float(sum(self.donation) / self.donation_count), 2)
+
+    def thank_you_letter(self, donor_name, gift):
+        # Thank you email to donor for their gift
+        email = f"Thank you {donor_name} for your generous donation of ${gift:.2f}, your kindness is very appreciated." + "\n"
```
natasha-aleksandrova, line 30: you have the instance attribute for the name `self.name` - this function should not have `donor_name` input

natasha-aleksandrova: you can also grab the last element in self.donation for the gift value

students/Steven/lesson09/donor_models.py
```diff
@@ -0,0 +1,97 @@
...
+        for donor in self.donor_list:
+            if donor.name == donor_name.title():
+                donor.new_donation(new_donation)
+        return self.donor_list
+
+    # Generate the current list of donors that is also sorted.
+    def list_donors(self):
+        all_donors = []
+        for donor in self.donor_list:
+            all_donors.append([donor.name, donor.donation_total, donor.donation_avg])
```
natasha-aleksandrova, line 67: this loop is wasteful - you can pass in `self.donor_list` directly to `sorted()` function.

natasha-aleksandrova (approved):
@s-mar206 overall good job, few issues to review in my comments.

PR #1115, opened 2020-09-15
Lesson09 Object Oriented Mailroom Files

nicksama88:
Files for the Lesson09 Object Oriented version of the mailroom program.  

Split the files into 3:
The command line interface (cli_main.py) 
The classes (donor_models.py)
The test file (test_mailroom_oop.py).  

I'm sure there are more clever ways for checking against capitalization, and for organizing the donors within the DonorCollection Class.  I'd be interested in any insights on improvements!


students/Nick_Haury/Lesson09/mailroom_oo/donor_models.py
```diff
@@ -0,0 +1,122 @@
...
+        for donor in sorted(self.donors.values(), reverse=True):
+            report_list.append(f"{donor.name:26}"
+                               f"${sum(donor.donations):14,.2f}"
+                               f"{len(donor.donations):11}  "
+                               f"${st.mean(donor.donations):16,.2f}\n")
+        return report_list
+
+    def add_donation(self, donor_name, donation):
+        if donor_name.lower() in self.lower_names:
+            # check to see if different capitalization is being used
```
natasha-aleksandrova, line 114: you could store all names in lower case and capitalize on any display by using a property

natasha-aleksandrova (approved):
@nicksama88 excellent work here!

students/Nick_Haury/Lesson09/mailroom_oo/donor_models.py
nicksama88: I was wondering about this... I was thinking I'd utilize the string.title() method, but it doesn't work for William Gates, III (results in William Gates, Iii).  Would have to create a specific property for that... I think I read about something called... regular expressions?  Maybe this is a good use case for that?

PR #1122, opened 2020-09-18
Lesson09 OO mailroom changes from feedback

nicksama88:
Made changes to OO mailroom based on feedback; have names saved to classes and dictionaries as lower case, then have properties for showing capitalized "display versions" for reporting, etc.  Cleaned up some of the doc strings as well.


natasha-aleksandrova (approved):
👍

PR #1127, opened 2020-09-20
Lesson01 - Exercises

LongSA:
Submitting lesson 01 exercises - errors (1) and puzzles (8).


students/steve-long/lesson01-enviro/activities-exercises/task-1-exploring-errors/break_me.py
```diff
@@ -0,0 +1,89 @@
+# Python210 | Fall 2020
+# ----------------------------------------------------------------------------------------
+# Lesson01
+# Task 1: Exploring Errors (break_me.py)
+# Steve Long 2020-09-10
+
+# /Users/steve/Documents/Project/python/uw_class/python210/lessons/lesson01-enviro/break_me.py
+
+def runWithNameError():
```
charlesrobison, line 9: All function names should be snake_case in Python: run_with_name_error():

charlesrobison: Looks like indentations are over indented too.  Make sure your text editor is set up for 4 spaces.

charlesrobison (approved):
Looking good Steve - nice start here!  Would recommend installing a linter on your text editor to catch any PEP8 style issues.  All indentations should be 4 spaces, for example.  Let me know if you have questions.

Ready to merge

LongSA:
I just go this message from you guys. Thanks.Indentations are set for four spaces in my BBEdit application. I’ve had the same set-up for Ruby, C++, Java, JavaScript, and now Python in BBEdit for quite a while (though I don’t use it much for Java anymore.) It transfers just fine between BBEdit (Mac) and VisualStudio (Windows) though I’ve seen a few glitches in the text windows in GitHub when transferring files authored in either application. I believe that’s an HTML element issue.Do you really want to use what you call “snake case”? The code executes just fine the way I’m doing it, is a lot less ugly, and takes up less room. If you want to do it that way, it’s fine with me - just a little odd.Where can I find the style guide for Python that will make my source code comfortable for everyone?Steve LongOn Sep 21, 2020, at 8:51 PM, Charles Robison <notifications@github.com> wrote:@charlesrobison approved this pull request.Looking good Steve - nice start here!  Would recommend installing a linter on your text editor to catch any PEP8 style issues.  All indentations should be 4 spaces, for example.  Let me know if you have questions.Ready to mergeIn students/steve-long/lesson01-enviro/activities-exercises/task-1-exploring-errors/break_me.py:
> @@ -0,0 +1,89 @@
+# Python210 | Fall 2020
+# ----------------------------------------------------------------------------------------
+# Lesson01
+# Task 1: Exploring Errors (break_me.py)
+# Steve Long 2020-09-10
+
+# /Users/steve/Documents/Project/python/uw_class/python210/lessons/lesson01-enviro/break_me.py
+
+def runWithNameError():
All function names should be snake_case in Python: run_with_name_error():

In students/steve-long/lesson01-enviro/activities-exercises/task-1-exploring-errors/break_me.py:
> @@ -0,0 +1,89 @@
+# Python210 | Fall 2020
+# ----------------------------------------------------------------------------------------
+# Lesson01
+# Task 1: Exploring Errors (break_me.py)
+# Steve Long 2020-09-10
+
+# /Users/steve/Documents/Project/python/uw_class/python210/lessons/lesson01-enviro/break_me.py
+
+def runWithNameError():
Looks like indentations are over indented too.  Make sure your text editor is set up for 4 spaces.—You are receiving this because you authored the thread.Reply to this email directly, view it on GitHub, or unsubscribe.

LongSA:
Greetings…

I have a question regarding the best choice for an Exception to raise. This relates to some augmentation I have made to the Circle and Sphere classes for the exercise in Lesson08.

I have a constructor that takes a center and radius to make the object. If the object is a Circle, center must be a tuple consisting of two int or float elements; three if the object is a Sphere. I have a quality assurance step that checks the center. It raises the following exceptions for the following conditions:

(1) TypeError => Center is NOT a tuple or an element of the tuple is not an int or float
(2) RuntimeError => Number of elements in tuple not 2 if object is a Circle, not 3 if object is a Sphere

It’s the second one I’m not sure about. There did not seem to be an exact fit for this in the built-in Exceptions list and it does not really seem like a TypeError or ValueError. I don’t want to gen up a custom Exception just for this assignment. The advice I see often online is to use RuntimeError as a catch-all for everything else is such cases.

What do you think is the best economical choice?

Thanks,

Steve

natasha-aleksandrova:
You can choose any one of those three -- the intent is that your client can catch the exception and understand what the issue is, the type of exception isn't really making that big of a difference besides clarity of your library or if there is some reliance on exceptions inheritance.

In a production setting you would just make a custom exception that makes it very clear what the issue is like `raise InvalidSphereParams` and the client can catch those specific exceptions instead of more generic ones like ValueError for example.

We don't talk about custom exceptions in this Lesson because it requires some understanding about classes.

PR #1129, opened 2020-09-21
updated lesson07 files

belarsongit:
let me know if I need to revise else I think I might be ready to turn in Lesson07


natasha-aleksandrova (approved):
@belarsongit looks great 👍

PR #1131, opened 2020-09-21
Lesson07 Assignment 6

bcamp3:
I tried to go without the tutorial and almost got there.  I was having trouble with all the possible indentation scenarios.  Also, I mainly used trial and error with the 'run_html_render.py' execution, but then tackled the 'test_html_render.py' tests.  In the future I will definitely try to add tests before adding features to my code.


charlesrobison (approved):
Great job pushing through on this!  Definitely a challenging assignment.  Code looks great, tests pass, produces correct results.

Ready to merge

PR #1136, opened 2020-09-25
Found issue in writing OneLineTag tags. 

belarsongit:
Commented out extra write (coder error!). Ran test and regenerated all html pages to confirm fix


natasha-aleksandrova (approved):
thank you!

PR #1137, opened 2020-09-27
Lesson #7

Dhpeterka:
HTML render assignment ready for review.


natasha-aleksandrova (approved):
@Dhpeterka nice work here!

PR #1139, opened 2020-09-28
Finished Ex 3.3 String Formats

bapandaOK44:
String Formatting Exercise
Goal
In this exercise we will reinforce the important concepts of string formatting, so that these start to become second nature!

### DIFF
--- a/students/barb_matthews/lesson3/mailroom.py
+++ b/students/barb_matthews/lesson3/mailroom.py
@@ -1,0 +1,11 @@
+#! /usr/bin/env python3
+
+## Lesson 3.3 Mailroom Part 1
+## By: B. Matthews
+## 9/28/2020
+## This is the first of a four-part assignment that will keep track of an imaginary charity's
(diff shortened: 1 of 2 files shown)


charlesrobison (approved):
Code looks good but mailroom is running as expected?  Is this ready to be reviewed?  Just making sure.  None of the options are working for me - nothing in report, can't add donor and donation, etc.  Let me know if you need help or have questions.  There are several versions of mailroom in the course that build on this, so want to make sure your off to a good start.

Ready to merge

bapandaOK44:
Only the string format part is ready to grade. I’m working on the mailroom now, but I do have a few questions if we could do a zoom sometime this weekend. 

-- Barb

bapandaOK44:
Hi Charles & Natasha,

I think there was an old file submitted, try this new one mailroom_one.py!
You are too generous on my grade, if it wasn’t printing the report :)
Can I resubmit? I want to get it all working. Let me know if it’s still missing things, but it seems to work when I run it at home.  — barb

PR #1153, opened 2020-10-04
Final updates to Lesson06, Lesson07 and Lesson08

belarsongit:
will follow up by submitting for grade.


natasha-aleksandrova:
@belarsongit I only see 2 files with formatting changes, both for lesson06.

I can go ahead and merge it but did you mean to add more files?

belarsongit:
I think (still getting up to speed with the pull requests) that Lesson07
and Lesson08 were on earlier pull requests.  I don't think that I had any
updates for them since then.  So, I guess that sounds like what you are
saying...lol.  Thanks!!

PR #1174, opened 2020-10-12
Lesson 3.3 String Lab

phrost101:


students/phrost/Lesson03/Exercises/String.py
```diff
@@ -0,0 +1,72 @@
...
+"""Out[20]: 'the 3 numbers are: 2, 3, 5'"""
+
+"""In [21]: formatter((2,3,5,7,9))"""
+"""Out[21]: 'the 5 numbers are: 2, 3, 5, 7, 9'"""
+
+print("\nTask 3:\n")
+def formatter(in_tuple):
+    fstring = ""
+    for _ in range(0, len(in_tuple)):
+        fstring += '{:d}, '
```
natasha-aleksandrova, line 29: since you're not using the index from range() you can just loop over the tuple directly
```
for _ in in_tuple:
    ...
```

natasha-aleksandrova (approved):
👍

PR #1176, opened 2020-10-12
Adding series.py file to Lesson02

kgolenk:
This is file with Fibonacci, Lucas and sum series. All tests passed.


natasha-aleksandrova (approved):
@kgolenk looks great 👍

PR #1178, opened 2020-10-13
Updates to OO Mailroom

belarsongit:


students/becky_larson/lesson09/donor_models.py
```diff
@@ -0,0 +1,293 @@
...
+        """
+
+        self._donor_name = donor_name
+        self._donations = []
+
+        if not donations:
+            self._donations = []
+        elif hasattr(donations, "__iter__"):
+            try:
+                [self.add_donation(donation) for donation in donations]
```
natasha-aleksandrova, line 33: this is considered an anti-pattern - you'll want to use a regular for loop here.

the above list comprehension does not consume the result from a function (which is None)

students/becky_larson/lesson09/donor_models.py
```diff
@@ -0,0 +1,293 @@
...
+        return is_donor_new
+
+    def append(self, donor: Donor):
+        """
+        add Donor object to collection
+        :param donor: Donor object or sequence of Donor objects
+        :return: None
+        """
+        if hasattr(donor, '__iter__'):
+            for _ in donor:
```
natasha-aleksandrova, line 256: `_` is usually used to indicate an unused variable, since you're actually using it you should use a real variable name.

natasha-aleksandrova (approved):
@belarsongit great work here!

PR #1196, opened 2020-10-18
First Python Push-up from Lesson01

Alvame118:
#NameError, TypeError, SyntaxError, AttributeError

def error_one():
    NBA_teams = ["Heat","Lakers","Celtics","Nuggets"]

    print (NBA_team);

error_one()

def error_two():
    NBA_teams = ["Bulls","Warriors","Pelicans"]
    number = 1

    result = number + NBA_teams
    print (result);

error_two()

def error_three():
    x = int(input('Enter a number:'))

    whilee x%2 == 0:
        print ('You have entered an even number')
        break
    else:
        print('You have entered an odd number')

error_three()

def error_four():
    x = 100
    x.append(5)

error_four()

### DIFF
--- a/students/Mario_Alvarenga/lesson01/break_me.py
+++ b/students/Mario_Alvarenga/lesson01/break_me.py
@@ -1,0 +1,36 @@
+#creating a script for 4 basic Python errors
+#NameError, TypeError, SyntaxError, AttributeError
+
+def error_one():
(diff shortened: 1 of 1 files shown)


natasha-aleksandrova (approved):
@Alvame118 note that you just want the code in python files. no need to also paste it in the comment section of the pull request.

PR #1197, opened 2020-10-19
#Posted my solution from one of the warmups

Alvame118:
def parrot_trouble(talking, hour):
  if talking and (hour< 7 or hour > 20):
    return True
  return False

parrot_trouble(8,5)


natasha-aleksandrova (commented):
@Alvame118 your indentation is incorrect in your code, should be 4 spaces not 2.

Also no need to paste your code into PR comment.

PR #1206, opened 2020-10-25
Lesson 05 Exercise and Assignment for Mark Taylor

mark-l-taylor:
Completed Exceptions exercise and updates to mailroom.
For mailroom.py:
I added exceptions to two of the the methods that handle user inputs.  I only found one area that could be converted to list comprehension while a implementing more changes to the code structure.  I believe all other areas of the code where list comprehension could be applied would be less efficient than the current methods being used.

### DIFF
--- a/students/mark-l-taylor/lesson05/except_exercise.py
+++ b/students/mark-l-taylor/lesson05/except_exercise.py
@@ -1,0 +1,57 @@
+#!/usr/bin/python
+
+"""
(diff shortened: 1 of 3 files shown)


charlesrobison (approved):
Looks good.  Ideally you would have a couple more exceptions in mailroom.  Agree with your sentiments on comprehensions.

Ready to merge

PR #1208, opened 2020-10-26
Lesson09 - Assignment 8 - Object Oriented Mailroom

bcamp3:
Submitting Mailroom OO assigment.  

My initial attempt used a list for the DonorCollection class, but I found myself having to return an index location based on the donor name I wanted to query.  I then switched to a dict in the DonorCollection class to be able to query by Donor name.

Also, I realize I could possibly inherit from the dict class itself but the complexity seemed more burdensome than this assignment required, and using a dict() structure inside of DonorCollection class seemed straight forward.

### DIFF
--- a/students/bcamp3/lesson09/mailroom_oo/cli_main.py
+++ b/students/bcamp3/lesson09/mailroom_oo/cli_main.py
@@ -12,43 +12,44 @@ This script accepts user input to perform the following donation database tasks:
 
 from donor_models import *
 
-c = DonorCollection()
 
 def send_thank_you():
     """Process donation and send Thank You to individual donor."""
-    name = input('\nEnter the full name of the donor > ')
+    name = input('\nEnter the full name of the donor >  ')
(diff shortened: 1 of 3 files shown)


students/bcamp3/lesson09/mailroom_oo/cli_main.py
```diff
@@ -0,0 +1,101 @@
...
+Mailroom Object Oriented script.
+
+This script accepts user input to perform the following donation database tasks:
+    1 - Display the donor database list or update the donation database with a new donation.
+        Then display a 'Thank You' message to the donor of the new donation.
+    2 - Display a donation database report
+    3 - Quit out of the script
+"""
+
+from donor_models import *
```
natasha-aleksandrova, line 13: note that using star imports is a bad practice.

natasha-aleksandrova (approved):
@bcamp3 I agree, using dictionary is a good idea and more efficient.

PR #1210, opened 2020-10-26
Updated mailroom.py with sort option

kgolenk:
Hi Natasha,

I updated mailroom.py file with sort option and also re-wrote code to avoid duplication.

Thank you!


students/kgolenk/lesson03/mailroom.py
```diff
@@ -65,23 +66,23 @@ def thank_you():
...
-                break
             else:
                 add_name(full_name)
-                add_donation(full_name)
-                send_email(full_name)
-                break
+
+            add_donation(full_name)
+            send_email(full_name)
+            break
```
natasha-aleksandrova, line 74: nice!

natasha-aleksandrova (approved):
👍

PR #1216, opened 2020-10-31
OO Mailroom Final Project

oieeri26:


students/erik_oien/lesson09/mailroom_oo/donor_models.py
```diff
@@ -0,0 +1,66 @@
...
+        "Thank you,\n",
+        "Kittens and Metal Charity\n"))
+        return thank_you
+
+class DonorCollection():
+    
+    def __init__(self):
+        self.donors = {}
+    
+    def initialize_donor_dict(donor_data):
```
natasha-aleksandrova, line 31: this should be either a staticmethod or classmethod - the way it is defined is not correct.

natasha-aleksandrova (approved):
@oieeri26 nice work here 👍

PR #1227, opened 2020-11-04
Adding lesson03 exercises

LongSA:


charlesrobison (approved):
Very thorough code here @LongSA !  A few issues with mailroom:
1) Need to be able to add new donations to new and existing donors
2) When adding new donation, user should be able to input 'list' to see current donors

Also, make sure you submit your work for the list and string formatting labs on the course site to receive grades.  Let me know if you have questions.

Ready to merge

LongSA:
All of the functionality you are asking for exists. You can add new gifts to new and existing donors. The default main display lists donors by name and selection ID. If you display the report you can select by name. This displays the detail for the individual donor. On this screen you can see their gifts, add new ones, and send thank-you messages. All of the functionality is handled. 

As for the other assignments, they should have been in the same lesson03 folder in separate subfolders. I will check again. 

Thanks - Steve

Fat finger mail

LongSA:
Ready to merge?

If you run the application mailroom.py you will see that it satisfies all of the requirements even though the requirements were written poorly. The attached PDF will show this.

The other three exercises were uploaded to GitHub in the same collective folder (lesson03) as the mailroom.py assignment. However, it appears that the web “submittal” pages have reset themselves or there is some other bug that has prevented you from receiving notification of my pull request. Would it be acceptable if I made an insignificant change to those assignments and push everything to GitHub again so that I can log different pull requests for each assignment?

Steve Long

natasha-aleksandrova:
@LongSA 

Please review assignment requirements:
> The script should prompt the user (you) to choose from a menu of 3 actions: “Send a Thank You”, “Create a Report” or “quit”.


can you explain to us how a user can add donations to new or existing user?


----


for submissions on the course website, copy the Pull Request PR link and paste into each submission box for lesson 03 (there will be a total of 4)

natasha-aleksandrova:
you can create a new PR with fixes for missing requirements.

LongSA:
I can certainly explain how mailroom.py works although you should have seen
a very detailed pdf file on this subject in my response @ 02:56. 

As you may have noticed, it's written for someone as though it were a real 
application. Putting a little more work in is how I learn and retain the 
skills. Putting every functionality on one page when the data model requires
a minimum of two levels (donor, gift) would not pass muster where I work.

Here are demo steps similar to what I write for user-acceptance tests I've 
authored. It’s very procedural.

1-In a terminal window, START my (steve long) application ./mailroom.py. 

2-VERIFY that at the top you see an app title "Mailroom v1.0".

3-VERIFY that below the app title you see the page header that starts with
  "-- MAILROOM".

NOTE: The following line verifies the implied display feature of the "list" 
      requirement.
      
4-VERIFY that you see an enumerated list of names (the donors.)

5-VERIFY that at the bottom you see a user input prompt line that looks like:
  "INPUT: Enter donor ID to access, [:N]ew donor, [:R]eport, [:Q]uit > "

NOTE: The following steps verify the "report" requirement.

6-TYPE-AND-ENTER ":R" (for "report".)

7-VERIFY that you see the page header that starts with "-- MAILROOM".

8-VERIFY that you see a table entitled "Donor Report".

NOTE: The following steps verify the implied command feature of the "list"
      requirement.
      
9-VERIFY that you see the user input prompt line that looks like:
"INPUT: Enter donor name to access, [:N]ew donor, [:L]ist donors, [:Q]uit > "

10-TYPE-AND-ENTER ":L" (to "list" donors.)

11-VERIFY that you see the page header that starts with "-- MAILROOM".

12-VERIFY that you see the enumerated list of names from Step 4.

NOTE: The following steps verify the requirement for accessing an existing 
      donor and adding a new gift.
      
13-TYPE-AND-ENTER the ID for "Albert Camus", e.g., "1" (to open the DONOR 
   page for selected donor.)

14-VERIFY that you see the page header "-- MAILROOM: DONOR (Albert Camus)".

15-VERIFY that you see the user input prompt line that looks like:
"INPUT: Enter gift value ($), [:M]essenger to send thank you, or e[:X]it >"
 
16-RECORD (screen capture) the donor's page.
   
17-TYPE-AND-ENTER a new gift value, "21.63", for the donor.

18-VERIFY that (a) the donor's "Num Gifts" has increased by 1 from Step 16,
   (b) the donor's "Gifts" now shows the value entered from Step 17, and 
   that (c) the donor's "Total Given" and "Average Gift" values have updated.

NOTE: The following steps verifies the implied command and display features of 
      the "list" requirement.

19-TYPE-AND-ENTER ":X" (to exit back to MAILROOM page.)

20-VERIFY that you see the page header that starts with "-- MAILROOM".

21-VERIFY that you see the enumerated list of names from Step 4.

NOTE: The following steps verify the requirement for adding a new donor.

22-TYPE-AND-ENTER the ":N" (for signaling app to accept the next typed value
   as the name of a new donor.)

23-TYPE-AND-ENTER "Noam Chomsky".

24-VERIFY that the name "Noam Chomsky" appears in the donor list and that the
   donor count has increased by 1.

NOTE: The following steps verify the requirement for accessing a new donor 
      and adding the donor's first gift.

25-TYPE-AND-ENTER the ID for donor "Noam Chomsky" (probably "7".)

26-VERIFY that you see the page header "-- MAILROOM: DONOR (Noam Chomsky)".

27-VERIFY that you see the user input prompt line that looks like:
"INPUT: Enter gift value ($), [:M]essenger to send thank you, or e[:X]it >"
 
28-RECORD (screen capture) the donor's page.
   
29-TYPE-AND-ENTER a new gift value, "186000.00", for the donor.

30-VERIFY that (a) the donor's "Num Gifts" has increased by 1 from Step 28,
   (b) the donor's "Gifts" now shows the value entered from Step 29, and 
   that (c) the donor's "Total Given" and "Average Gift" values have updated.

NOTE: The following steps verifies the implied command and display features of 
      the "list" requirement.
      
31-TYPE-AND-ENTER ":X" (to exit back to MAILROOM page.)

32-VERIFY that you see the page header that starts with "-- MAILROOM".

33-VERIFY that you see the enumerated list of names from Step 24.

QED.

students/steve-long/lesson03-seq-iter-strings/ex-3-4-mailroom_part_1/mailroom.py
```diff
@@ -0,0 +1,1218 @@
...
+    containing a decimal.
+    """
+    status = False
+    f = None
+    try:
+        f = float(v)
+        status = (f is not None)
+    except Exception:
+        status = False
+    return status
```
natasha-aleksandrova, line 215: this can be simplified to 
```
try:
    float(v)
except Exception:
    return False
else:
    return True
```

```

natasha-aleksandrova:
I've looked again and I now see the steps to add donation.

LongSA:
Thanks. That’s the kind of feedback that I can use.

LongSA:
One question wrt your recommendation: In other languages (and on some projects I’ve worked on in the past) there was a policy that said that there should be a single exit point in a function. This was my reason for declaring a variable whose valhe gets assigned conditionally and passed out of the function at the end. Is there a style or policy or trend toward such a practice in the Python world?

natasha-aleksandrova:
Python world will usually lean towards a more readable / DRY code; this will usually tie directly to less lines of codes.


There isn't usually simple one answer and it will depend on the code. 

Your function has simple logic but the code is a bit too complicated for such simple logic, thus simplifying the code now matches the logic.

In another case, if you had a function with more complex business logic you could benefit from reducing multiple exit points to a single exit point to make it more readable.

LongSA:
I wasn’t sure if I would be giving away too much by asking about this in the forum (not sure what the rules are about that) so here is my question:

The .append() instance method on the Element class does not appear to have a requirement that we check whether an html tag is equal to or subordinate to the one you are appending to. Should that be something that we do and if so, should it throw an exception for appending a tag in the wrong place? For instance, and Html should not be appended to an Li (list item).

Steve Long

natasha-aleksandrova:
In that assignment each class represents an element, so the append() method would only ever append to that specific element.

LongSA:
There are several ways of solving the object problems for lesson 7. You have 3 methods specified in the base class (Element)...

.__init__
.append
.render

…and two basic kinds of content you are “appending” to an element…

another Element
text

Turning the tag attributes and contents into strings requires quite a bit for string formatting.

The pseudo-explanation in the exercise text suggests using one of two solution paths in which you work in (a) the .append method or (b) the .render method with significant customization of each depending on the kind of tag.

I have built a third solution that adds another method I call .to_string for string representation of the instance (and its content) without writing it to file. Most of the differences between each tag class are covered in this method. Method .to_string is called by the .append method.

My question is: Are you looking for a specific solution or is my solution valid? I can’t tell from the way these exercises are written and the “Python style” seems arbitrary.

The answer to this takes on importance to me in my work. I’ve had to tear apart several scripts that my organization depends on because the object-oriented Python code was not very object-oriented. Methods had the same name and carried a lot of redundant steps that could have been refactored into “internal” methods (and thus have made subsequent work purely additive). Your suggestion here will influence how I build code for my organization in the future.

Thanks,

Steve Long

LongSA:
Folks,

I’m following another solution path for this lesson so you can disregard my specific question about the solution. I’m still using helper methods because that makes the code easier to look at.

I’m still interested in your feedback on the tradeoff between making code very terse and less object-oriented versus more object-oriented.

One other thing: This lesson’s test module for Step 9 doesn’t work because the test code is expecting you to pass a literal string (a set of space chars) for the indentation. I am passing the 
number of chars to offset by, which is how I usually do this sort of thing because strings are harder to eyeball and write automation for. In the Step nine directions, it demonstrates using the same technique that I am using. If you are using this test to check our work, it will fail. I did a manual check and found that my code is doing what I expected it to do.

Steve Long

natasha-aleksandrova:
Thanks for giving us heads up on the step 9 tests.

As far as your question for what solution we're looking for - as long as it follows good OO practices then it's a good solution.

Totally fine to add another method .to_string() as long as you're not doing logic in there checking which element you're generating string for and instead it's defined in each unique class. I don't have a full picture on your solution and looking forward to seeing it.

PR #1230, opened 2020-11-07
submitting lesson 3 string formatting assignment

cbrown285:


students/cbrown/Lesson 3/String Formatting.py
```diff
@@ -0,0 +1,43 @@
+#Lesson 3 String Formatting
```
natasha-aleksandrova, line 1: tip: avoid having spaces in python files names, every python file is a module by default which cannot have spaces

natasha-aleksandrova (approved):
👍

PR #1242, opened 2020-11-13
submitting mailroom assignment part 1

cbrown285:


students/cbrown/Lesson 3/mailroom.py
```diff
@@ -0,0 +1,103 @@
...
+                name[1].append(amount)
+                print(donors)
+
+            else:
+                continue
+
+    #Donor name not found in list
+    else:
+        tup = (donor_name,amount)
+        donors.append(tup)
```
natasha-aleksandrova, line 51: this loop is a bit inefficient - the `donor_name in donor_list` will search the whole list, then `for name in donors` will do the search again.

You can just have a single loop to accomplish the same.

```
for donor in donors:
    if name[0] == donor_name:
        donor[1].append(amount)
        break
else:
    # not found
    donors.append((donor_name, amount))
```

students/cbrown/Lesson 3/mailroom.py
```diff
@@ -0,0 +1,103 @@
...
+        item = donations[1]
+        return sum(item)
+
+    sorted_donations = sorted(donors,key = sum_value, reverse=True)
+
+    for item in sorted_donations:
+        name = item[0]
+        Total_Given = sum(item[1])
+        Num_Gift = len(item[1])
+        Avg_Gift = Total_Given/Num_Gift
```
natasha-aleksandrova, line 79: all var names should be lower case

students/cbrown/Lesson 3/mailroom.py
```diff
@@ -0,0 +1,103 @@
...
+        if response == "1":
+            send_thanks()
+
+        elif response == "2":
+            create_report()
+
+        elif response == "3":
+            exit_from()
+
+main()
```
natasha-aleksandrova, line 103: this should be in `if __name__ == "__main__"` block

natasha-aleksandrova (approved):
overall nice work!

PR #1243, opened 2020-11-13
Mail Room 2 + Exceptions

phrost101:


students/phrost/Lesson04/Mailroom_Part_2.py
```diff
@@ -0,0 +1,113 @@
...
+        if response == "1":
+            thank_you()
+        elif response == "2":
+            report()
+        elif response == "3":
+            thank_all()
+        elif response == "4":
+            break
+        else:
+            print("Please select a number between 1 and 4")
```
natasha-aleksandrova, line 32: you should be using a switch dict in this lesson, however I see that you've applied that in lesson 05 so that will work.

natasha-aleksandrova (approved):
great work here!

PR #1251, opened 2020-11-21
Lesson 4 exercises and assignment for Emory Christenson

lightener:


students/emorychristenson/lesson04/mailroom_pt2.py
```diff
@@ -0,0 +1,103 @@
...
+        print('{:<20} {:>5}{:>9.2f}{:>13} {:>9}{:>8.2f}'.format(name, '$', total, num, '$', average))
+    main()
+
+def send_all_thanks():
+    for donor, amount in donors.items():
+        total = sum(amount)
+        filename = ('./' + donor + '.txt')
+        with open(filename, 'w') as f:
+         f.write(thanks_letter(donor, total)) 
+    main()
```
natasha-aleksandrova, line 84: you don't want to call main() manually -- you should call main() in while True loop; see lesson03 mailroom guidelines.

your functions should just exit.

students/emorychristenson/lesson04/mailroom_pt2.py
```diff
@@ -0,0 +1,103 @@
...
+    response = int(input("Please enter a number: "))
+    if response in menu:
+        menu[response]()
+    else:
+        response = int(input(f"{response} is not a valid input: "))
+
+
+if __name__ == "__main__":
+    print("Welcome to the donors list!")
+    main()
```
natasha-aleksandrova, line 103: main() call should be wrapped in while True loop.

students/emorychristenson/lesson04/mailroom_pt2.py
```diff
@@ -0,0 +1,103 @@
...
+        main()
+    else:
+        amount = float(input("\nEnter a donation amount: "))
+        
+    for donor in donors.keys():
+        if response.lower() == donor.lower():
+            donors[donor].append(amount)
+        else:        
+            donors[response] = [amount]
+        send_email(donor, amount)
```
natasha-aleksandrova, line 50: with dictionaries you should be checking if a key exists, instead of looping over keys:


```
if donors.get(donor):
     # append amount
else:
    # add new donor
```

natasha-aleksandrova (commented):
@lightener please see my comments on the mailroom and address at your convinience.

you can resubmit a new PR when you're ready.

PR #1252, opened 2020-11-22
Lesson 8

roslynm:
Lesson 8 Assignment


students/roslyn_m/lesson08/test_circle.py
```diff
@@ -0,0 +1,143 @@
+import pytest
+from circle import *
```
natasha-aleksandrova, line 2: in general you want to avoid start imports, instead -> `from circle import Circle`

natasha-aleksandrova (approved):
@roslynm nice work here!

PR #1253, opened 2020-11-22
Lesson05 mailroom.py part 3, exception_exercise.py

kylenod:
Hello,

Submitting assignments for lesson 5. I built out some exception handling for mailroom.py, but I did not think there were any good places to use Comprehensions (i.e., no places in which a sequence of items needed to be processed to create a new sequence of items). I could be wrong, or maybe my code was missing something it should have had? Let me know if I should make any updates.

Best,
Kyle


natasha-aleksandrova (approved):
@kylenod code looks really good and clean.

one thing to keep in mind is to make sure you're exception handling is just around the code that can throw that exception; as far as list comprehensions, it's ok if you can't find any place for them, it is not uncommon.

PR #1261, opened 2020-11-23
python puzzles exercise

damclaugh:
Some of the warmup python exercises from codingbat.com


natasha-aleksandrova (approved):
you do want to make sure that your indentation is 4 spaces as recommended by python; you currently have just two spaces.

PR #1265, opened 2020-11-25
Added solutions to Lesson 4 Assignments 1-3 and supporting files.

Brent-K:


charlesrobison (approved):
Great job with trigrams!  Looks good, works well.  Mailroom is throwing an error when adding new donors - please take another look and reach out if you need help.

Ready to merge

Brent-K:
Charles,
Could you help me understand the error you ran into better?I tried adding a few donors using: ‘[2] Create thank you for new donor.’  The only error I could generate was if thedonation amount provided could not be converted into a float (perhaps the userentered a ‘$’ or the donation entered was a word). I did run into an error for :‘[1] Create thank you for existing donor.’ If the name entered was not on thelist of donors already (case sensitize). For this, the error doesn’t happenuntil the call to ‘thank_you()’ because it tries to use a dictionary that wasnever built (I think nothing is actually passed into the ‘thank_you()’ call). Theprogram assumes a new user will only ever be added by using option [2]. I wasn’tsure how much error handling to include at this point. Are these examples ofthings I should trap for on this assignment? Was your error something else entirely?
 

v/r,

Brent Kieszling


    On Friday, November 27, 2020, 5:32:52 PM PST, Charles Robison <notifications@github.com> wrote:  
 
 


@charlesrobison approved this pull request.

Great job with trigrams! Looks good, works well. Mailroom is throwing an error when adding new donors - please take another look and reach out if you need help.

Ready to merge

charlesrobison:
Hi Brent - I revisited this and it's running correctly.  I must have made a mistake before.  I'll adjust your grade.  Regarding tests - we'll get to that in a later version of mailroom.

PR #1271, opened 2020-12-01
added lesson05 folder

LongSA:
List comprehension and exception handling; two graded assignments, two ungraded assignments. Please check README.txt files for instructions.

### DIFF
--- a/students/steve-long/lesson05-excepts-comps/ex-5-4-mailroom_part_3/mailroom3.py
+++ b/students/steve-long/lesson05-excepts-comps/ex-5-4-mailroom_part_3/mailroom3.py
@@ -1,0 +1,1522 @@
+#!/usr/bin/env python3
+# =============================================================================
+# Python210 | Fall 2020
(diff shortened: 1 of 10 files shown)


students/steve-long/lesson05-excepts-comps/ex-5-4-mailroom_part_3/mailroom3.py
```diff
@@ -0,0 +1,1522 @@
...
+    Can a value be coerced to a floating-point value? Implemented
+    because just checking for <v>.isnumeric() won't work with a value
+    containing a decimal.
+
+    HIST: 001/Incorporated Natasha's brevity suggestion.
+          000/New
+    """
+    try:
+        float(v)
+    except Exception:
```
natasha-aleksandrova, line 140: note that it is best practice to catch specific exceptions like TypeError/ValueError etc

it is pretty obvious in this case though on what you're attempting to catch.

natasha-aleksandrova (approved):

PR #1272, opened 2020-12-01
Mailroom 2

bapandaOK44:


charlesrobison (approved):
Code looks good but missing a couple of features:
1 - report should be sorted by total given in descending order
2 - when adding new donations, new amounts are not added to report nor letter

Please take another look and let us know if you need any help.

Ready to merge

bapandaOK44:
Hey Charles,

Ok, I’ll check the sorting. I thought I did do it …
Are you sure about new amounts not being added? Is that for the ‘all’ letters written to files? I added them to Mailroom 1 so was that wrong? Maybe I read the assignment wrong??

— barb m

The Program: Part 1 <https://uwpce-pythoncert.github.io/PythonCertDevel/exercises/mailroom-part1.html#the-program-part-1>
Write a small command-line script called mailroom.py. This script should be executable. The script should accomplish the following goals:

It should have a data structure that holds a list of your donors and a history of the amounts they have donated. This structure should be populated at first with at least five donors, with between 1 and 3 donations each. You can store that data structure in the global namespace.
The script should prompt the user (you) to choose from a menu of 3 actions: “Send a Thank You”, “Create a Report” or “quit”.
Send a Thank You <https://uwpce-pythoncert.github.io/PythonCertDevel/exercises/mailroom-part1.html#send-a-thank-you>
If the user (you) selects “Send a Thank You” option, prompt for a Full Name.
If the user types list show them a list of the donor names and re-prompt.
If the user types a name not in the list, add that name to the data structure and use it.
If the user types a name in the list, use it.
Once a name has been selected, prompt for a donation amount.
Convert the amount into a number; it is OK at this point for the program to crash if someone types a bogus amount.
Add that amount to the donation history of the selected user.

students/barb_matthews/lesson4/mr2.py
```diff
@@ -0,0 +1,141 @@
...
+        print("\n" * 50, "Dear %s,\n\nThank you for your generous donation of $%.2f. "
+                          "We appreciate your support.\n\nSincerely, Grateful Team" % (name, what))
+        print("\n" * 5)
+
+    else:
+        what_list = donors[name]
+        all = sum(what_list) + what
+        print("\n" * 50, "Thank you, %s for your continuing generous donations of $%.2f.\n"
+                          "We appreciate your support.\n\nSincerely, Grateful Team" % (name, all))
+        print("\n" * 5)
```
natasha-aleksandrova, line 105: @bapandaOK44 
when it is an existing donor you can see here that you're not actually updating donations for existing donor in here ^^

you would need to do something like `donors[name].append(what)`

PR #1274, opened 2020-12-02
Lesson 09

roslynm:


students/roslyn_m/lesson09/mailroom_lesson9.py
```diff
@@ -0,0 +1,205 @@
...
+    def print_to_user(string):
+        """ print data to user
+            return: nothing
+                """
+        print(string)
+
+
+# ---------------- Processing Classes ------------------ #
+
+class Processing():
```
natasha-aleksandrova, line 106: I believe this should be a DonorCollection class that holds reference to multiple donor objects and below functions should be class methods instead of staticmethods.

Notice that they all accept donor_all param - which usually is an indicator that it should be part of the class instance.

You can review the mailroom guidelines for more details.

natasha-aleksandrova (commented):
Donor class looks fine but you should incorporate DonorCollection class.

PR #1279, opened 2020-12-05
Mailroom Part 3

cbrown285:


students/cbrown/Lesson 5/mailroom_part3.py
```diff
@@ -0,0 +1,117 @@
+#mailroom part 3: adding exceptions & comprehension
+#!/usr/bin/env python3
+#Mailroom Script Part 3. Upgrading with comprehensions & exceptions
+import sys
+import tempfile
```
charlesrobison, line 5: tempfile imported but not used

students/cbrown/Lesson 5/mailroom_part3.py
```diff
@@ -0,0 +1,117 @@
...
+        if donor_name == 'List':
+            print('\n',"Here is A Complete List of Donor Names\n ")
+            print(donor_list)
+        else:
+            break
+
+    #prompt for donation amount
+    try:
+        amount = float(input("How Much Did This Person Donate?\n "))
+    except: #start prompt over again if exception occurs
```
charlesrobison, line 37: Ideally you'd want an explicit exception here

students/cbrown/Lesson 5/mailroom_part3.py
```diff
@@ -0,0 +1,117 @@
...
+
+    #function for sorting key
+    def sum_value(donations):
+        item = donations[1]
+        return sum(item)
+
+    #sort donations dict by total amount
+    sorted_donations = sorted(donors.items(),key= sum_value,reverse=True)
+
+    for k,v in sorted_donations:
```
charlesrobison, line 76: Need a space after comma

students/cbrown/Lesson 5/mailroom_part3.py
```diff
@@ -0,0 +1,117 @@
...
+
+    #Write Thank You Note
+    message = ('Dear {donor_name}, \n\nThank you for your show of support and generosity. '
+      'Your Donation of ${amount} will contribute to saving Olympic Marmots '
+      'in Washington State. These Marmota are special and a unique gift to the Olympic '
+      'National Park ecosystem. As a way of saying thank you. '
+      'You will be receiving your very own Olympic Marmot t-shirt in the mail!\n\n'
+      'Sincerely,\n\nThe Olympic Marmot Wildlife Foundation\n').format(**thanks_dict)
+
+    print(message)
```
charlesrobison, line 61: Getting an error when adding more than one donation as well as printing report after adding donation.

charlesrobison (approved):
Nice start on exceptions and comprehensions but having some functionality issues with the database of donors when adding donations.  Take a look and let me know if you have questions.

Ready to merge

students/cbrown/Lesson 5/mailroom_part3.py
cbrown285: Hi Charles,

Thank you very much for the feedback. I have modified the program and uploaded a newer version. I found the bug that caused the error. When a new donor name was being added to the dictionary (if it didn't already exist) I was adding just a float value to the dictionary instead of a list. So the next time the user went to add a second value to that new donor entry it was trying to append a float value to another float value instead of a list. 

New changes have fixed issue.

Thanks,

Chris

charlesrobison (approved):
Great!  Glad you were able to figure it out.  I'll update the grade.

PR #1280, opened 2020-12-06
Emory Christenson - Updating mailroom assignment based on Charles' feedback

lightener:
I removed the while loop on the amount input, and added exception handling to the main function.


charlesrobison (approved):
Good updates - works well!  Grade adjusted.

Ready to merge

PR #1284, opened 2020-12-08
Submitting Lesson03 and Updating labs

JohnSkull:
I think I incorporated all requested changes, and I updated some labs as well.


students/john_hunter/lesson03/mailroom.py
```diff
@@ -0,0 +1,219 @@
...
+        donation = int(input('Enter a donation amount, enter \'0\' to stop adding donation values: '))
+        if donation == 0:
+            break
+        else:
+            int(donation)
+            donations.insert(-1,donation)
+    for i in range(len(donors)):
+        if name == donors[i][0]:
+            donors[i] = donors[i] + donations
+            break
```
natasha-aleksandrova, line 149: there is no need to use range() to loop over your list:
```
for donor in donors:
    if name == donor[0]:
        donor.append(donation)
        break
else:
    donors.append([name, donation])
```

this is much simpler ^^ and covers a lot of logic across `add_donations` and `add_donor` functions.

natasha-aleksandrova (commented):
@JohnSkull please review my comments.

slicing assignment needs a few functions fixed to update to use slicing. 

mailroom assignment also needs to have the comments addressed, as well as I'm getting an error when running it - however if you apply my suggested code I think it should resolve the bug.


```
Enter full name of donor or 'list' to request list of donors: natasha
Would you like to add more donations for natasha? y/n: y
Enter a donation amount, enter '0' to stop adding donation values: 19938347
Enter a donation amount, enter '0' to stop adding donation values: 0
['natasha', '100', 19938347]
  File "student.py", line 217, in <module>
    main()
  File "student.py", line 210, in main
    send_ty()
  File "student.py", line 98, in send_ty
    totals = add_donations(inputName)
  File "student.py", line 151, in add_donations
    return sum(donors[i][1:])
TypeError: unsupported operand type(s) for +: 'int' and 'str'```


Another note, code has a lot of pep8 violations - please run a separate linter or I recommend using Pycharm which has everything built-in.

natasha-aleksandrova:
@JohnSkull make sure you're submitting your PR links on the course website to receive a grade.

PR #1291, opened 2020-12-13
Lesson 09 Assignment

mellamokrysti:
Mailroom refactoring and making it OO.


students/Kristy_Martini/lesson09/mailroom.py
```diff
@@ -0,0 +1,149 @@
...
+        Cathy = Donor("Cathy Martini", 0, 7, 63833)
+        Bill = Donor("Bill Martini", 0, 10, 60000)
+        Nick = Donor("Nick Martini", 0, 50, 47484949)
+
+        initial_donors = [Kristy, Mikey, Nick, Bill, Cathy]
+        [self.add_donor(donor) for donor in initial_donors]
+
+    def add_donor(self, Donor):
+        """ Add new donor to report"""
+        self.donors[Donor.name] = Donor
```
natasha-aleksandrova, line 19: should be lower case var -> `donor`

students/Kristy_Martini/lesson09/mailroom.py
```diff
@@ -0,0 +1,149 @@
...
+            new_report = Report()
+        new_report.sort_donors(new_report)
+        report_lines = []
+        report_lines.append("Donor Name          | Total Given   | Num Gifts | Average Gift")
+        report_lines.append("--------------------------------------------------------------")
+        for value in new_report.sorted_dict.values():
+            new_line = '{0:21}'.format(value.name) + "$" + '{0:14}'.format(value.total_gift_value) + '{0:12}'.format(value.num_gifts) + " $" + '{0:12}'.format(value.average_gift)
+            report_lines.append(new_line)
+        report_lines.append("\n")
+        return report_lines
```
natasha-aleksandrova, line 43: above patterns aren't following best practices:

`new_report = self` - your report data is local to the function. Use a local variable and return the sorted data.
if you really want to store the data, create a class attribute `self.sorted_data` in `__init__`

same with `create_report`,  you would not create an instance of itself unless it was a `staticmethod` which you don't need anyways. This function should use variables with local scope -- to access your sort function: `self.sort_donors()`

students/Kristy_Martini/lesson09/mailroom.py
```diff
@@ -0,0 +1,149 @@
...
+        """ Report class holds the list of all donors and information associated with them."""
+        self.donors = dict()
+        Kristy = Donor("Kristy Martini", 0, 3, 3000)
+        Mikey = Donor("Mike Martini", 0, 7, 3424834)
+        Cathy = Donor("Cathy Martini", 0, 7, 63833)
+        Bill = Donor("Bill Martini", 0, 10, 60000)
+        Nick = Donor("Nick Martini", 0, 50, 47484949)
+
+        initial_donors = [Kristy, Mikey, Nick, Bill, Cathy]
+        [self.add_donor(donor) for donor in initial_donors]
```
natasha-aleksandrova, line 15: this is an anti-pattern, your comprehension returns a list which should be stored into something.

in your case you should just use a regular for loop.

students/Kristy_Martini/lesson09/mailroom.py
```diff
@@ -0,0 +1,149 @@
...
+            new_line = '{0:21}'.format(value.name) + "$" + '{0:14}'.format(value.total_gift_value) + '{0:12}'.format(value.num_gifts) + " $" + '{0:12}'.format(value.average_gift)
+            report_lines.append(new_line)
+        report_lines.append("\n")
+        return report_lines
+
+    def display_report(self, new_report):
+        report_lines = new_report.create_report(new_report)
+        for line in report_lines:
+            terminal_print(line)
+        prompt_user(new_report)
```
natasha-aleksandrova, line 49: same here - `self.create_report` is how you would use your class method.

also, prompt_user should be outside of this class.

students/Kristy_Martini/lesson09/mailroom.py
```diff
@@ -0,0 +1,149 @@
...
+        name_to_thank= input_call("Please enter the full name of the donor you'd like to thank, or enter 'quit' to exit: ")
+        if name_to_thank == "quit":
+            prompt_user(new_report)
+    try: 
+        donor = new_report.donors[name_to_thank]
+    except KeyError:
+        terminal_print(name_to_thank, " is a new donor. They will be added to our database.")
+        newDonor = Donor(name_to_thank, 0, 0, 0)
+        new_report.add_donor(newDonor)
+        donor = new_report.donors[newDonor.name]
```
natasha-aleksandrova, line 89: you should have a method for this access -- remember that we don't want users to have this implementation detail if we decide to change the data structure.

natasha-aleksandrova (commented):
@mellamokrysti please review my comments and let me know if you have any questions to understand OO concepts better.

PR #1325, opened 2021-01-18
add lesson09

zzsocool:
add cli_main.py 
add donor_models.py
add test_mailroom_oo.py

### DIFF
--- a/students/Z_shen/lesson09/donor_models.py
+++ b/students/Z_shen/lesson09/donor_models.py
@@ -1,0 +1,68 @@
+class Donor(object):
+    def __init__(self, name=None, amount=None):
+        self.name = name.title()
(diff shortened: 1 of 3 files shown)


students/Z_shen/lesson09/donor_models.py
```diff
@@ -0,0 +1,68 @@
...
+                                              '''.format(self.name, self.amount))
+        return content
+
+
+class DonorCollection:
+
+    def __init__(self, donor_list=None):
+        self.donors = {}
+        if donor_list is not None:
+            self.donors = donor_list
```
natasha-aleksandrova, line 30: you should be using your Donor class here

create Donor instance and store it in this collection class, here is some stubbed out code as an example:

PR #1327, opened 2021-01-20
Completed lesson07 and lesson08 and lesson 09 Exercises and Assignments

craigsimmons:
This PR contains new code for lessons 07 and 08.


craigsimmons:
Added Lesson09 files and oo Mailroom Assignment

students/csimmons/Lesson09/donor_models.py
```diff
@@ -0,0 +1,89 @@
...
+
+    donors_db = {
+    'Mary Newcomer' : [10000, 2500, 300],
+    'Christine Ruotolo' : [3000, 6000, 750, 20000],
+    'Martin Acevedo' : [2000, 5000],
+    'Sutton Keaney' : [24500, 500, 3000, 5000, 1000],
+    'David Basilio' : [750, 750, 750, 750, 5000, 750, 750],
+    'Andrew Laughlin' : [2500, 500, 40000, 50],
+    'Hussein Saffouri' : [1000, 1000, 2100, 7000, 55000],
+    }
```
natasha-aleksandrova, line 45: this data should be on instance level not class level.

Use class attributes for things that should belong to a class, for example something like a configurable setting.

Instance attributes should be unique to each instance of the class, and in this case donors are unique to each instance.

students/csimmons/Lesson09/donor_models.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def update_donor(self, donor, donations):
+        new = Donor(donor, donations)
+        self.donors_db.setdefault(new.name, []).append(new.donations)
+        return self.donors_db
+
+    def edit_donor(self, donor, donations):
+        new = Donor(donor, donations)
+        self.donors_db.setdefault(new.name, []).append(new.donations)
+        return self.donors_db
```
natasha-aleksandrova, line 76: ideal design here is to have a collection of Donor objects -- right now you have a simple data structure.

natasha-aleksandrova (approved):
great start on lesson09 but few issues with design.

please see an example on the integration of the two classes:

PR #1335, opened 2021-02-03
mailroom oo. Resubmitting with improvements & changes 

cbrown285:
Hi Natasha,

I've made some changes and attempted to address the issues you raised.

Thanks for your feedback!

-Chris

### DIFF
--- a/students/cbrown/Lesson 9/mailroom_oo.py
+++ b/students/cbrown/Lesson 9/mailroom_oo.py
@@ -1,0 +1,223 @@
+import sys
+
+
+class DonorCollection():
+
+    def __init__(self):
+
+        self.donor_dict = {"Bill Gates": [539000, 235642],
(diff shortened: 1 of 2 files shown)


students/cbrown/Lesson 9/mailroom_oo.py
```diff
@@ -0,0 +1,223 @@
...
+    '''
+    creates and writes a letter to each donor in the donor database
+    '''
+
+    for name, amount in donor_collection.donor_dict.items():
+        donor = Donor(name, amount)
+
+        message = create_message(donor)
+
+        write_file(donor.name, message)
```
natasha-aleksandrova, line 158: your cli code should be interacting with DonorCollection class only. 

Donor objects are referenced in the DonorCollection class.

see example:
https://repl.it/@naaleksandrova/Lesson9Example#main.py

